PR #1, opened 2021-03-01
Implement primitive word-jumping

nschoellhorn:
With this patch, if you hold ALT (or Option on Mac) and then press Arrow Left or Arrow Right, the cursor will jump over the whole word instead of just one character to allow for more natural text editing.

The current implementation is very primitive, as it only counts `\t` and ` ` (&lt;space&gt;) as whitespace. Also, the cursor placement algorithm only works for the current one-line-editing. As soon as we add soft-wrapping when a line exceeds the terminal width, the placement wouldn't work anymore, which shouldn't be an issue right now since this is the same for backspace as well as far as I can tell.


jonathandturner:
Looks good!

PR #5, opened 2021-03-07
Use Unicode words for movement

basile-henry:
This PR also makes word movement skip multiple consecutive non-word characters (word boundaries) in a similar way to how bash/readline do it.

### DIFF
--- a/src/line_buffer.rs
+++ b/src/line_buffer.rs
@@ -100,36 +100,61 @@ impl LineBuffer {
     }
 
     pub fn move_word_left(&mut self) -> usize {
-        match self
-            .buffer
-            .rmatch_indices(&[' ', '\t'][..])
-            .find(|(index, _)| index < &(self.insertion_point - 1))
-        {
-            Some((index, _)) => {
-                self.insertion_point = index + 1;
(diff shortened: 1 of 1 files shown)


src/line_buffer.rs
```diff
@@ -100,38 +100,63 @@ impl LineBuffer {
...
+
+            return self.insertion_point;
         }
-        self.insertion_point
     }
 }
 
+/// Match any sequence of characters that are considered a word boundary
+fn is_word_boundary(s: &str) -> bool {
+    !s.chars().any(char::is_alphanumeric)
```
basile-henry, line 157: Note: This logic to identify a str as being a word boundary or not is copied from [`unicode-segmentation`](https://github.com/unicode-rs/unicode-segmentation/blob/master/src/word.rs#L678-L681) as it doesn't seem to be exposed as part of the public API.

jonathandturner: Seems okay. We might want to put an issue on `unicode-segmentation` and ask if they'd be willing to make it public (if you haven't already)

jonathandturner:
Looks good. thanks for simplifying the logic, too.

PR #6, opened 2021-03-08
Simple history implementation

nschoellhorn:
This PR provides a very simple history implementation by saving buffers into a `VecDeque` when they are "confirmed" by pressing enter. The history currently allows 100 entries, which can be changed by changing the `HISTORY_SIZE` constant in `main.rs`.  
Navigation through the history works by pressing Up/Down arrow keys and feels pretty much the same as in bash or zsh. The implementation comes with a little bit of code duplication that I left there intentionally since I didn't want to take the fun of doing cool refactorings from the stream ;) 

If you want, I could of course improve that implementation by moving repainting of the current line buffer into a separate method or something like this. Just let me know!

The current implementation doesn't support persisting the history on disk for re-using it between launches and also does no fancy matching of the characters that are currently in the buffer to only match on history entries that start with the same characters, like zsh does it. If you'd like me to add something like this, also let me know, but that might be better for a separate PR. Also, I wasn't too sure if a history implementation belongs to the line editor or more to the application that uses the line editor. Most likely, the history should be more like an API that the outside program uses to provide the history to the line editor?!

### DIFF
--- a/src/line_buffer.rs
+++ b/src/line_buffer.rs
@@ -25,6 +25,16 @@ impl LineBuffer {
         &self.buffer
     }
 
+    pub fn set_buffer(&mut self, buffer: String) {
+        self.buffer = buffer;
+    }
+
+    pub fn move_to_end(&mut self) -> usize {
+        self.insertion_point = self.buffer.len();
+
+        self.insertion_point
+    }
+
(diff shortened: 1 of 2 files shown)


jonathandturner:
Looks good! Thanks :)

PR #9, opened 2021-03-10
Improve inc/dec of insertion point

jasonrhansen:
Removed need to collect grapheme indices into a `Vec`. Also, we can slice
the buffer from or to the insertion point to not have to iterate through
all of the indices.


jonathandturner:
Much cleaner! I like it

PR #10, opened 2021-03-10
Implement several new commands

jasonrhansen:
- More keys for moving around.
- `Ctrl+k`, `Ctrl+u`, `Ctrl+w`, and `Alt+d` cut text that can be pasted with `Ctrl+y`.
- Fixes a bug with backspace. Before when the insertion point was 0, but the buffer wasn't empty, it would still delete characters as if pressing delete.


jonathandturner:
This looks really good. Thanks!

PR #13, opened 2021-03-13
add diagnostic routine with cmd line arg

fdncred:
If ran with `cargo run -- -k` reedline with execute in diagnostic mode in order to show you the events that are captured from the keyboard. This is just a quick hack that allows you to see if crossterm can see the events.

If ran with `cargo run` reedline executes normally.

No work was done to formalize command line parameter parsing. This is just a quick hack to help me figure out that iterm2 appears to be eating my up and down arrow event on my mac.

Looking at the diff, it looks like a lot of code changed but i really only added one fn `print_events` and a few lines in `main`. The rest are `fmt` changing the format.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -91,92 +91,176 @@ fn main() -> Result<()> {
     if args.len() > 1 && args[1] == "-k" {
         print_message(&mut stdout, "Ready to print events:")?;
         print_events(&mut stdout)?;
-    } else {
-        let mut buffer = LineBuffer::new();
-        let mut history = VecDeque::with_capacity(HISTORY_SIZE);
-        let mut history_cursor = -1i64;
-        let mut has_history = false;
-        let mut cut_buffer = String::new();
+        terminal::disable_raw_mode()?;
+        println!();
+        return Ok(());
+    };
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -53,176 +53,130 @@ fn buffer_repaint(stdout: &mut Stdout, buffer: &LineBuffer, prompt_offset: u16)
...
     let mut stdout = stdout();
 
     terminal::enable_raw_mode()?;
+    // quick command like parameter handling
+    let args: Vec<String> = std::env::args().collect();
+    // if -k is passed, show the events
+    if args.len() > 1 && args[1] == "-k" {
+        print_message(&mut stdout, "Ready to print events:")?;
+        print_events(&mut stdout)?;
+    } else {
```
jonathandturner, line 94: One suggestion I think would help readability:

If `-k` just does a "print event loop", could we just return after the `print_events` above. This would keep you from having to use `else` and indent everything under it.

fdncred: ok. I added an early return instead of else.

jonathandturner:
Thanks!

Issue #18, opened 2021-03-16
Improving `read_line` output


jonathandturner:
From the twitch chat:

"boom_rang: In my mind the "exit" keyword is a concern of the REPL, read_line could return a enum where a string is one option and then Ctrl-C and Ctrl-D can have a special return each"

sholderbach:
Other special commands at the shell/line editor boundary would include `Ctrl+L` / `clear` which could break things for people that would want to use `reedline` in an other than shell REPL context like single command entry but are probably among the things people expect to work in a shell

jonathandturner:
@sholderbach would you mind filling an issue for clear? I'm going to close this issue as the original point about needing a way of mapping signals has been fixed

jonathandturner:
Closing this as fixed in #22

PR #22, opened 2021-03-16
Minimal enum to address #18

sholderbach:
Reuse outer Result for error handling only pass the special case
signals.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -48,9 +48,9 @@ pub struct Engine {
 }
 
 pub enum Signal {
-    SUCCESS(String),
-    SIGINT, // Typically Ctrl-c
-    EOF, // Typically Ctrl-d
+    Success(String),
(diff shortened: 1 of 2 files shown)


src/engine.rs
```diff
@@ -47,6 +47,12 @@ pub struct Engine {
     has_history: bool,
 }
 
+pub enum Signal {
+    SUCCESS(String),
+    SIGINT, // Typically Ctrl-c
+    EOF, // Typically Ctrl-d
```
jonathandturner, line 53: I think you could use `CtrlC` and `CtrlD` here instead of SIGINT/EOF. Helps with the readability and probably is a little more accurate if we're running on a platform without signals.

jonathandturner:
Just a couple nits. Should be go after that and resolving the conflict.

jonathandturner:
Looks good!

PR #23, opened 2021-03-17
Implement Deref for LineBuffer

jasonrhansen:
This removes the need for `get_buffer()` everywhere.


jonathandturner:
That's definitely cleaner. Much appreciated

PR #26, opened 2021-03-17
merged change for prompt

fdncred:
jt, i made this a draft because i was working on it while you were streaming. it may not be worthy but it's a quick and dirty prompt functionality. i originally had the struct as &str's with lifetimes but reverted to strings with clones so it may not be good enough.

it looks like this
image


jonathandturner:
@fdncred I'm totally okay with landing a fancy default prompt. Mind rebasing against the latest in main?

fdncred:
welp, it was working but not any longer with the latest code. with every character typed the prompt prints again. back to the drawing board.

PR #27, opened 2021-03-17
Implement basic history search

basile-henry:
This PR implements a basic reverse search through history mapped to `Ctrl-R`. It behaves in a similar way to bash/readline.


jonathandturner:
Very cool!

PR #29, opened 2021-03-18
added a slightly-fancy prompt

fdncred:
* added slightly fancy prompt with path on the left and datetime on the right
* todo - add some prompt configurability
* todo - fix 'attack of the clones' - replace with &str
* ran clippy, moved unwrap() to expect()
* had to revert some of the previous code, specifically the `position()?;` in `read_line` and on `buffer_paint`, in order to get the prompt to function appropriately
* have only tested on windows, crossing my fingers on mac and linux :)

would eventually like to get to this level of functionality (from fish shell)
```cpp
/// This is the main function for the screen putput library. It is used to define the desired
/// contents of the screen. The screen command will use its knowledge of the current contents of the
/// screen in order to render the desired output using as few terminal commands as possible.
///
/// \param s the screen on which to write
/// \param left_prompt the prompt to prepend to the command line
/// \param right_prompt the right prompt, or NULL if none
/// \param commandline the command line
/// \param explicit_len the number of characters of the "explicit" (non-autosuggestion) portion of
/// the command line
/// \param colors the colors to use for the comand line
/// \param indent the indent to use for the command line
/// \param cursor_pos where the cursor is
/// \param pager the pager to render below the command line
/// \param page_rendering to cache the current pager view
/// \param cursor_is_within_pager whether the position is within the pager line (first line)
void s_write(screen_t *s, const wcstring &left_prompt, const wcstring &right_prompt,
             const wcstring &commandline, size_t explicit_len,
             const std::vector<highlight_spec_t> &colors, const std::vector<int> &indent,
             size_t cursor_pos, pager_t &pager, page_rendering_t &page_rendering,
             bool cursor_is_within_pager);
```


jonathandturner:
Looks good to me on linux, let's land

Issue #28, opened 2021-03-18
Clear screen/Ctrl+L/Form Feed


sholderbach:
To emulate the behavior in bash/zsh etc. :

Maybe we consider a separation of concerns between core line editor and typical interactive shell terminal to make screen clearing optional for non shell/REPL usecases e.g. single command entry like vi `:`

On the side of the line editor `Engine`:
* Respond to `Ctrl+L`
* Keep current line_buffer state (including cursor position/`insertion_point`)

Helper to provide the functionality (MVP: call this in main):
* Print a screen worth of newlines (At least thats the behavior of 
FormFeed under gnome-terminal)
* Scroll new prompt to first line and repaint including cursor position

jonathandturner:
Fixed with #30

PR #30, opened 2021-03-18
Basic clear

sholderbach:
Addresses #28

Succeeds in providing a clear new screen for both the line editing and
reverse search mode.
Solution is conservative and prints more newlines than strictly
necessary as solutions with scrolling didn't seem to work as expected and exactly computing the length of the current written space is not trivial as multiline support is not fully settled yet.


src/engine.rs
```diff
@@ -81,6 +82,16 @@ pub fn print_crlf(stdout: &mut Stdout) -> Result<()> {
     Ok(())
 }
 
+pub fn clear_screen(stdout: &mut Stdout) -> Result<()> {
+    let (_, num_lines) = terminal::size()?;
+    for _ in 0..2 * num_lines {
+        stdout.queue(Print("\n"))?;
+    }
+    stdout.queue(MoveTo(0, 0))?;
+    stdout.flush()?;
```
jonathandturner, line 91: You might be able to run this crossterm command for a clear in one step: https://docs.rs/crossterm/0.19.0/crossterm/terminal/enum.ClearType.html

sholderbach: `ClearType::All` only blanks the currently visible screen, so you cannot scroll back up to the previous output as with bash et al.

jonathandturner: Today I learned :)

sholderbach:
With bash or zsh under gnome-terminal the amount of whitespace used by `Ctrl-L` to separate seems to be inversely proportional to how many lines were already used on the terminal.
image 
(zoomed out after the fact)

The `clear` command at least for me completely annihilates the scrollback buffer.

With my current draft you would have to scroll over one complete page of whitespace to get back to the old stuff.

jonathandturner:
@sholderbach - sounds good. Once we fix the conflict fixed we can land.

sholderbach:
Merging with the fancy prompt currently has some weird interaction with my requirement to repaint the line_buffer content at the beginning of `Engine::read_line` to match the `Ctrl+L` behavior. Will have to take a look into this.

sholderbach:
Combined with the fix from @jasonrhansen pulled from #31 and a slight reordering of my redraw, the prompt remains correct in all known single-line circumstances. 
The redraw logic right now seems somewhat convoluted with different functions and might warrant a refactor later.

jonathandturner:
Sounds good. Let's land, then see if we can refactor the prompt parts

PR #31, opened 2021-03-18
Redraw prompt indicator after history search

jasonrhansen:
The prompt indicator gets drawn over when searching the history so we
need to restore it when done searching.


jonathandturner:
Merged into #30

PR #32, opened 2021-03-19
Simplify prompt code and handle narrow terminals

jasonrhansen:
When there isn't enough room for both the left and right prompts, only
show the left one. Also truncate the left prompt if it's too wide.


jonathandturner:
Looks great.

@jasonrhansen - any interest in having commit access?

jasonrhansen:
@jonathandturner - thanks, but probably not necessary.

PR #34, opened 2021-03-20
Only append unique new lines to history

sholderbach:
To resemble the behavior of the typical shells check if the last command was identical and ignore repeat calls


jonathandturner:
Great, thanks!

PR #36, opened 2021-03-21
Remove Deref implementation and other refactorings

jasonrhansen:
`Deref` no longer makes sense for `LineBuffer` since it can have multiple lines.


jonathandturner:
Looks good

PR #38, opened 2021-04-10
Extract history management into struct

sholderbach:
* Encapsulates the up/down arrow browsing
* Still fully in-memory but easy to add file based constructors and
flushing
* No further integration with the statefulness of the Ctrl-R search
(yet)


jonathandturner:
Looks great!

PR #39, opened 2021-04-10
Simplify prints and newline after `read_line()`

sholderbach:
Make the return after a succesful entry part of `Reedline.read_line`'s
contract


jonathandturner:
Looks good

PR #43, opened 2021-04-24
added OS's clipboard

glella:
Copies the cut_buffer cuts to the OS clipboard using the "clipboard 0.5.0" crate.
Should work on Linux, macOS and Windows.
Pasting from the clipboard to reedline worked perfectly too (at least on macOS)
Thank you for doing the streams!


jonathandturner:
Thanks!

Issue #40, opened 2021-04-10
[ENH] Persist `History` in a file


sholderbach:
**Type:** Good first issue

Add support to load/store history to a file similar to `~/.bash_history`. Make the path reasonably configurable. 

**Bonus:** Only write the new lines of the current session and avoid races on the file.

PR #44, opened 2021-04-27
Support history file

sholderbach:
History has a new constructor to remember a file to read from / write to.
Writing is deferred to the Drop.
Binary supports history file via `REEDLINE_HISTFILE` env variable

Additional changes:

Inverted the ordering of the history VecDeque to simplify file IO and
history internals. This makes the search code slightly more brittle.

Addresses #40

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -88,8 +88,14 @@ impl Reedline {
         }
     }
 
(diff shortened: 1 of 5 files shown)


jonathandturner:
Looks good!

PR #45, opened 2021-04-27
Refactor Clipboard behind feature flag

sholderbach:
The clipboard crate breaks the CI in a non GUI context. Make use of it
optional by feature flag `system_clipboard`.
Clipboard could be changed during runtime based on trait.

Additional fix: read back from system clipboard


jonathandturner:
Good catch! Thanks for fixing that!

PR #47, opened 2021-05-02
Basic API docs to give an overview of the current state

sholderbach:
Includes some doc tests for the `History`. Code examples for `Reedline` are still very bare bones. 

Some of the APIs probably subject to significant change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "getrandom"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


sholderbach:
@jonathandturner to check my documentation of the line buffer backend I covered most of it in doctests. I later realized we don't actually have to expose the `LineBuffer`. Making it private would break doctests. What is your take on clean API surface vs. easier onboarding of contributors at this stage?

jonathandturner:
@sholderbach we should go ahead and make it public if it's an important part of the API

sholderbach:
It's currently accidentally already public but has no direct public interaction with the read_line part. I would tend to merge the docs first and go about hiding the implementation details later. I think the testing story has to evolve over time some more.

PR #49, opened 2021-06-10
Add dynamic prompt color

mzanrosso:
Added support for automatically changing the prompt color.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -19,8 +19,6 @@ use std::{
     time::Duration,
 };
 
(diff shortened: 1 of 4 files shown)


sholderbach:
Looks good!

mzanrosso:
I want detail better since I was in a bit of a hurry before; I created this pr so anyone can with their own interpretation of the prompt trait change the color that was previously by default to blue.

sholderbach:
Do you want to make additional changes before merging?

mzanrosso:
No thanks, for me it's fine

PR #54, opened 2021-06-13
Various proposed refactors

sholderbach:
* Small cleanup stuff
* Tests for the new up/down prefix search

Potentially opinionated:

* Make the reset of the history cursor to now part of the `History.append()` contract
* To slim down `src/engine.rs` extract the `EditCommand` and `Signal` enums into a new module


jonathandturner:
Looks good!

PR #56, opened 2021-06-14
syntax highlighting fileio

fdncred:
This is a first pass at reading syntax highlighting json files in order to color the command line syntax later. The end goal of this was to get something out there to tweak on while being somewhat compatible with what nushell likes in terms of coloring cratest and json crates. This code does nothing as is with `dead_code` sprinkled in liberally but there is a commented out `main` function as well as a `test` to see how to use it.


assets/sample_docs/syntax.json
```diff
@@ -0,0 +1,9 @@
+{
+  // some comment
+  "internal_command": "#ff0000|#ffffff|b",
+  // some other comment
```
jonathandturner, line 4: nit: json doesn't support comments (technically)

assets/sample_docs/syntax.json
```diff
@@ -0,0 +1,9 @@
...
+  "external_command": "#ff0000|#ffffff|n",
+  "sub_command": "#ff0000|#ffffff|bli",
+  "keyword": "|#ffffff|u",
+  "punctuation": "#ff0000||s"
```
jonathandturner, line 8: Thinking through this a bit, now that I see it, would it make sense to split these strings into their parts?

```
"sub_command": {
  "bg": "#ff0000",
  "fg": "#ffffff",
  "attr": "n"
}
```

I think mostly people would copy paste to add new ones

assets/sample_docs/syntax.json
fdncred: it does with our json crate

assets/sample_docs/syntax.json
fdncred: i was trying to keep it on one line per key and not have a sprawling json file. i tried that before and it was just a mess and hard to keep up with. if you think about most configs it key value but our will offer much more with fg anf bg and over a half dozen attributes. let's try to keep it small.

fdncred: i updated all this mess just because I like you. :)

jonathandturner:
That definitely looks better, even if it's bigger :slightly_smiling_face: 

So the next step for syntax highlighting is that we need to have a classifier return back the spans and the class of each of the spans. Then we match up the class with the color using this logic.

Sounds good. Thanks!

PR #71, opened 2021-06-21
Extract operations on buffer into line buffer

sherubthakur:
https://github.com/jonathandturner/reedline/issues/61#issuecomment-864374303

The basic idea here is to pull out the functionality of editing into `line_buffer`.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -293,7 +293,7 @@ impl Reedline {
 
     fn insert_char(&mut self, c: char) {
         let insertion_point = self.line_buffer.insertion_point();
-        self.line_buffer.insert_char(insertion_point, c);
+        self.line_buffer.insert_char_at(insertion_point, c);
     }
 
     fn backspace(&mut self) {
@@ -406,7 +406,8 @@ impl Reedline {
     fn insert_cut_buffer(&mut self) {
         let insertion_offset = self.insertion_point().offset;
         let cut_buffer = self.cut_buffer.get();
-        self.line_buffer.insert_str(insertion_offset, &cut_buffer);
+        self.line_buffer
+            .insert_str_at(self.insertion_point(), &cut_buffer);
         self.set_insertion_point(insertion_offset + cut_buffer.len());
     }
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -392,11 +392,8 @@ impl Reedline {
     }
 
     fn insert_cut_buffer(&mut self) {
-        let insertion_offset = self.insertion_point().offset;
         let cut_buffer = self.cut_buffer.get();
-        self.line_buffer
-            .insert_str_at(self.insertion_point(), &cut_buffer);
-        self.set_insertion_point(insertion_offset + cut_buffer.len());
+        self.line_buffer.insert_str(&cut_buffer);
     }
 
     fn uppercase_word(&mut self) {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -297,20 +297,11 @@ impl Reedline {
     }
 
     fn backspace(&mut self) {
-        let left_index = self.line_buffer.grapheme_left_index();
-        let insertion_offset = self.insertion_point().offset;
-        if left_index < insertion_offset {
-            self.clear_range(left_index..insertion_offset);
-            self.set_insertion_point(left_index);
-        }
+        self.line_buffer.delete_left_grapheme();
     }
 
     fn delete(&mut self) {
-        let right_index = self.line_buffer.grapheme_right_index();
-        let insertion_offset = self.insertion_point().offset;
-        if right_index > insertion_offset {
-            self.clear_range(insertion_offset..right_index);
-        }
+        self.line_buffer.delete_right_grapheme();
     }
 
     fn backspace_word(&mut self) {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -313,8 +313,6 @@ impl Reedline {
 
     fn clear(&mut self) {
         self.line_buffer.clear();
-        // Note: This should be redundant
-        self.set_insertion_point(0);
     }
 
     fn append_to_history(&mut self) {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -292,8 +292,7 @@ impl Reedline {
     }
 
     fn insert_char(&mut self, c: char) {
-        let insertion_point = self.line_buffer.insertion_point();
-        self.line_buffer.insert_char_at(insertion_point, c);
+        self.line_buffer.insert_char(c)
     }
 
     fn backspace(&mut self) {
@@ -826,10 +825,7 @@ impl Reedline {
                                 };
                                 if self.maybe_wrap(terminal_size.0, line_start, c) {
                                     let (original_column, original_row) = position()?;
-                                    self.run_edit_commands(&[
-                                        EditCommand::InsertChar(c),
-                                        EditCommand::MoveRight,
-                                    ]);
+                                    self.run_edit_commands(&[EditCommand::InsertChar(c)]);
                                     self.buffer_paint(prompt_offset)?;
 
                                     let (new_column, _) = position()?;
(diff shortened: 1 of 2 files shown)


src/line_buffer.rs
```diff
@@ -135,16 +141,28 @@ impl LineBuffer {
...
+        let pos = self.insertion_point();
+        self.lines[pos.line].insert(pos.offset, c);
+        self.move_right();
+    }
+
     /// Insert a single character at the given cursor postion
-    pub fn insert_char(&mut self, pos: InsertionPoint, c: char) {
+    pub fn insert_char_at(&mut self, pos: InsertionPoint, c: char) {
         self.lines[pos.line].insert(pos.offset, c)
     }
```
sherubthakur, line 153: I believe we don't really have a use-case where the newly created `insert_char` is not sufficient, we can remove this method from the public interface and only use that.

Reasoning: 
It leads to an inconsistent internal state and the user has to be cognizant about making sure that the insertion point is updated properly. I think we should only expose this sort of API if there is something we can't do with the high-level API that we have.

src/line_buffer.rs
```diff
@@ -135,16 +141,28 @@ impl LineBuffer {
...
+        let pos = self.insertion_point();
+        self.lines[pos.line].insert_str(pos.offset, string);
+        self.insertion_point.offset = pos.offset + string.len();
+    }
+
     /// Insert `&str` at the `idx` position in the current line.
     ///
     /// TODO: Check unicode validation
-    pub fn insert_str(&mut self, idx: usize, string: &str) {
-        self.lines[self.insertion_point.line].insert_str(idx, string)
```
sherubthakur, line 147: I believe we don't really have a use-case where the newly created `insert_str` is not sufficient, we can remove this method from the public interface and only use that.

Reasoning:
It leads to an inconsistent internal state and the user has to be cognizant about making sure that the insertion point is updated properly. I think we should only expose this sort of API if there is something we can't do with the high-level API that we have.

src/line_buffer.rs
```diff
@@ -194,13 +212,280 @@ impl LineBuffer {
...
+    }
+
+    #[test]
+    fn delete_left_grapheme_on_an_empty_buffer_is_a_no_op() {
+        let mut line_buffer = buffer_with("");
+        line_buffer.delete_left_grapheme();
+
+        let expected_line_buffer = buffer_with("");
+
+        assert_eq!(expected_line_buffer, line_buffer);
```
sherubthakur, line 410: https://docs.rs/pretty_assertions/0.7.2/pretty_assertions/index.html can prettify the diffs. I think it would be a nice lib to add.

src/line_buffer.rs
```diff
@@ -194,13 +212,280 @@ impl LineBuffer {
...
+    #[test]
+    fn delete_right_grapheme_works_with_emojis() {
+        let mut line_buffer = buffer_with("This is a test 😊");
+        line_buffer.move_left();
+        line_buffer.delete_right_grapheme();
+
+        let expected_line_buffer = buffer_with("This is a test ");
+
+        assert_eq!(expected_line_buffer, line_buffer);
+    }
```
sherubthakur, line 433: In python we have things like `pytest.paramertize` which can make this sort of testing super easy. I am not sure what is the idomatic way to do this in rust. I was looking at libs like `rstest`,`quickcheck`,`proptest` etc that can help in this regard but they seem either too limited or too complex. Let me know your thoughts?

src/line_buffer.rs
```diff
@@ -194,13 +212,280 @@ impl LineBuffer {
...
+        line_buffer.move_word_left();
+        line_buffer.uppercase_word();
+
+        let expected_line_buffer = buffer_with("This is a TEST");
+
+        assert_eq!(expected_line_buffer, line_buffer);
+    }
+}
+
 #[test]
```
sherubthakur, line 489: I think we can remove this now.

sherubthakur (commented):
I have only added basic tests for now as I want to know what you folks think about this. If it looks alright then I can add more tests to this.

src/line_buffer.rs
sholderbach: I'm not aware of an established fixture-based testing framework. I tend to formulate my tests for a whole aspect and have sometimes more than one assert in my rust tests if it makes sense. Other than that, functions in the test module are fair game.

src/engine.rs
```diff
@@ -293,39 +293,28 @@ impl Reedline {
...
     fn delete_word(&mut self) {
-        let right_word_index = self.line_buffer.word_right_index();
-        self.clear_range(self.insertion_point().offset..right_word_index);
+        self.line_buffer.delete_word_right();
     }
 
     fn clear(&mut self) {
         self.line_buffer.clear();
+        // Note: This should be redundant
         self.set_insertion_point(0);
```
sholderbach, line 318: Make sure and remove the unnecessary code

src/engine.rs
```diff
@@ -293,39 +293,28 @@ impl Reedline {
 
     fn insert_char(&mut self, c: char) {
         let insertion_point = self.line_buffer.insertion_point();
-        self.line_buffer.insert_char(insertion_point, c);
+        self.line_buffer.insert_char_at(insertion_point, c);
```
sholderbach, line 296: Replace with stateful `insert_char`

src/line_buffer.rs
```diff
@@ -194,13 +212,280 @@ impl LineBuffer {
...
 /// Match any sequence of characters that are considered a word boundary
 fn is_word_boundary(s: &str) -> bool {
     !s.chars().any(char::is_alphanumeric)
 }
 
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    fn buffer_with(content: &str) -> LineBuffer {
```
sholderbach, line 341: Maybe also have a version to test cursor/insertion point movements

sholderbach (commented):
Do you want to rebase this on #62 with `EditEngine` or keep the other PR as a draft for a little bit longer? Would be fine to merge into `main` as well.

Regarding tests, in 08ff9c1 I wrote some doctests for most of the functions, but didn't end up merging that upstrream as doctests don't run on private parts and the API would change anyway. But might have some inspiration for what we might test in the future.

src/line_buffer.rs
sherubthakur: I will try to limit the scope of this PR to this change only. Will add another PR that adds this dependency.

src/line_buffer.rs
```diff
@@ -57,6 +58,11 @@ impl LineBuffer {
         &self.lines[self.insertion_point.line]
     }
 
+    /// Output the current line in the multiline buffer
+    pub fn current_line(&self) -> &str {
+        &self.lines[self.insertion_point.line]
+    }
+
```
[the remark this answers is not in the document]
sherubthakur: Will remove this.

sherubthakur:
> Do you want to rebase this on #62 with EditEngine or keep the other PR as a draft for a little bit longer? Would be fine to merge into main as well.

I was thinking that PR to just be a sort of discussion point where we can try out a few ideas and not necessarily get it merged. If we try to get that merged it might block other contributors, which is something I would like to avoid. We can extract the bits that make sense there into smaller PRs that way we make continuous progress. Moreover, the more I think about that, the less convinced I am about the inheritance hierarchy that I have created there.
So, I would prefer we get this one merged first.

src/line_buffer.rs
sherubthakur: Right now there is not really any place we are making use of this internally so, I am removing this for now. If there is a use for this we can re-add this. Is that ok?

src/line_buffer.rs
sherubthakur: For now, I will keep these as-is. I will update if folks decide that we want to include any of these.

sholderbach:
@sherubthakur Ready to merge?

sholderbach (approved):

sherubthakur:
> @sherubthakur Ready to merge?

Yup

PR #72, opened 2021-06-21
fix: prompt_origin and prompt_offset issues

ahkrr:
* Fix the infinite scrolling with cursor in the last line. 
* Use the return values of `.full_repaint()`
* Handle more than one linewrap
* Hide cursor on `.full_repaint()`
* Handle resizing better (not all cases are handled)


fdncred:
I really like the idea of hiding the cursor, doing work, showing cursor. I think it'll give a much smoother feel to reedline.

jonathandturner:
@ahkrr - thanks!

PR #73, opened 2021-06-23
Add pretty_assertions lib

sherubthakur:
This improves the test diff output a lot and consequently makes going through diffs a joy! Attaching a screenshot to showcase the difference this will make.

### Before

scrot_2021-06-23-162005

### After

scrot_2021-06-23-161942

ref: https://github.com/jonathandturner/reedline/pull/71#discussion_r655307424

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,6 +10,15 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jonathandturner:
Looks good

Issue #76, opened 2021-06-23
odd moving of text when on bottom row of terminal


fdncred:
this is hard to describe but it appears that in Windows using Windows Terminal, when you're on the bottom row of the terminal and you type something like 'now is the time for all good men' and hit enter, it briefly moves the text ~10 characters to the right, then accepts the text and prints it as normal. It's really odd and hard to explain.

I reproduce this by doing:
1. clear
2. cargo clean
3. cargo run
4. ls <enter>
5. now is the time for all good men<enter>
6. and watch it shift and then accept

i've tried to capture it with screentogif but i'm not able to.

ok, here's what it's doing. i changed screentogif to run at 60 fps and i caught the error. here are the 3 frames.

frame 1 - when i typed in some (any) text
![image](https://user-images.githubusercontent.com/343840/123157437-28070700-d430-11eb-81fc-8bf3c0bb9595.png)

frame2 - the flashing that i see, which i characterized above as ~10 chars to the right. it actually is moving the text to the right because it's prefixed by `Our buffer: `
![image](https://user-images.githubusercontent.com/343840/123157567-4d941080-d430-11eb-92e9-50dd81518c32.png)

frame3 - when it accepts the text
![image](https://user-images.githubusercontent.com/343840/123157617-61d80d80-d430-11eb-94ad-bd6f6ac30ac7.png)

sholderbach:
Triage: Has this been resolved on Windows by the fixes from #77?

stormasm:
Same exact issue on the Mac as described in #132

fdncred:
@sholderbach I haven't tested recently. Feel free to close this issue if you think it's fixed by #77.

stormasm:
@fdncred @sholderbach this issue is not fixed on the mac...

fdncred:
oh, @stormasm, I thought you were saying that issue #132 wasn't fixed on the mac. Sorry.

stormasm:
@fdncred yep ---- sorry about that ---- I should have been more clear...

sholderbach:
Probably out of date, painting logic and handling of the bottom of the screen case was changed significantly with #217

PR #77, opened 2021-06-23
fix: overwriting of line #76

ahkrr:
This happened, because the `prompt` would be drawn at the `origin`, which was determined to be at the `position()?`, which was in the previous line. 
Now, move one line downwards before determining the `origin`.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -768,14 +768,18 @@ impl Reedline {
 
         let mut terminal_size = terminal::size()?;
 
-        self.stdout.queue(Print("\r\n"))?.flush()?;
-
         let mut prompt_origin = {
             let (column, row) = position()?;
-            if row + 1 == terminal_size.1 {
-                (column, row.saturating_sub(1))
+            if (column, row) == (0, 0) {
+                (0, 0)
+            } else if row + 1 == terminal_size.1 {
+                self.stdout.queue(Print("\r\n\r\n"))?.flush()?;
+                (0, row.saturating_sub(1))
+            } else if row + 2 == terminal_size.1 {
+                self.stdout.queue(Print("\r\n\r\n"))?.flush()?;
+                (0, row)
             } else {
-                (column, row)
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -768,6 +768,8 @@ impl Reedline {
 
         let mut terminal_size = terminal::size()?;
 
+        self.stdout.queue(Print("\r\n"))?.flush()?;
```
jonathandturner, line 771: Do we always want to send a crlf? I was wondering if we may just want to see that we're at the bottom of the screen and scroll the screen up a line if so.

ahkrr:
While thinking about this issue, I noticed that maybe `reedlines` use in `main.rs` should result in behavior like this
```
/home/hk/Dokumente/rus_test/reedline      06/24/2021 11:32:24 PM
〉ls
Our buffer: ls

/home/hk/Dokumente/rus_test/reedline      06/24/2021 11:32:26 PM
〉
```
because in `main.rs `
`line_editor.print_line(&format!("Our buffer: {}", buffer))?;`
is called, which prints a `\n` at the end placing the cursor in the next row. 
If an "application" would print a line without a trailing newline, this line would have been eaten by reedline previously.
The new changes now lead to the output looking like the example above.

ahkrr:
Or should it be detected, if the position is already at (0, some_row) -> work from here, 
and if the position is ( column > 0, some_row) -> start in the next line

ahkrr:
``` bash
oh-my-zsh
> printf "jo"
jo%
oh-my-zsh
> printf "jo\n"
jo
oh-my-zsh
>
```

jonathandturner:
@ahkrr yeah, we could put the logic in to detect if we're not in the 0th column. Do you want to do that in a separate PR?

ahkrr:
Lets do that in a separate PR.

jonathandturner:
sounds good

PR #79, opened 2021-06-24
Improve Prompt trait and it's handling

Urgau:
Move out of `Box<dyn Prompt>` by using `&dyn Prompt`.
Also use `Cow<str>` instead of `String` for the `Prompt` return types.

Both of this changes permit the use of reference but most importantly, it avoids unnecessary clones.


jonathandturner:
@Urgau - yeah, this definitely cleans it up. Can you merge from the latest main? We just landed a few updates to how prompts are rendered.

Urgau:
Rebased.

jonathandturner:
Thanks!

PR #80, opened 2021-06-28
Extract screen painting stuff into a separate struct

sherubthakur:
From https://github.com/jonathandturner/reedline/pull/62 > https://github.com/jonathandturner/reedline/pull/62#issuecomment-864614387

> Extracting the Painter seems to absolute clean up some clutter and makes changes to output code centralized. Full endorsement from my side.
Future thought: Maybe the Painter could also do the bookkeeping on the terminal positions regarding where the current prompt started/where the current offsets are (in this case maybe named TerminalOutput) (need to be called on resize)
With a good design it is probably reasonable to hand out exclusive references of the Painter to (plugin) components.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -2,6 +2,7 @@ use crate::{
     clip_buffer::{get_default_clipboard, Clipboard},
     default_emacs_keybindings,
     keybindings::{default_vi_insert_keybindings, default_vi_normal_keybindings, Keybindings},
+    painter::Painter,
     prompt::{PromptEditMode, PromptHistorySearch, PromptHistorySearchStatus, PromptViMode},
     Prompt,
 };
@@ -12,19 +13,12 @@ use crate::{
 };
 use crate::{EditCommand, EditMode, Signal, ViEngine};
 use crossterm::{
-    cursor,
-    cursor::{position, MoveTo, MoveToColumn, RestorePosition, SavePosition},
+    cursor::position,
     event::{poll, read, Event, KeyCode, KeyEvent, KeyModifiers},
-    style::{Color, Print, ResetColor, SetForegroundColor},
-    terminal::{self, Clear, ClearType},
-    QueueableCommand, Result,
+    terminal, Result,
 };
 
-use std::{
-    collections::HashMap,
-    io::{stdout, Stdout, Write},
-    time::Duration,
-};
+use std::{collections::HashMap, io::stdout, time::Duration};
 
 /// Line editor engine
 ///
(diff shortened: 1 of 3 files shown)


src/engine.rs
```diff
@@ -771,18 +707,18 @@ impl Reedline {
             if (column, row) == (0, 0) {
                 (0, 0)
             } else if row + 1 == terminal_size.1 {
-                self.stdout.queue(Print("\r\n\r\n"))?.flush()?;
+                self.painter.paint_carrige_return()?;
```
sherubthakur, line 710: I am unsure about this interface. Any recommendations?

src/painter.rs
```diff
@@ -0,0 +1,193 @@
...
+};
+
+use std::io::{Stdout, Write};
+
+pub struct Painter {
+    // Stdout
+    stdout: Stdout,
+}
+
+impl Painter {
```
sherubthakur, line 19: Most functions queue up the operations. Some also flush. Which makes the interface a bit inconsistent but satisfy the requirements more closely. What's other's take on 
- completely queue-based interface?
- flush after every function
- or a combination (proper naming)

src/painter.rs
```diff
@@ -0,0 +1,193 @@
...
+        self.stdout.queue(cursor::MoveTo(column, row))?;
+
+        Ok(())
+    }
+
+    /// Queue the complete prompt to display including status indicators (e.g. pwd, time)
+    ///
+    /// Used at the beginning of each [`Reedline::read_line()`] call.
+    pub fn queue_prompt(&mut self, prompt: &dyn Prompt, prompt_mode: PromptEditMode) -> Result<()> {
+        let (screen_width, _) = terminal::size()?;
```
sherubthakur, line 34: Can also inject this? Not sure if this optimization is worth extending the interface that we provide.

src/engine.rs
jonathandturner: I think that's fine

src/painter.rs
jonathandturner: We'll probably need to do some measurements to see how bad flushing a lot is. My instinct is to say that we should only flush when we have to, so before reading an input (including terminal size), and not flush between normal writes to the screen. Probably need to play with a bit to find what feels best.

src/painter.rs
jonathandturner: Reading the terminal::size is pretty slow, as behind the scenes crossterm has to write out an ansi sequence, wait, and read back the answer. We want to remember the terminal size we saw before and reuse it

jonathandturner (commented):
I like the refactor - this should help a lot

Had a comment about making sure we use the cached terminal size, but other than that I think this should be ready.

sherubthakur:
@jonathandturner made the change you asked for.

jonathandturner:
Looks good!

PR #81, opened 2021-06-30
Bugfix after recent changes in prompt module

mzanrosso:
Simple bug fix with use declaration and public scope, it should be straightforward to approve, thanks.


jonathandturner:
Looks good, thanks!

PR #82, opened 2021-06-30
Feature buffer color

mzanrosso:
Hey everyone, I added this new feature that allows you to change the color of the buffer if the command is recognized or not, obviously I have implemented it so that it is backwards compatible adding default values and customizable for those who have this need. 

I will post here a little gif so you can understand better.

example

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -4,7 +4,7 @@ use crate::{
     keybindings::{default_vi_insert_keybindings, default_vi_normal_keybindings, Keybindings},
     painter::Painter,
     prompt::{PromptEditMode, PromptHistorySearch, PromptHistorySearchStatus, PromptViMode},
-    Prompt,
+    DefaultHighlighter, Highlighter, Prompt,
 };
 use crate::{history::History, line_buffer::LineBuffer};
 use crate::{
@@ -13,7 +13,8 @@ use crate::{
 };
 use crate::{EditCommand, EditMode, Signal, ViEngine};
 use crossterm::{
-    cursor::position,
+    cursor,
+    cursor::{position, MoveLeft, MoveTo, MoveToColumn, RestorePosition, SavePosition},
     event::{poll, read, Event, KeyCode, KeyEvent, KeyModifiers},
     terminal, Result,
 };
@@ -65,6 +66,9 @@ pub struct Reedline {
 
     // Vi normal mode state engine
     vi_engine: ViEngine,
+
+    // Manage buffer highlights
+    highlighter: Box<dyn Highlighter>,
 }
 
 impl Default for Reedline {
@@ -95,7 +99,13 @@ impl Reedline {
             need_full_repaint: false,
             partial_command: None,
             vi_engine: ViEngine::new(),
+            highlighter: Box::new(DefaultHighlighter::default()),
         }
+    }
+
+    pub fn with_highlighter(mut self, highlighter: Box<dyn Highlighter>) -> Reedline {
+        self.highlighter = highlighter;
+        self
     }
 
     pub fn with_history(
@@ -618,6 +628,9 @@ impl Reedline {
     ///
     /// Requires coordinates where the input buffer begins after the prompt.
     fn buffer_paint(&mut self, prompt_offset: (u16, u16)) -> Result<()> {
+        let new_index = self.insertion_point().offset;
+        let insertion_line = self.insertion_line().to_string();
+        let offset = insertion_line.len() - new_index;
         // Repaint logic:
         //
         // Start after the prompt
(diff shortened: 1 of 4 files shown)


mzanrosso:
The PR is not ready yet, I would like also some suggestion on how to improve this functionality.

src/engine.rs
```diff
@@ -71,6 +74,11 @@ pub struct Reedline {
 
     // Vi normal mode state engine
     vi_engine: ViEngine,
+
+    // To manage line highlight hint
+    buffer_success_color: Color,
```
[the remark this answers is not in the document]
mzanrosso: Thanks, I'm working on a better algorithm, after completed that I will go with your suggestion.

mzanrosso:
Logic update to consider command and sub-command

example

jonathandturner:
@mzanrosso - great! It's nice to start to see some syntax highlighting

Something I had in my notes for highlighting it to come up with some trait that a highlighter would implement. Something like:

```
trait Highlighter {
  fn highlight(&self, text: &str) -> Vec<(Span, Style)>
}
```

Where Style is something like what @fdncred mentioned with being able to do more than colours.

An alternative one is closer to what Rustyline uses:
```
trait Highlighter {
  fn highlight(&self, text: &str) -> String
}
```
Where the output is the ANSI styled version of the string you pass in.

Either way, it lets anyone using reedline implement that trait, and then nushell could call into the highlighter for doing the paint of the user's input.

Are you interested in adding that to this PR? We could also work on it as a follow-up PR (maybe on stream?) as well.

mzanrosso:
Hey @jonathandturner, thanks for your comment; yes I'm very interested in upgrade this functionality; let me see if I can work on it directly on this pr.

mzanrosso:
@jonathandturner, @fdncred
Try to have a look now, thanks.

src/highlighter.rs
```diff
@@ -0,0 +1,80 @@
...
+    nu_ansi_term::{Color, Style},
+    std::{borrow::Cow, io::Write},
+};
+
+pub static DEFAULT_BUFFER_MATCH_COLOR: Color = Color::Green;
+pub static DEFAULT_BUFFER_NEUTRAL_COLOR: Color = Color::White;
+pub static DEFAULT_BUFFER_NOTMATCH_COLOR: Color = Color::Red;
+
+pub trait Highlighter {
+    fn highlight_external_command<'l>(&self, line: &'l str) -> Cow<'l, str>;
```
jonathandturner, line 11: how about `highlight`? Making it more general so anyone could implement Highlighter and configure it to highlight other things also

src/engine.rs
```diff
@@ -101,9 +104,15 @@ impl Reedline {
             need_full_repaint: false,
             partial_command: None,
             vi_engine: ViEngine::new(),
+            highlighter: Box::new(DefaultHighlighter::default()),
         }
     }
 
+    pub fn with_highlighter(mut self, external_commands: Vec<String>) -> Reedline {
```
jonathandturner, line 111: Nice! One small tweak: instead of taking in the external_commands here, take in the whole Highlighter.

Then, when you build a Reedline, you have more control over what Highlighter is used.

mzanrosso: Thank you;  initially i wanted to do it too but I had some compilation errors; if you can, try to propose your change, thanks.

src/highlighter.rs
mzanrosso: The initial idea was to have a function for each type of highlighting but we can discuss it.

src/engine.rs
jonathandturner: @mzanrosso - not sure how best to propose changes, but I did a branch based on yours: https://github.com/jonathandturner/reedline/tree/feature/buffer-colors

maybe you can compare the two and see the few spots I changed?

mzanrosso: That's okay, I usually meant a diff or a git prepare-patch but no problem, i will compare it asap.

mzanrosso: @jonathandturner  
I have implemented your changes, is there still something we can improve?

jonathandturner:
Looks good!

mzanrosso:
@jonathandturner it wasn't really ready for merge since the last commit change a bit the paint logic, let me fix this with another PR

jonathandturner:
@mzanrosso - I noticed that after the merge. I've got a new PR up which I think fixes it

mzanrosso:
@jonathandturner no problem

src/highlighter.rs
```diff
@@ -0,0 +1,80 @@
...
+
+pub trait Highlighter {
+    fn highlight<'l>(&self, line: &'l str) -> Cow<'l, str>;
+}
+
+pub struct DefaultHighlighter {
+    external_commands: Vec<String>,
+    match_color: Color,
+    notmatch_color: Color,
+    neutral_color: Color,
```
fdncred, line 18: I'm still of the opinion that these should be `NuStyle`s as [defined here](https://github.com/jonathandturner/reedline/blob/54215568952db347f8bdffdcf830d0fa59fa4022/src/syntax_highlighting_fileio.rs#L10-L14) instead of `Color`. Perhaps in a later PR we can update this and read these colors from a config file using `syntax_highlighting_fileio.rs`.

PR #85, opened 2021-07-01
Highlighter in painter

sherubthakur:
As per our discussion - https://discord.com/channels/601130461678272522/855886335980994600/860066323819003924


jonathandturner:
@sherubthakur - I think this looks good. Some recent changes look like they conflict. Once those are fixed I think this one is good

sherubthakur:
@jonathandturner done

PR #87, opened 2021-07-02
Vi normal mode parser returning separated actions/motions.

sholderbach:
Basic parser for the mayor operations in normal mode (some oddballs included)

Provides:

- [x] Enums with the core functionalities expected for normal mode (some could be translated to visual mode)
- [x] Tests of the basic parser
- [x] Supports abort on invalid sequences as well as recognition of incomplete sequences

Not provided:

- [ ] Key remapping
- [ ] Parsing logic to dispatch modifying letters used both for actions and movements like more complex `g` or `z` based sequences. Parsing is just using look ahead of one char and rejects either invalid actions or invalid moves thus no `gg` at the same time as `gu` etc.
- [ ] Unicode support for char replacement or char search
- [ ] Logic for the key sequence buffer.
- [ ] Implementation of the action or movement based editing

### DIFF
--- a/src/edit_mode/vi/parser.rs
+++ b/src/edit_mode/vi/parser.rs
@@ -2,84 +2,227 @@ use std::{iter::Peekable, str::Bytes};
 
 #[derive(Debug, PartialEq, Eq)]
 enum Motion {
-    Word,
+    NoMove,
+    LeftChar,
+    RightChar,
+    Up,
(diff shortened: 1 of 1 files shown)


sholderbach:
Superseded by more complete #211 by @elferherrera. May still be used as inspiration for test cases etc.

PR #90, opened 2021-07-05
[Improve] highlighter logic for external commands adding case insensitive feature

mzanrosso:
Manage the case when the buffer line contains for some reasons some uppercase character.

### DIFF
--- a/src/highlighter.rs
+++ b/src/highlighter.rs
@@ -20,17 +20,18 @@ pub struct DefaultHighlighter {
 impl Highlighter for DefaultHighlighter {
     fn highlight(&self, line: &str) -> StyledText {
         let mut styled_text = StyledText::new();
+        let lowercase_line = line.to_lowercase();
 
         if self
             .external_commands
             .clone()
             .iter()
-            .any(|x| line.contains(x))
+            .any(|x| lowercase_line.contains(x))
         {
             let matches: Vec<String> = self
                 .external_commands
                 .iter()
-                .filter(|c| line.contains(*c))
+                .filter(|c| lowercase_line.contains(*c))
                 .map(|c| c.to_string())
                 .collect();
             let longest_match = matches.iter().fold("".to_string(), |acc, item| {
@@ -40,16 +41,22 @@ impl Highlighter for DefaultHighlighter {
                     acc
                 }
             });
-            let buffer_split: Vec<&str> = line.splitn(2, &longest_match).collect();
+            let lowercase_buffer_split: Vec<&str> =
+                lowercase_line.splitn(2, &longest_match).collect();
+            let before_longest_match = line[..lowercase_buffer_split[0].len()].to_string();
+            let after_longest_match =
+                line[(lowercase_buffer_split[0].len() + longest_match.len())..].to_string();
 
+            styled_text.push((Style::new().fg(self.neutral_color), before_longest_match));
             styled_text.push((
-                Style::new().fg(self.neutral_color),
-                buffer_split[0].to_string(),
+                Style::new().fg(self.match_color),
+                line[lowercase_buffer_split[0].len()
+                    ..(lowercase_buffer_split[0].len() + longest_match.len())]
+                    .to_string(),
             ));
-            styled_text.push((Style::new().fg(self.match_color), longest_match));
             styled_text.push((
                 Style::new().bold().fg(self.neutral_color),
-                buffer_split[1].to_string(),
+                after_longest_match,
             ));
         } else if !self.external_commands.is_empty() {
             styled_text.push((Style::new().fg(self.notmatch_color), line.to_string()));
--- a/src/highlighter.rs
+++ b/src/highlighter.rs
@@ -15,23 +15,27 @@ pub struct DefaultHighlighter {
     match_color: Color,
     notmatch_color: Color,
     neutral_color: Color,
+    case_sensitive: bool,
 }
 
 impl Highlighter for DefaultHighlighter {
     fn highlight(&self, line: &str) -> StyledText {
         let mut styled_text = StyledText::new();
-        let lowercase_line = line.to_lowercase();
+        let updated_line = match self.case_sensitive {
+            true => line.to_string(),
+            false => line.to_lowercase(),
+        };
 
         if self
             .external_commands
             .clone()
             .iter()
-            .any(|x| lowercase_line.contains(x))
+            .any(|x| updated_line.contains(x))
         {
             let matches: Vec<String> = self
                 .external_commands
                 .iter()
-                .filter(|c| lowercase_line.contains(*c))
+                .filter(|c| updated_line.contains(*c))
                 .map(|c| c.to_string())
                 .collect();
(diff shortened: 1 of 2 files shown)


src/highlighter.rs
```diff
@@ -40,16 +41,17 @@ impl Highlighter for DefaultHighlighter {
...
+            let before_longest_match = line[..lowercase_buffer_split[0].len()].to_string();
+            let after_longest_match =
+                line[(lowercase_buffer_split[0].len() + longest_match.len())..].to_string();
 
-            styled_text.push((
-                Style::new().fg(self.neutral_color),
-                buffer_split[0].to_string(),
-            ));
+            styled_text.push((Style::new().fg(self.neutral_color), before_longest_match));
             styled_text.push((Style::new().fg(self.match_color), longest_match));
```
mzanrosso, line 51: Here I left longest_match but we can discuss if its better to put the real string with some uppercase character or not

jonathandturner: Do you have an example of what this will look like?

mzanrosso:
```
From 1a882553028b51f11dca8676f0e0aabccd87724b Mon Sep 17 00:00:00 2001
From: Marco Zanrosso <mzanrosso@storvix.eu>
Date: Mon, 5 Jul 2021 23:09:12 +0200
Subject: [PATCH] Alternative output for the matched string

---
 src/highlighter.rs | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/src/highlighter.rs b/src/highlighter.rs
index 9dd9ed7..96b1141 100644
--- a/src/highlighter.rs
+++ b/src/highlighter.rs
@@ -48,7 +48,12 @@ impl Highlighter for DefaultHighlighter {
                 line[(lowercase_buffer_split[0].len() + longest_match.len())..].to_string();

             styled_text.push((Style::new().fg(self.neutral_color), before_longest_match));
-            styled_text.push((Style::new().fg(self.match_color), longest_match));
+            styled_text.push((
+                Style::new().fg(self.match_color),
+                line[lowercase_buffer_split[0].len()
+                    ..(lowercase_buffer_split[0].len() + longest_match.len())]
+                    .to_string(),
+            ));
             styled_text.push((
                 Style::new().bold().fg(self.neutral_color),
                 after_longest_match,
--
2.28.0
```

jonathandturner: Seems like the PR should be a draft if you haven't tested it, yet. Please do definitely test before it lands.

Once you do, you can take a screenshot or copy/paste the output here. It'll help with reviewing.

mzanrosso: > Seems like the PR should be a draft if you haven't tested it, yet. Please do definitely test before it lands.
> 
> Once you do, you can take a screenshot or copy/paste the output here. It'll help with reviewing.

Only the example wasn't tested but I will drop here a before/after screenshot so you can easily understand this improvement.

mzanrosso: Before the PR changes, a command with some upper case characters, is not recognised:
Schermata 2021-07-08 alle 09 43 07

With this PR changes, the highlights is recognised even if you put some upper case characters:
Registrazione schermo 2021-07-08 alle 09 45 49

As you can see the command has been recognized and re-wrote all in a lower case, in one of my previous comment I have illustrated approximately how to rewrite the command without force it all in lower case.

jonathandturner: I wonder if this functionality should be part of the completer. If the user types `cLeA` and then another character, we probably want to just style that the best way that matches what the highlighter says.

Another way to say it is that we can separate the responsibility of the different parts of the system. The highlighter styles the code you type, the completer can update the code. We could have a second type of pass that does a kind of "autocorrect" pass also.

mzanrosso: Now that we have the completer yes I agree

mzanrosso: @jntrnr According to the latest changes with the Completer, we still want to continue for that road or do we keep this implementation?

jntrnr: A related thing I think we could add is that when we complete, the completer could be fuzzy and match against the text even if the lowercase/uppercase didn't quite match.

So basically we wouldn't correct during the completion.

Curious to hear your thoughts

mzanrosso:
With this PR in the end enable the recognition of phrases even with some uppercase characters by the highlighter without modifying the line_buffer in any way.

Registrazione schermo 2021-07-13 alle 13 35 54

src/highlighter.rs
mzanrosso: @jntrnr Practically it is the same case that we are also facing here with the highlighter; surely this improvement will also be added to the completer.

sholderbach:
I did not fully follow your conversation, but my expectation would be as most languages are case sensitive that we only highlight case sensitive matches/or require that implementers opt into specific case insensitivity. For completion especially of file paths having the option to be case insensitive would be great but the semantic change in this case would be deliberate after hitting tab.

fdncred:
please keep a case insensitive option for internals, externals, and completions. we've run into those problems a few times in nushell and it's just a pain to deal with. the specific case that comes to mind is completions with paths and with upper and lowercase windows .exe files.

sholderbach:
@fdncred are you talking about completions or do you have other cases in mind for case insensitivity?

If I understand this PR correctly it is only concerned with the basic syntax highlighter. For this application i would assume the reasonable default is case sensitive and anything else should be explicitly opt in from the side of the consuming PL implemention.

fdncred:
I was having PTSD from completions not working on `C:\Some\Path` but worked with `c:\some\path`.  Let's say we make syntax highlighting work for primitive like a filepath, a string, a number, and a date. i would want mixed case to be highlighted differently for filepath, string, date (if it has text like monday).

I agree that programming languages in general are mostly case sensitive, so perhaps language keywords could/should be case-sensitive but other things may not be.

jntrnr:
Sorry for the delay - looking through this, it still seems like it's adding case-insensitivity to the highlighter. I think a better place for that is to show how to do that in the completion part of the engine. The DefaultHighlighter should probably just be really simple and just show how to change the style for the text only.

mzanrosso:
@jntrnr there will be another PR for the completer part but this one aim to just be able to recognise the case insensitive words, some use-case requires that.
for example, someone insert this phrase:
ThIs Is A tEsT
I don't want to translate into "this is a test" I just want to be able to color it; please tell me what you don't understand of this PR.

jntrnr:
@mzanrosso - If it helps, think about how reedline will likely be used. In a lot of cases, it'll be up to whoever is using reedline to say how an input should be styled. We don't need that logic in the DefaultHighlighter. Instead, they'll implement their own highlighters and sometimes they will use case-sensitivity and sometimes they'll use case-insensitivity.

Our job with the DefaultHightlighter is to show the simple case, so they can learn from it and then apply that to their own projects. That's why I'd like to keep DefaultHighlighter simple and not put more logic into it than it needs.

mzanrosso:
@jntrnr ok, never mind.

Issue #55, opened 2021-06-13
Unify functionality for history search


sholderbach:
Currently there are two ways to search the `History`. 

1. There is the `history_search::BasicSearch` implementation used to provide the `Ctrl+R` backwards search. 
  - It is integrated with a modal search prompt but lacks functionality to walk the search back and forward
  - seems to have accumulated some issues due to unrelated development (Using the arrow keys breaks the prompt)
  - Independent struct more amenable in an API to use more advanced search providers (e.g. fuzzy matching)

2. `History` itself now provides a simple stateful prefix search.
  - Keeps track of cursor and searches based on that -> up and down very easy. 
  - More readable and self contained search logic
  - Managing the search prefix is up to the user.

sherubthakur:
@sholderbach I think we can close this. Right?

sholderbach:
Yes!

PR #92, opened 2021-07-06
Unify history operations under a struct. Add traits to make different history approaches feasible

### LINKED ISSUE #86: [Design][Pluggable Components] History and searching
Continuing with the theme of more pluggable components as discussed in https://github.com/jonathandturner/reedline/issues/61. I wanted to get thoughts of people of the following `History` design. This is very close to what we already have and also takes into consideration https://github.com/jonathandturner/reedline/issues/55.

# Approach

Thinking about the ways we can search history. I could come up with the following ways -
- Going forward/back in history normally
- Going forward/back in history based on a prefix
- Searching for a substring (optionally going forward and backwards)
- Fuzzy search a given string (optionally going forward and backwards)

We can encapsulate this sort of a query into a struct like as follows -
```rust
pub enum HistoryNavigationQuery {
    Normal,
    PrefixSearch(String),
    SubstringSearch(String),
    // Suffix Search
    // Fuzzy Search
}
```

Now keeping this in mind we can have a couple of interfaces we can design for our History. 

```rust
pub trait History {
    // append any given string (a command) into the history - store
    fn append(&mut self, entry: String) -> ();
    
    // This will set a new navigation setup and based on input query
    fn set_navigation(&mut self, navigation: HistoryNavigationQuery);
    
    // This moves the cursor backwards respecting the navigation query that is set
    // - Results in a no-op if the cursor is at the initial point
    fn back(&mut self);

     // This moves the cursor forwards respecting the navigation-query that is set
    // - Results in a no-op if the cursor is at the latest point
    fn forward(&mut self);
    
    // Returns the string (if present) at the cursor
    fn string_at_cursor(&self) -> Option<String>;
    
    // we will need an `init` method to I guess which Reedline can call to setup the more complicated ones.
}
```

Some features of this particular design
- We can now have multiple implementors of this like `FileBackedHistory`, `InMemoryHistory`, `DBBackedHistory`, etc.
- We can add any level of intelligence in the implementors as we like as long as they can adhere to this interface.
- All the extensions would be to the type, on the other hand we would have to create more and more functions: like `go_back`, `go_forward`, `go_back_with_prefix`, `go_forward_with_prefix`, `go_back_with_substring`, `go_back_with_fuzzy`, etc
- [Con] We can only have one kind of history navigation at a time in one Reedline process, as we maintain an internal state based on this info. As far as I can see this should be a reasonable assumption to make as a person will only do one kind of search. Well we can circumvent this by using a HashMap to store queries I guess, but I don't think that complication is necessary given that we can assume that there isn't more than one history navigation going on at a given time in one process.

# Implementation

Based on whatever the discussion we have here I can start working on this. Though this will conflict heavily with https://github.com/jonathandturner/reedline/pull/60. I am not super sure of what the state is there.

sherubthakur:
Resolves #86 and #55


src/engine.rs
```diff
@@ -218,31 +220,38 @@ impl Reedline {
...
+                        let new_string = format!("{}{}", substring, c);
+                        self.history
+                            .set_navigation(HistoryNavigationQuery::SubstringSearch(new_string));
+                    } else {
+                        self.history
+                            .set_navigation(HistoryNavigationQuery::SubstringSearch(format!(
+                                "{}",
+                                c
+                            )))
+                    }
```
sherubthakur, line 234: There should be a better way to append string here.

src/engine.rs
```diff
@@ -43,8 +50,9 @@ pub struct Reedline {
     cut_buffer: Box<dyn Clipboard>,
 
     // History
-    history: History,
-    history_search: Option<BasicSearch>, // This could be have more features in the future (fzf, configurable?)
+    history: Box<dyn History>,
+    // TODO change this shit
```
sherubthakur, line 54: Remove this

src/engine.rs
```diff
@@ -218,31 +220,38 @@ impl Reedline {
...
+                            .grapheme_indices(true)
+                            .last()
+                            .map(|(i, _)| i)
+                            .unwrap_or(0);
+
+                        substring.replace_range(grapheme_left_index..substring.len(), "");
+
+                        self.history
+                            .set_navigation(HistoryNavigationQuery::SubstringSearch(substring));
+                    }
```
sherubthakur, line 250: Operation like delete single grapheme?

src/engine.rs
```diff
@@ -300,33 +309,51 @@ impl Reedline {
...
+            (false, HistoryNavigationQuery::PrefixSearch(_)) => {}
+            (false, _) => {
+                let buffer = self.insertion_line().to_string();
+                self.history
+                    .set_navigation(HistoryNavigationQuery::PrefixSearch(buffer));
+            }
         }
+
+        self.history.forward();
     }
```
sherubthakur, line 351: Any way to avoid this duplication?

src/history/base.rs
```diff
@@ -0,0 +1,37 @@
...
+    SubstringSearch(String),
+    // Suffix Search
+    // Fuzzy Search
+}
+
+pub trait HistoryAppender {
+    // append any given string (a command) into the history - store
+    fn append(&mut self, entry: String) -> ();
+
+    fn iter_chronologic(&self) -> Iter<'_, String>;
```
sherubthakur, line 16: Make this type signature more generic.

src/history/file_backed.rs
```diff
@@ -0,0 +1,543 @@
...
+                let mut cursor = self.cursor;
+
+                let previous_match = self.string_at_cursor();
+
+                while cursor > 0 {
+                    cursor -= 1;
+                    let entry = &self.entries[cursor];
+                    if entry.starts_with(&prefix) {
+                        if previous_match
+                            // TODO Get rid of this clone
```
sherubthakur, line 91: Remove this clone.

src/history/file_backed.rs
```diff
@@ -0,0 +1,543 @@
...
+                            .clone()
+                            .map_or(false, |value| &value == entry)
+                        {
+                            continue;
+                        } else {
+                            break;
+                        }
+                    }
+                }
```
sherubthakur, line 122: Lot's of duplication. Any way to avoid this?

src/engine.rs
jonathandturner: Can you say a bit more here?

jonathandturner: Also let's keep it "family friendly" as they say

src/engine.rs
jonathandturner: If the History trait had some way of adding to the history, that seems like it would simplify the API call there

src/engine.rs
jonathandturner: I think we have this logic in other places already. Can we re-use the logic the buffer has for the string?

src/engine.rs
jonathandturner: You can move line 342 up:

```rust
            (true, HistoryNavigationQuery::Normal) => {}
            (false, HistoryNavigationQuery::PrefixSearch(_)) => {}
            _ => {
                let buffer = self.insertion_line().to_string();
                self.history
                    .set_navigation(HistoryNavigationQuery::PrefixSearch(buffer));
            }
```

src/history/file_backed.rs
jonathandturner: If they're identical you can carry two difference cases in a match like `A | B => { ... }` Is that what you're asking?  You'd need to bind to the same name, but assuming you do and they're the same type, the rest would become a single match body

jonathandturner (commented):
Added some comments. The points you have about duplication and removing clones are fine, but to my mind these could be part of an optimization pass you do later.

Are you feeling comfortable with the general direction in the PR? I noticed in the design, you also mention things like fuzzy searching, which it looked like we could build onto this PR.

src/engine.rs
jonathandturner: One idea would be for the history call to take a callback. In the callback, it could pass you what the current string is, and then from there you could append or change the string.

That said, I'm all for quick-and-dirty first, and then going back over it and cleaning it up once it works.

src/history/file_backed.rs
jonathandturner: Oh I see where they differ now. Unfortunately, the only way I know of is to introduce macros here or functions that take functions as parameters. That could work, but it's difficult to do that but also keep the code readable.

src/engine.rs
sherubthakur: I will revist this point again in the future if that is acceptable.

src/engine.rs
sherubthakur: This was an TODO for me, earlier I had a boolean here to control some flow, which I was not a fan of. I removed the boolean but forgot to remove the comment. The Boxed history is fine with me too. 

Will keep the language in check :)

src/engine.rs
sherubthakur: I can think of a way to make this into a function and then reuse it here and in line buffer. Thoughts?

src/history/file_backed.rs
sherubthakur: I would also like to avoid the usage of macros. Maybe I can pass in matching criteria as a function and then have a generic function perform all the other operations. Though this is something that can be picked up in a later PR.

src/engine.rs
jonathandturner: @sherubthakur sounds good

sherubthakur:
> Added some comments. The points you have about duplication and removing clones are fine, but to my mind these could be part of an optimization pass you do later.

Are you feeling comfortable with the general direction in the PR? I noticed in the design, you also mention things like fuzzy searching, which it looked like we could build onto this PR.

I also agree with the point of revisiting to clean up duplication and do optimizations if the general direction of the PR is good. (which I hope is the case here)

I am pretty comfortable with how the unification worked out overall. I did not and fuzzy search as part of the PR as I wanted to not mix adding feature and clean up of code (i.e. merging both histories). I think is this something that we can take up at a later stage.

src/history/file_backed.rs
sherubthakur: https://github.com/jonathandturner/reedline/pull/92/commits/0cc04453f76bc82d74b2ee26bcb20e454559433e

src/engine.rs
sherubthakur: In one case I am setting navigation to Normal and in one to `PrefixSearch` so unfortunately I can't really do this.

sherubthakur: https://github.com/jonathandturner/reedline/pull/92/commits/2a77ebb8c59089046c1276a73bf0a76e9349bcc1 Extracts the command part as an abstraction into a function.

src/history/base.rs
sherubthakur: I Will revisit this one again in a separate PR

src/history/file_backed.rs
sherubthakur: I will revisit this again in a separate PR

src/engine.rs
sherubthakur: https://github.com/jonathandturner/reedline/pull/92/commits/d36601b0279c2add059a0440f90156ee546bac8d I only extracted the part in engine into a separate function and did not touch line_buffer in this PR. LineBuffer has a lot of custom logic and for at-least now I would like to not worry about unifying it with line_buffer stuff.

sherubthakur:
@jonathandturner I have marked conversations resolved by adding a comment with the rationale behind it. If there is anything that you would me to take care of please mark the conversation unresolved. 

P.S. I am not super sure about the dev flow that you have in this repo, like is it ok that I (PR author) marks the conversation resolved or should the reviewer be the only one that does so.

jonathandturner:
@sherubthakur - yup, it's totally okay to mark things as resolved if you think they're resolved. Giving this PR another look now

jonathandturner:
This looks good to me.

PR #93, opened 2021-07-06
[Add] first implementation of auto completion

mzanrosso:
Initial commits for adding the autocomplete feature; I would need to understand how we want to recognize the tab event and especially how to recognize multiple tabs event in series to go on with the suggestions; let me know if you have some ideas

### DIFF
--- a/src/completer.rs
+++ b/src/completer.rs
@@ -1,0 +1,433 @@
+use {
+    crate::line_buffer::LineBuffer,
+    std::{
+        collections::{BTreeMap, BTreeSet},
+        rc::Rc,
+        str::Chars,
+    },
+};
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Ord, PartialOrd, Hash)]
+pub struct Span {
+    pub start: usize,
+    pub end: usize,
+}
+
+impl Span {
+    // Creates a new `Span` from start and end inputs.
+    // The end parameter must be greater than or equal to the start parameter.
+    pub fn new(start: usize, end: usize) -> Span {
+        assert!(
+            end >= start,
+            "Can't create a Span whose end < start, start={}, end={}",
+            start,
+            end
+        );
+
+        Span { start, end }
+    }
+}
+
+pub trait TabHandler {
+    fn handle(&mut self, line: &mut LineBuffer);
+    fn reset_index(&mut self);
+}
+
+pub struct DefaultTabHandler {
+    completer: Box<dyn Completer>,
+    initial_line: LineBuffer,
+    index: usize,
+}
+
+impl DefaultTabHandler {
+    pub fn with_completer(mut self, completer: Box<dyn Completer>) -> DefaultTabHandler {
+        self.completer = completer;
+        self
+    }
+}
+impl Default for DefaultTabHandler {
+    fn default() -> Self {
+        DefaultTabHandler {
+            completer: Box::new(DefaultCompleter::default()),
+            initial_line: LineBuffer::new(),
+            index: 0,
+        }
+    }
+}
+impl TabHandler for DefaultTabHandler {
+    // With this function we handle the tab events.
+    //
+    // If completions vector is not empty we proceed to replace
+    //  in the line_buffer only the specified range of characters.
+    // If internal index is 0 it means that is the first tab event pressed.
+    // If internal index is greater than completions vector, we bring it back to 0.
+    fn handle(&mut self, line: &mut LineBuffer) {
+        if self.index == 0 {
+            self.initial_line = LineBuffer::new();
+            self.initial_line.set_buffer(line.get_buffer().into());
+            self.initial_line
+                .set_insertion_point(line.insertion_point());
+        } else {
+            line.set_buffer(self.initial_line.get_buffer().into());
+            line.set_insertion_point(self.initial_line.insertion_point())
+        }
+        let completions = self.completer.complete(
+            self.initial_line.get_buffer(),
+            self.initial_line.insertion_point().offset,
+        );
+        if !completions.is_empty() {
+            match self.index {
+                index if index < completions.len() => {
+                    self.index += 1;
+                    let span = completions[index].0;
+                    let mut insertion_point = line.insertion_point();
+                    insertion_point.offset += completions[index].1.len() - (span.end - span.start);
+
+                    // TODO improve the support for multiline replace
+                    line.replace(span.start..span.end, 0, &completions[index].1);
+                    line.set_insertion_point(insertion_point);
+                }
+                _ => {
+                    self.reset_index();
+                }
+            }
+        }
+    }
+
+    // This function is required to reset the index
+    // when following the completion we perform another action
+    // that is not going to continue with the list of completions.
+    fn reset_index(&mut self) {
+        self.index = 0;
+    }
+}
+
+pub trait Completer {
+    fn complete(&self, line: &str, pos: usize) -> Vec<(Span, String)>;
+}
+
+#[derive(Debug, Clone)]
+pub struct DefaultCompleter {
+    root: CompletionNode,
+    inclusions: Rc<BTreeSet<char>>,
+    min_word_len: usize,
+}
+
+impl Default for DefaultCompleter {
+    fn default() -> Self {
+        let inclusions = Rc::new(BTreeSet::new());
+        Self {
+            root: CompletionNode::new(inclusions.clone()),
+            inclusions,
+            min_word_len: 2,
+        }
+    }
+}
+impl Completer for DefaultCompleter {
+    /// Returns a vector of completions and the position in which they must be replaced;
+    /// based on the provided input.
+    ///
+    /// # Arguments
+    ///
+    /// * `line`    The line to complete
+    /// * `pos`   The cursor position
+    ///
+    /// # Example
+    /// ```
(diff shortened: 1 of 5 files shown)


mzanrosso:
The crossterm event is: 
`KeyEvent { code: Tab, modifiers: NONE }`

mzanrosso:
For the multiple tab pressed problem we can introduce some sort of memory of the previous command:

For example, if someone press tab, we store that information + we store the index of outputted completion (since i made the complete function returning an array of possible candidates).

jonathandturner:
This looks good.

For the tab completions, I was thinking that we'd have to send a little more context to the completer. Like tell them how many repetitions of the completion key we saw.

mzanrosso:
Thanks for the comment @jonathandturner ,
i introduced the "index" argument to the complete function so we can tell how many repetitions of the completion key we saw.

> We could introduce a second trait that handles doing the actual completion. So that could do the popup, or the rotation or showing a menu.

Do you have some precise structure in mind ?

fdncred:
Unrelated to the code in this PR, [I saw this](https://github.com/railsware/upterm/blob/master/src/plugins/completion/Git.ts) and thought it was an interesting way to do external completions. Granted, this is hard-coded completions, which I don't think is good, but it is still interesting. There are a lot more in this repo too.

mzanrosso:
> Unrelated to the code in this PR, [I saw this](https://github.com/railsware/upterm/blob/master/src/plugins/completion/Git.ts) and thought it was an interesting way to do external completions. Granted, this is hard-coded completions, which I don't think is good, but it is still interesting. There are a lot more in this repo too.

I'm testing this crate: 
https://github.com/blightmud/rs-complete

mzanrosso:
Massive update, now we have a prototype fully working !

fdncred:
so excited to try this out!!!

mzanrosso:
@fdncred I updated the main.rs file, so you can easily test from there 👍

src/completer.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// completions.insert(vec!["batman","robin","batmobile","batcave","robber"].iter().map(|s| s.to_string()).collect());
+    /// assert_eq!(
+    ///     completions.complete("bat",0),
+    ///     Some("batcave".to_string()));
+    ///
+    /// assert_eq!(
+    ///     completions.complete("to the bat",0),
+    ///     None);
+    /// ```
+    fn complete(&self, line: &str, index: &mut usize) -> Option<String> {
```
jonathandturner, line 95: I was reading the description to this function, and I think the description is a good place to start. If you take in the line and then return a vector of results, then another part of the code could display those options to the user and they could pick one.

Here, if I'm reading it right, we have to already know what index we want?

Apologies if you explain this elsewhere.

Oh, now that I reread it, the TabHandler is the thing showing the completions to the user and here we're maybe rotating through the options?

I still think it'd make sense to have completer return all possible options, and then have the completions handler deal with the vector. That way it works with a lot of different types of completions (like popping up a table the user uses arrow keys on, for example)

mzanrosso: Hey @jonathandturner thanks for the comment, for me it is clear that this complete function only takes care of completing the line, how we display the options is managed in the tabhandler handle function

jonathandturner: @mzanrosso - something I wasn't sure of is how the tabhandler knows what the completions are. How does it know what the list is to show the user?

mzanrosso: I understand the point, now I managed only the inline tab so I don't need a vector but only the candidate, we should find a proper struct to do both in future

mzanrosso:
Now I just  resolved the branch conflicts

mzanrosso:
Don't mind latest 2 force-pushed I was testing an improvement but I will rethink at it later.

src/completer.rs
jonathandturner: I think returning the vector is the right call here. I was also thinking about making reedline available as a line editor for Nushell, and it'd be great to be able to use a similar competitions style, which also returns a vector of results and then the line editor handles the rest.

mzanrosso: Ok let me implement this change

mzanrosso:
@jonathandturner changes made!

mzanrosso:
In my opinion is pretty stable, anyone suggest other changes ?

sholderbach:
Awesome! Do we want to send the span that could be changed by the completer from the engine (requiring the engine to be token aware through another component) or be able request which span would be suggested to change by the Completer? I would imagine we might want to be able to highlight that or have behavior when trying to complete inside a line (like changing a path in a longer command line)

mzanrosso:
@sholderbach Good reflection, at the moment I don't have an answer but I imagine the completer has only the task of completing a missing part of a well-defined phrase so from my point of view we should go to act on the tabtabhandler in some way.

mzanrosso:
What do you guys prefer more as name, completer or completor ?

sholderbach:
Maybe the simple compromise would be to pass the cursor position (`InsertionPoint` even though I would bikeshed that name to `BufferCursor`) to both the `TabHandler` and `Completer` and let them deal with that for the moment if someone wants to make it smarter?

mzanrosso:
@sholderbach 
The point is that if you have a phrase of this type, very common if you have subcommands:

> test1 tes2t h test3 test4

And the cursor is immediately after the 'h' , you have 3 choices to complete:

> test1 test2 help3 ...
> ... test2 help2 ...
> ... help1 ...

So surely we need an indicator of how many words take before the actual one.

sholderbach:
Yes this is certainly the advanced stuff if we want to take into account not only which type of token we are completing but also what is semantically allowed at the current position with tolerance for possible missing bits. But if your PR brings basic completion like dumb (ba)sh we are certainly making enough progress! We can implement the git subcommand completion later ;)

mzanrosso:
@sholderbach yeah, I think that is better if we focus on a feature that we can move forward faster rather than staying here to think of all possible improvements.

mzanrosso:
@jntrnr do you have other's suggestions?

jntrnr:
@mzanrosso - something we can do in this PR or in a following one would be to see if we could simplify what the completer needs to implement. It looks like currently anyone implementing a Completer needs to implement four different methods:

```
pub trait Completer {
    fn complete(&self, line: &str) -> Option<Vec<String>>;
    fn get_completions(&self) -> Option<&Vec<String>>;
    fn set_completions(&mut self, completions: Vec<String>);
    fn reset_completions(&mut self);
}
```

I think we only need the first one, and we should be call into the Completer, get the results, and then handle the other three from inside the engine rather than requiring the completer to allow us to set completions and get completions.

It's less bookkeeping for the implementer of Completer, which will be the user of the library.

mzanrosso:
@jntrnr Yes, I am of your opinion, the simpler it is to implement the better;  but in this case how would you not force the implementation of those functions?

jntrnr:
@mzanrosso - the way I understand it is that when the user hits the completion key, we pass in the line and then the `complete` method gives back all the completions for that position (could be a `Vec<String>` rather than an `Option<Vec<String>>` to show that it always returns values).

Once this runs, the engine now has all the completions for that given state. If you user hits tab again, and the edit state hasn't changed, I think that's where you can move to the next completion. If the edit state has changed, then we re-run the method above and get the new list.

I think the logic for how to handle the list should live outside of the completer and base it on whether the user is moving to the next completion or if they've edited the contents or moved the cursor since the last completion. Having said that, I wonder if we should pass in the cursor position to the completer, so we know where to check for completions in the string.

mzanrosso:
If I have to be sincere, I prefer more implementation with the `Option<Vec<String>>` because it is more intuitive since if the function fails to complete the line does not return the various options.
The logic is a little different from the one you guessed: 
for how I wrote the current implementation at each event that is not the tab, i update the tab-handler object with the`update() ` function; but when the user invokes the Tab event I go to check the index and only later return the right completion. This implementation takes into account any updates to the line while pressing repeated tabs, takes into account the starting line even with multiple tabs and updates the starting line after each event that is not tab, so if a user is satisfied with the completion and goes on to write, the tab does not reset the line.

mzanrosso:
@jntrnr For the too many functions in trait problem I had thought of keeping the object by default as a reference and go to act with appropriate functions to make it dynamic as in this case:
`Reedline::new().with_completer(Box::new(DefaultCompleter::new_with_wordlen(completion_list, 2)))`
So my strategy is to avoid having to re-implement all the object from 0, making the default object full customizable.

jntrnr:
@mzanrosso - I think you have the right idea here for building a better performance completion engine. By letting the completer keep a snapshot of the data, you can definitely get better results out of it. I've seen similar approaches for IDEs, for example.

For the line editor, I think the advantage of simplicity outweighs the better performance that your design gives. It does mean throwing away and redoing the parse that the Completer has to do, but since each input we're taking in is at most only a few lines, we may not need that extra performance.

I think that's why I'm leaning towards not keeping any state in the Completer (beyond what the implementer puts in it) and instead just having that simpler trait with the one method.

If you're on the [Nushell discord](https://discord.gg/NtAbbGn), we've got a #reedline channel there for doing design brainstorming. Might be easier to chat there. We could explore a few different approaches and see which one works best.

mzanrosso:
Registrazione_schermo_2021-07-13_alle_00 26 11

Updated with inside line completion

jntrnr:
I think we're good to land this and continue integrating more features into it.

PR #95, opened 2021-07-12
Remove some unnecessary stuff from the engine

sholderbach:
* tool for event tracking is only relevant for testing -> main
* unnecessary enable_raw_mode call
* we don't enable mouse mode yet
* clippy suggested removals of overspecified type and clone


jntrnr:
@sholderbach - I landed the completer first, since it was the larger change. Would you mind rebasing? The fixes look good

PR #96, opened 2021-07-13
Remove the updating clock

sherubthakur:
This clock currently interferes with updates to the display buffer.
We should hold off implementing it till the point where we can do so
without interfering with the history searching display etc.

Note to reviewers: most of this is whitespace change and using `?w=1` query params will help in seeing the actual diff that matters.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -734,117 +734,112 @@ impl Reedline {
         }
 
         loop {
-            if poll(Duration::from_secs(1))? {
-                match read()? {
-                    Event::Key(KeyEvent { code, modifiers }) => {
-                        match (modifiers, code, self.edit_mode) {
-                            (KeyModifiers::CONTROL, KeyCode::Char('d'), _) => {
-                                if self.line_buffer.is_empty() {
-                                    return Ok(Signal::CtrlD);
-                                } else if let Some(binding) = self.find_keybinding(modifiers, code)
+            match read()? {
+                Event::Key(KeyEvent { code, modifiers }) => {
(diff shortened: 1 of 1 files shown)


mzanrosso:
In my opinion we should find another way to fix the history command disappear problem without removing the 1s pool

sholderbach:
Yes the 1 sec interval can be annoying when debugging (in that case set it temporarily to a longer duration). But the goal should generally be that after every repainting (mode switch, resize etc.) we result in a valid output based on the state of the buffer, prompt and other modal widgets. I personally wouldn't require a clock for my terminal, but I think it is a good exercise to get everything right for reedline and also a good visual indicator if we hit any performance regressions.

sherubthakur:
> In my opinion we should find another way to fix the history command disappear problem without removing the 1s pool

I really think that we should add this feature later on. When we have some way to add widgets (or something) to reedline. An if-else does not feel like a proper solution for this particular case.
And practically while debugging it gets in the way a lot which was the main motivation for this particular PR.

jntrnr:
Seconded @sholderbach - @sherubthakur I agree that it feels like it needs a bit more work, but I think it's actually good work for us to do. We we're not able to do the 1sec refresh, we're probably in need of looking at parts of the architecture that can't handle it. Once that's worked out, we should be able to handle different types of prompts the user can throw at us.

That said, what are some better ways we could address the issue. @sherubthakur you mention if/else doesn't feel like a proper solution, what are some better solutions we should work with?

fdncred:
one of my favorite features of reedline so far is watching the seconds tick by in the prompt.

PR #97, opened 2021-07-13
History fix cursor update

sherubthakur:
Various fixes to history execution

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -330,7 +330,13 @@ impl Reedline {
             (true, _) => {
                 self.history.set_navigation(HistoryNavigationQuery::Normal);
             }
-            (false, HistoryNavigationQuery::PrefixSearch(_)) => {}
+            (false, HistoryNavigationQuery::PrefixSearch(prefix)) => {
+                let buffer = self.insertion_line().to_string();
+                if prefix != buffer {
+                    self.history
+                        .set_navigation(HistoryNavigationQuery::PrefixSearch(buffer));
+                }
+            }
             (false, _) => {
                 let buffer = self.insertion_line().to_string();
                 self.history
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -436,6 +436,23 @@ impl Reedline {
             return;
         }
 
+        if self.input_mode == InputMode::HistoryTraversal {
+            for command in commands {
+                match command {
+                    EditCommand::PreviousHistory => {}
+                    EditCommand::NextHistory => {}
+                    _ => {
+                        if self.history.get_navigation() == HistoryNavigationQuery::Normal {
+                            if let Some(string) = self.history.string_at_cursor() {
+                                self.set_buffer(string)
+                            }
+                        }
+                        self.input_mode = InputMode::Regular;
+                    }
+                }
+            }
+        }
+
         // Vim mode transformations
         let commands = match self.edit_mode {
             EditMode::ViNormal => self.vi_engine.handle(commands),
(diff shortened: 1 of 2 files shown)


sholderbach:
Still some usability painpoints:

* Ctrl-R search has the refresh bugs
* moving the cursor left/right without editing breaks the ability to go up/down
* still skips the most recent entry going up
* you can not return to the (empty) linebuffer by going down beyond the history (some shells like bash allow you to return to a partial manual entry if they don't support the prefix searching history browsing)

For the last two points the iterator like interface of moving while either returning Some(entry) when a move was succesful or None when exhausted had some benefits over the new setup. We want to be able to handle those cases. I started some changes on that, if I get them to work I will file a PR.

sholderbach:
OK Update regarding point two on my list. zsh's history traversal would also stop at this point as it then thinks it is in prefix mode. Yet bash allows horizontal cursor movement without interfering with the basic up/down history navigation. So this seems to be somewhat of an UI/UX choice.

sherubthakur:
> Ctrl-R search has the refresh bugs

IIRC this has existed for a long time, I wasn't really focused on this one. I really think we should get rid of the clock update for now and introduce it later down the road.

> moving the cursor left/right without editing breaks the ability to go up/down

Can you elaborate on this one (ideally with an example). So, I can understand what the behaviour should be. I believe the present state (in the PR) should be that left/right movement etc should stop history searching altogether for Normal and PrefixSearch modes. That is what I wanted to implement.
Though I think when doing `crtl-r` search this should be allowed (i.e. going left/right). (Not sure what the present behaviour is) 

> still skips the most recent entry going up

Yeah, I plan to fix this one.

> ou can not return to the (empty) line buffer by going down beyond the history (some shells like bash allow you to return to a partial manual entry if they don't support the prefix searching history browsing)

I had not really thought this through. Definitely, something that should happen.

jntrnr:
@sherubthakur - taking a look at this now. I'm seeing something related to a previous PR you did to clean up history and I think it'd be great to fix it as part of this update.

If you press up/down and wait a second, the input disappears. Before #92 this worked fine, but after that it seems that after the second ticks by we only draw the prompt and not the user's input.

src/line_buffer.rs
```diff
@@ -65,8 +65,17 @@ impl LineBuffer {
 
     /// Set to a single line of `buffer` and reset the `InsertionPoint` cursor
     pub fn set_buffer(&mut self, buffer: String) {
```
sholderbach, line 67: Are other users of `set_buffer` maybe expecting a different behavior? Else we may want to have have `set_buffer_cursor_end` or something else.

src/engine.rs
```diff
@@ -337,7 +337,13 @@ impl Reedline {
             (true, _) => {
                 self.history.set_navigation(HistoryNavigationQuery::Normal);
             }
-            (false, HistoryNavigationQuery::PrefixSearch(_)) => {}
+            (false, HistoryNavigationQuery::PrefixSearch(prefix)) => {
```
sholderbach, line 340: Out of curiosity: Against which state is this guarding?

sholderbach (commented):
Let's get this merged that we can remove this history bug, and check for the new interactions.

If the refactored line buffer will have to be taken in to account depends on what gets merged first.

jntrnr:
Yeah I'm also :+1: on landing and continuing to polish. @sherubthakur thanks for the fixes!

src/engine.rs
sherubthakur: Each up/down press will call this function. if the prefix does not change we don't want to reset the history state.

PR #98, opened 2021-07-14
[Draft][Improvement] Initial support for hints

mzanrosso:
Initial support for hints

Surely there is some problem with the cursor position to be solved, any help is appreciated.

### DIFF
--- a/src/completer.rs
+++ b/src/completer.rs
@@ -31,6 +31,7 @@ impl Span {
 pub trait TabHandler {
     fn handle(&mut self, line: &mut LineBuffer);
     fn reset_index(&mut self);
+    fn get_completer(&self) -> &Box<dyn Completer>;
 }
 
 pub struct DefaultTabHandler {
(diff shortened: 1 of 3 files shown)


mzanrosso:
Currently this draft was to reach an mvp and to prove that we can add the under-buffer hints; but with @jntrnr, we agreed to start first with the in-line hints and we also discuss how to implement in a better structured way.
So don't waste too much time to understand this feature right now, thanks.

fdncred:
How about a screenshot of what it looks like?

mzanrosso:
> How about a screenshot of what it looks like?

test

fdncred:
@mzanrosso That's so cool. Thanks for sharing! Hopefully it'll eventually get accepted.

mzanrosso:
@fdncred look a the the new PR, i writed the right hinter structure; I will refactor this one based on that and hopefully I will get rid of some bugs

sholderbach:
Would a more accurate name for the PR be (auto)suggestion list?

mzanrosso:
> Would a more accurate name for the PR be (auto)suggestion list?

Yes we can call the functionality in many ways, basically it's an under-buffer hint list

fdncred:
@mzanrosso I saw this the other day and thought about this PR https://github.com/marlonrichert/zsh-autocomplete

PR #101, opened 2021-07-15
[Add] in-line hints

mzanrosso:
Implemented features:
 - optional inside line hints
 - customizable hints style
 - hints from completer
 - hints from history

Example:

Registrazione schermo 2021-07-15 alle 10 28 16

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -5,6 +5,7 @@ use {
         clip_buffer::{get_default_clipboard, Clipboard},
         completer::{DefaultTabHandler, TabHandler},
         default_emacs_keybindings,
+        hinter::{DefaultHinter, Hinter},
         history::{FileBackedHistory, History, HistoryNavigationQuery},
         keybindings::{default_vi_insert_keybindings, default_vi_normal_keybindings, Keybindings},
         line_buffer::{InsertionPoint, LineBuffer},
@@ -88,7 +89,8 @@ impl Reedline {
         let history = Box::new(FileBackedHistory::default());
         let cut_buffer = Box::new(get_default_clipboard());
         let buffer_highlighter = Box::new(DefaultHighlighter::default());
-        let painter = Painter::new(stdout(), buffer_highlighter);
+        let hinter = Box::new(DefaultHinter::default());
+        let painter = Painter::new(stdout(), buffer_highlighter, hinter);
         let mut keybindings_hashmap = HashMap::new();
         keybindings_hashmap.insert(EditMode::Emacs, default_emacs_keybindings());
         keybindings_hashmap.insert(EditMode::ViInsert, default_vi_insert_keybindings());
@@ -108,6 +110,12 @@ impl Reedline {
             tab_handler: Box::new(DefaultTabHandler::default()),
         }
     }
+
+    pub fn with_hinter(mut self, hinter: Box<dyn Hinter>) -> Reedline {
+        self.painter.set_hinter(hinter);
+        self
+    }
+
     pub fn with_tab_handler(mut self, tab_handler: Box<dyn TabHandler>) -> Reedline {
         self.tab_handler = tab_handler;
         self
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -606,8 +606,12 @@ impl Reedline {
         let cursor_position_in_buffer = self.insertion_point().offset;
         let buffer_to_paint = self.insertion_line().to_string();
 
-        self.painter
-            .queue_buffer(buffer_to_paint, prompt_offset, cursor_position_in_buffer)?;
+        self.painter.queue_buffer(
+            buffer_to_paint,
+            prompt_offset,
+            cursor_position_in_buffer,
+            &self.history,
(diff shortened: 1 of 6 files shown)


fdncred:
Nice!

mzanrosso:
Before merge, i want the inside-line hints to be optional

fdncred:
Ya, i'd also like the hints to be dark gray/configurable but we can change that sometime in the future.

mzanrosso:
> Ya, i'd also like the hints to be dark gray/configurable but we can change that sometime in the future.

For the configurable style it's already implemented

mzanrosso:
> Before merge, i want the inside-line hints to be optional

Done.

jntrnr:
@mzanrosso - taking a look now.

For the history-based hints, I was testing it by typing something like:

```
> another test
> a
```

For that second line, I should see a hint for `another test`, right?

mzanrosso:
> @mzanrosso - taking a look now.
> 
> For the history-based hints, I was testing it by typing something like:
> 
> ```
> > another test
> > a
> ```
> 
> For that second line, I should see a hint for `another test`, right?

@jntrnr This is not my implementation fault; this is because the history file is updated only when you exit.

jntrnr:
@mzanrosso - if I exit and restart, I still don't get a hint for the history

mzanrosso:
> @mzanrosso - if I exit and restart, I still don't get a hint for the history

@jntrnr Sorry if it wasn't clear but you cannot enable completer and history hints at the same time

jntrnr:
@mzanrosso - if I switch to using the history, are there other settings I need to change so that I can complete the hint?  I tried switching, but if I hit tab, it completes the completion but not the history hint.

mzanrosso:
> @mzanrosso - if I switch to using the history, are there other settings I need to change so that I can complete the hint? I tried switching, but if I hit tab, it completes the completion but not the history hint.

@jntrnr I also noticed this when I did my implementation test; this is because I added the history support for the DefaultHinter and not for the DefaultTabHandler; this could be a next feature for the tab handler

sholderbach:
With fish or the appropriate zsh plugin it is typically right arrow or control right arrow to complete the hint at the end of the line either fully or just the next token. This would use a separate keybinding and also restrict greyed out history hints to the end of the line during entry. Depends on what we want to achieve.

jntrnr:
@sholderbach / @mzanrosso - sounds good, we can add just the hint, and then go back and add additional commands for doing the completions.

mzanrosso:
@jntrnr I just noticed that if I use `.with_history(history.clone())?` it doesn't save the history in the file, do you know why this happens ?

sholderbach:
In the current setup flushing to disk occurs on dropping the instance.

mzanrosso:
> In the current setup flushing to disk occurs on dropping the instance.

Why don't insert an option to flush every new command ?

sholderbach:
We could `flush()` on every `append()` but as soon as the file would be at the chosen capacity and you could not use the append function of the file system anymore, you would have to write the whole file on every `append()`. At least with the current impl.

mzanrosso:
@sholderbach I just noticed that the problem that block us occur only when I declare two `Box::new(FileBackedHistory::with_file(5, "history.txt".into())?)`

sholderbach:
Maintaining consistent views within our process via the FS doesn't seem right. Why would you want to use multiple history instances?

jntrnr:
We could set it up so that you have one History instance that's protected with locks.

mzanrosso:
Fixed.

mzanrosso:
@jntrnr , @sholderbach let me know if we are ready to merge or if I have to modify something else; before merge I need 1sec to squash as usual.

sholderbach:
> We could set it up so that you have one History instance that's protected with locks.

If I didn't loose track of potential consumers we basically would only have to hand out immutable views/ or iterators to the components like history browsing/search/the completion engine. Only upon finishing the `read_line` function successfully we would need the mutable access to the data structure (Or if we regularly poll for changes from other processes running reedlines) The history browsing and search are currently mutually exclusive thus them being able to be handled with mutable access to a single struct.

sholderbach:
> @jntrnr , @sholderbach let me know if we are ready to merge or if I have to modify something else; before merge I need 1sec to squash as usual.

We typically merge as squash commits, so you don't have to explicitly squash on your side.

mzanrosso:
> > @jntrnr , @sholderbach let me know if we are ready to merge or if I have to modify something else; before merge I need 1sec to squash as usual.
> 
> We typically merge as squash commits, so you don't have to explicitly squash on your side.

oh! this feature is new for me, I always used bitbucket and I never seen it there. let me search how It works.

sholderbach:
image

It has it's downsides as `git branch --no-merged` won't recognize those merges

mzanrosso:
I reworded some commit messages, nothing important here.

jntrnr:
Agreed, let's do this in stages. For the hints themselves, this looks good.

PR #102, opened 2021-07-16
[StyleFix] remove clippy warnings

mzanrosso:


sholderbach:
That looks reasonable

PR #103, opened 2021-07-16
[StyleFix] make the prompt implementation more elegant

mzanrosso:
I think that `Box<dyn Prompt>` it's more elegant than `&dyn Prompt`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -62,10 +62,8 @@ fn main() -> Result<()> {
                 .with_style(Style::new().italic().fg(Color::LightGray)),
         ));
 
-    let prompt = DefaultPrompt::new(1);
-
     loop {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -62,10 +62,8 @@ fn main() -> Result<()> {
                 .with_style(Style::new().italic().fg(Color::LightGray)),
         ));
 
-    let prompt = DefaultPrompt::new(1);
-
     loop {
-        let sig = line_editor.read_line(&prompt)?;
+        let sig = line_editor.read_line(Box::new(DefaultPrompt::new(1)))?;
```
sholderbach, line 66: This would cause a small heap allocation every time? What is the reasoning? References themselves are can be `dyn` for object safe traits and do not need additional indirection to overcome the `Sized` bound.

PR #105, opened 2021-07-16
Manual checklist to test the UX

sholderbach:
As some bugs are hard to find with automated tests, we should regularly
test the core behavior of reedline, to catch those bugs that might creep
in, when changing the I/O, frontend or core code.

I am looking for suggestions on how we make this manual checklist
comprehensive enough to cover most problems but not too tedious to do
manually. (Ideally we find a way to integration test most of it)


jntrnr:
Love it! I think this will catch the majority of issues people might run into on accident

jntrnr (approved):

PR #107, opened 2021-07-17
[Add] CtrlZ (undo) and CtrlG (redo) features

mzanrosso:
**Simple undo and redo**
Registrazione schermo 2021-07-19 alle 22 16 48

**Cursor position restored after CtrlA-CtrlE with undo and redo**
Registrazione schermo 2021-07-19 alle 22 19 37

**Undo and redo with middle line changes**
Registrazione schermo 2021-07-19 alle 22 21 58

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -566,32 +566,48 @@ impl Reedline {
         // Run the commands over the edit buffer
         for command in &commands {
             match command {
-                EditCommand::MoveToStart => self.move_to_start(),
+                EditCommand::MoveToStart => {
+                    self.move_to_start();
+                    self.line_buffer.set_previous_lines();
+                }
                 EditCommand::MoveToEnd => {
                     self.move_to_end();
+                    self.line_buffer.set_previous_lines();
                 }
-                EditCommand::MoveLeft => self.move_left(),
-                EditCommand::MoveRight => self.move_right(),
+                EditCommand::MoveLeft => {
+                    self.move_left();
+                    self.line_buffer.set_previous_lines();
+                }
+                EditCommand::MoveRight => {
+                    self.move_right();
+                    self.line_buffer.set_previous_lines();
+                }
                 EditCommand::MoveWordLeft => {
                     self.move_word_left();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::MoveWordRight => {
                     self.move_word_right();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::InsertChar(c) => {
                     self.insert_char(*c);
                 }
                 EditCommand::Backspace => {
                     self.backspace();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::Delete => {
                     self.delete();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::BackspaceWord => {
                     self.backspace_word();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::DeleteWord => {
                     self.delete_word();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::Clear => {
                     self.clear();
@@ -616,15 +632,19 @@ impl Reedline {
                 }
                 EditCommand::CutFromStart => {
                     self.cut_from_start();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::CutToEnd => {
                     self.cut_from_end();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::CutWordLeft => {
                     self.cut_word_left();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::CutWordRight => {
                     self.cut_word_right();
+                    self.line_buffer.set_previous_lines();
                 }
                 EditCommand::PasteCutBuffer => {
                     self.insert_cut_buffer();
@@ -649,6 +669,12 @@ impl Reedline {
                 }
                 EditCommand::EnterViNormal => {
                     self.enter_vi_normal_mode();
+                }
+                EditCommand::Undo => {
+                    self.line_buffer.undo();
+                }
+                EditCommand::Redo => {
+                    self.line_buffer.redo();
                 }
                 _ => {}
             }
@@ -896,6 +922,7 @@ impl Reedline {
                             {
                                 self.tab_handler.reset_index();
                                 self.run_edit_commands(&[EditCommand::ViCommandFragment(c)]);
+                                self.line_buffer.set_previous_lines();
                             }
                             (KeyModifiers::NONE, KeyCode::Char(c), x)
                             | (KeyModifiers::SHIFT, KeyCode::Char(c), x)
@@ -926,6 +953,7 @@ impl Reedline {
                                 } else {
                                     self.run_edit_commands(&[EditCommand::InsertChar(c)]);
                                 }
+                                self.line_buffer.set_previous_lines();
                             }
                             (KeyModifiers::NONE, KeyCode::Enter, x) if x != EditMode::ViNormal => {
                                 match self.input_mode {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -228,7 +228,7 @@ impl Reedline {
     fn run_history_commands(&mut self, commands: &[EditCommand]) {
         for command in commands {
             match command {
-                EditCommand::InsertChar(c) => {
+                InsertChar(c) => {
                     let navigation = self.history.get_navigation();
                     if let HistoryNavigationQuery::SubstringSearch(mut substring) = navigation {
                         substring.push(*c);
@@ -242,7 +242,7 @@ impl Reedline {
                     }
                     self.history.back();
                 }
-                EditCommand::Backspace => {
+                Backspace => {
                     let navigation = self.history.get_navigation();
 
                     if let HistoryNavigationQuery::SubstringSearch(substring) = navigation {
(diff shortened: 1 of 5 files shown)


mzanrosso:
Relevant features tested (leave open if you did not consider **those** areas touched by your PR):
- [x]  core editing and default Emacs 
- [ ]  keybindings
- [ ]  history
- [ ]  syntax highlighting
- [ ]  completion/hinting

**Build:**
- [x] debug 
- [x] release

**Platform:**
macOS 12.0 arm64

Terminal **emulator:**
Iterm2

**Basics**

- [x] Typing of a short line containing both upper- and lowercase characters.
- [x] Movement left/right using the arrow keys
- [x] Word to the left with Ctrl-b and Ctrl-Left, Word to the right with Ctrl-f
- [x] Enter to complete entry
- [x] Clearing

- [x] Type something and abort the entry with Ctrl-c, you should end up on an empty prompt below.
- [x] Type something and press Ctrl-l to clear the screen. Your current entry should still be there and passed through when pressing Enter

**Unicode and Emojis**

- [x] Paste the line Emoji test 😊 checks 🤦🏼‍♂️ unicode and move the cursor over the emojis.
- [x] Are you able to delete the smiley?
- [x] Home/End at accurate positions
- [x] Check that the emoji containing line can be entered

**History**

 Enter a unique entry you can remember.
- [ ] On the empty line press the up-arrow key to see if you can recall the previous entry
- [ ] Press Enter to execute this line (it should not be duplicated in the history, after checking leave history recall by down-arrow)
- [x] Leave the application with Ctrl-d on an empty line
- [ ] After rerun you can get the last line back with up-arrow.
- [ ] On an empty line start typing the beginning of a line in the history. Hit the up-arrow to find the matching entry.
- [ ] After that run Ctrl-r to start traditional reverse search. Type your initial search. Can you find more hits by pressing Ctrl-r or up-arrow?
- [x] Abort this search by pressing Ctrl-c

**Syntax highlighting**

- [x] Upon entering test, this word is highlighted differently.

jntrnr:
@mzanrosso - for undo and redo, I think we want to group together related actions. Let's say you type `hello`. If you do an undo, rather than removing a single character, I wonder if it should treat the whole edit as a single step

mzanrosso:
> @mzanrosso - for undo and redo, I think we want to group together related actions. Let's say you type `hello`. If you do an undo, rather than removing a single character, I wonder if it should treat the whole edit as a single step

We can decide when insert the "previous state" now at every character it will insert a new state but we can make only with new words

mzanrosso:
@jntrnr - Now it should be fine, can you confirm ?

jntrnr:
@mzanrosso I tried a few things, and I think the behaviour of undoing a word definitely does feel more natural.

Some things you should try for your tests that seem to be surprising:

* type something like "this is a test", then do CTRL-A, CTRL-K, then try to paste the line with CTRL-Y. This works in main, but doesn't work in this branch
* If you hit enter, you can type CTRL-Z to undo the line you're not on. We probably want to clear the undo history once the input has been submitted back to the user.

I noticed that you repeat the `self.line_buffer.set_previous_lines();` line a lot. I think a better fix for that would be to check what command just ran after you run it and check to see if it's one of a set of commands.

Something like:

`if [EditCommand::MoveToEnd, EditCommand::MoveToStart, ...].contains(command) {
  self.line_buffer.set_previous_lines();
}`

mzanrosso:
@jntrnr - Thanks for your accurate revision, I will improve as you suggested asap.

mzanrosso:
Cargo clippy + cargo test exceeded + introduced all pending improvements

src/keybindings.rs
```diff
@@ -129,7 +131,7 @@ pub fn default_emacs_keybindings() -> Keybindings {
     );
     keybindings.add_binding(
         KeyModifiers::CONTROL,
-        Char('y'),
+        Char('g'),
```
jntrnr, line 134: CTRL-Y is the correct keybinding for paste in emacs. You'll need to pick a different keybinding for redo

jntrnr: You can find a list here of some of the standard emacs keybindings: https://caiorss.github.io/Emacs-Elisp-Programming/Keybindings.html

mzanrosso: @jntrnr - We could reverse and put the redo as ctrl G

jntrnr: Sounds good

mzanrosso: Done

src/engine.rs
```diff
@@ -864,18 +900,21 @@ impl Reedline {
...
                                 return Ok(Signal::CtrlL);
                             }
                             (KeyModifiers::NONE, KeyCode::Char(c), x)
                             | (KeyModifiers::SHIFT, KeyCode::Char(c), x)
                                 if x == EditMode::ViNormal =>
                             {
                                 self.tab_handler.reset_index();
-                                self.run_edit_commands(&[EditCommand::ViCommandFragment(c)]);
+                                self.run_edit_commands(&[ViCommandFragment(c)]);
+                                self.line_buffer.set_previous_lines();
```
sholderbach, line 917: For vi mode the undo behavior will probably have to take additional pecularities into account, that are proabably outside the scope of this PR

src/engine.rs
```diff
@@ -904,20 +943,19 @@ impl Reedline {
                                         prompt_offset.1 -= 1;
                                     }
                                 } else {
-                                    self.run_edit_commands(&[EditCommand::InsertChar(c)]);
+                                    self.run_edit_commands(&[InsertChar(c)]);
                                 }
+                                self.line_buffer.set_previous_lines();
```
sholderbach, line 948: Have you attempted this with history search? I assume there is a potential unwanted interaction, but the traditional reverse search is broken enough that debugging the interaction is lower priority

sholderbach (commented):
Word based behavior mostly works

The only error case I encountered:

```
Pipes as cursor
test broken word|s  

edit:
test br|s
test break|s

undo:
test broken |s
undo:
test broken |
```

src/engine.rs
mzanrosso: > test broken |

@sherubthakur  Thanks for the great test, in fact I have not considered this case where the number of words is the same but the content for some reason has changed

mzanrosso:
@sherubthakur Thanks for the great test, in fact I have not considered this case where the number of words is the same but the content for some reason has changed

src/engine.rs
```diff
@@ -11,7 +11,9 @@ use {
         line_buffer::{InsertionPoint, LineBuffer},
         painter::Painter,
         prompt::{PromptEditMode, PromptHistorySearch, PromptHistorySearchStatus, PromptViMode},
-        DefaultHighlighter, EditCommand, EditMode, Highlighter, Prompt, Signal, ViEngine,
+        DefaultHighlighter,
+        EditCommand::{self, *},
```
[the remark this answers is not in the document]
mzanrosso: I think I don't understand what you want me to understand

src/engine.rs
```diff
@@ -558,92 +560,125 @@ impl Reedline {
...
                 _ => {}
             }
+
+            if [
+                MoveToEnd,
+                MoveToStart,
+                MoveLeft,
+                MoveRight,
+                MoveWordLeft,
+                MoveWordRight,
```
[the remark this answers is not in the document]
mzanrosso: I can say that the intent, given that I also save the old buffer positions is to be able to restore those actions as well.

src/engine.rs
mzanrosso: No I didn't test with history sorry

src/engine.rs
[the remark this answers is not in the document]
sholderbach: Importing the names of all `EditCommand` enum variants at the module top seems somewhat excessive.

src/engine.rs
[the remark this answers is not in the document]
sholderbach: I wasn't noticing undo of moves in the current version.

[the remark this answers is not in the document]
mzanrosso: now that you talk about it I am noticing some small errors, let me verify better

src/engine.rs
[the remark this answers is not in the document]
mzanrosso: > @sherubthakur Thanks for the great test, in fact I have not considered this case where the number of words is the same but the middle content for some reason has changed

In the end I found I missed a part of the logic when treating actions and char inserting

mzanrosso:
> Thanks for the great test, in fact I have not considered this case where the number of words is the same but the middle content for some reason has changed

@sholderbach - In the end I found I missed a part of the logic when treating actions and char inserting

src/line_buffer.rs
```diff
@@ -22,6 +22,9 @@ impl Default for InsertionPoint {
 /// In memory representation of the entered line(s) to facilitate cursor based editing.
 #[derive(Debug, PartialEq, Eq)]
 pub struct LineBuffer {
+    old_lines: Vec<Vec<String>>,
+    old_insertion_point: Vec<InsertionPoint>,
+    index_undo: usize,
```
sherubthakur, line 27: I think we can benefit from an intermediate abstraction of `Editor` or something that has the responsibility of doing core editing operations (this includes writing text, navigation, editing, changing etc but does not include stuff like hinting, history, completion etc). This is essentially the core editor.

Now with this, we can pull the `cut_buffer` in this `Editor`. and this undo/redo functionality too. Here are a few ways it can look like.

```rust
struct Editor {
    edit_buffer: LineBuffer,
    
    // I am thinking of this more like how vim has Registers that you can use to cut/yank stuff.
    // Probably should be a trait (Register) as we might want to implement this for the system clipboard too.
    registers: Map<RegisterID, Box<dyn Register>>,

    // undo/redo basically boils down to going back and forth in the list zipper structure.
    // we can have logic when we are attaching a node to the edits list Zipper.
    // we can probably even remove edit_buffer altogether but I think having it can help in decision making
    // for when to attach a node (like on every edit or we want to do clubbing)
    edits: ListZipper<LineBuffer>,
}
```
Note: [List Zipper (stack overflow link)](https://stackoverflow.com/questions/380438/what-is-the-zipper-data-structure-and-should-i-be-using-it)

Let me know your thoughts on this?

sherubthakur:
> @sherubthakur Thanks for the great test, in fact I have not considered this case where the number of words is the same but the middle content for some reason has changed

I guess you tagged the wrong person with the name starting with `sh` 😝

mzanrosso:
> > @sherubthakur Thanks for the great test, in fact I have not considered this case where the number of words is the same but the middle content for some reason has changed
> 
> I guess you tagged the wrong person with the name starting with `sh` 😝

Oh yes, my fault; the comment was meant for @sholderbach

src/line_buffer.rs
mzanrosso: @sherubthakur - Yes I fully agree to separate the logic parts but if we really want to follow this way right now and in this PR, we should first discuss together the best efficient way to do it

sherubthakur: I guess we should work on this in a later PR. As you said this would need some discussion and an agreement from the collaborators on this project.

src/engine.rs
sholderbach: From my understanding of the history logic now, this line should not cause an issue. But you probably want to think about how you incorporate the change after recalling something from history.

src/engine.rs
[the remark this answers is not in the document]
sholderbach: Most editors don't undo moves so that might be personal taste and we could offer that as an option.

src/engine.rs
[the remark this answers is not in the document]
jntrnr: Agreed, please don't do glob imports like this. It'd be better to print in the enum and work with the cases

src/engine.rs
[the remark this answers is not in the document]
jntrnr: For undo, I think we should focus on changes to the buffer itself. So if I insert, then delete something, undo would handle undoing the delete, then undoing the insert, but if I moved the cursor around between those, the movement wouldn't change.

so, for example:

```
> this is an input
```
then moved to here:
```
> this is an<cursor> input
```
Then inserted:
```
> this is an ordinary input
```

If I did `undo` from there, I'd get back to `"this is an input", with the cursor sitting where it was when I started the insert.

Mostly just going off how undo works in text editors, but I think it'd work well for us, too.

src/engine.rs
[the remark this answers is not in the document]
mzanrosso: Done

mzanrosso:
@jntrnr I removed the glob import as you and @sholderbach  suggested

sholderbach:
Would you mind rebasing onto `main` so we can check the interactions with recalling entries from the history?

sholderbach:
Awesome! Undoing a recalled history entry seems to work and the state of the history returns to default. Redoing did not register an entry previously entered via history.

mzanrosso:
I have some problem with the gpg signature right now, let me solve that before merge, thanks

sholderbach:
When typing something after the history recall (without the enter mentioned in #117), the redo seemed to work.

jntrnr:
Looks good, and sounds like @sholderbach's tests went well. Let's land and keep improving this.

PR #110, opened 2021-07-19
Fix the history behavior (#94)

### LINKED ISSUE #94: [Bug] History command disappear after few seconds + cursor position not updated with history command len
Step to reproduce:
- Checkout main
- Cargo run 
- Type a command
- Use low arrow to navigate the history
- Wait few seconds

Registrazione schermo 2021-07-09 alle 22 04 12

sholderbach:
- [x] Repaint mode switching after the 1sec case
- [x] Back to history cursor with len for not used -> fixes the skipping the
most recent
- [x] Traversal and prefix search now work as expected
- [x] Ctrl-R working (not super elegant but browsable)

Closes #94

### DIFF
--- a/src/history/base.rs
+++ b/src/history/base.rs
@@ -1,8 +1,10 @@
 use std::collections::vec_deque::Iter;
 
+use crate::line_buffer::LineBuffer;
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum HistoryNavigationQuery {
-    Normal,
+    Normal(LineBuffer),
     PrefixSearch(String),
     SubstringSearch(String),
     // Suffix Search
--- a/src/history/file_backed.rs
+++ b/src/history/file_backed.rs
@@ -90,7 +90,7 @@ impl HistoryView for FileBackedHistory {
     fn forward(&mut self) {
         match self.query.clone() {
             HistoryNavigationQuery::Normal => {
-                if (self.cursor as isize) < self.entries.len() as isize - 1 {
+                if self.cursor < self.entries.len() {
                     self.cursor += 1;
                 }
             }
@@ -104,11 +104,10 @@ impl HistoryView for FileBackedHistory {
     }
 
     fn string_at_cursor(&self) -> Option<String> {
-        if self.entries.is_empty() {
-            return None;
-        }
-
-        let entry = self.entries[self.cursor].to_string();
+        let entry = match self.entries.get(self.cursor) {
+            Some(entry) => entry.to_owned(),
+            None => return None,
+        };
 
         match self.query.clone() {
             HistoryNavigationQuery::Normal => Some(entry),
@@ -299,11 +298,7 @@ impl FileBackedHistory {
 
     /// Reset the internal browsing cursor
     fn reset_cursor(&mut self) {
-        if self.entries.is_empty() {
-            self.cursor = 0
-        } else {
-            self.cursor = self.entries.len() - 1;
-        }
+        self.cursor = self.entries.len();
     }
 }
 
--- a/src/history/file_backed.rs
+++ b/src/history/file_backed.rs
@@ -222,13 +222,14 @@ impl FileBackedHistory {
 
     fn back_with_criteria(&mut self, criteria: &dyn Fn(&str) -> bool) {
         if !self.entries.is_empty() {
+            let previous_match = self.entries.get(self.cursor);
             if let Some((next_cursor, _)) = self
                 .entries
                 .iter()
                 .take(self.cursor)
                 .enumerate()
                 .rev()
-                .find(|x| criteria(x.1))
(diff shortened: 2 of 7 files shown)


mzanrosso:
@sholderbach - I'm glad someone is fixing it, tell me if I can help you in any way

mzanrosso:
@sholderbach - maybe it's useful to integrate with this attempt to fix by sherubthakur https://github.com/sherubthakur/reedline/commits/history-fix-cursor-update

sholderbach:
I rebased onto that. But stepped onto new weird interactions

src/history/file_backed.rs
```diff
@@ -218,50 +221,34 @@ impl FileBackedHistory {
...
+            if let Some((next_cursor, _)) = self
+                .entries
+                .iter()
+                .take(self.cursor)
+                .enumerate()
+                .rev()
+                .find(|x| criteria(x.1))
+            {
+                // set to entry
+                self.cursor = next_cursor
```
sherubthakur, line 234: This is so much better than what I had. Awesome job!

src/history/file_backed.rs
```diff
@@ -218,50 +221,34 @@ impl FileBackedHistory {
...
-                    break;
-                }
+        if !self.entries.is_empty() {
+            if let Some((next_cursor, _)) = self
+                .entries
+                .iter()
+                .take(self.cursor)
+                .enumerate()
+                .rev()
+                .find(|x| criteria(x.1))
```
sherubthakur, line 231: nit-pick: I find `.find(|(_idx, string)| criteria(string)` more readable than this. Thoughts?

src/history/base.rs
```diff
@@ -1,8 +1,10 @@
 use std::collections::vec_deque::Iter;
 
+use crate::line_buffer::LineBuffer;
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum HistoryNavigationQuery {
-    Normal,
+    Normal(LineBuffer),
```
sherubthakur, line 7: Do we really want the HistoryNavigation to be aware of `LineBuffer`? in my opinion `HistoryNavigation` stuff should be able to exist on it's own without any dependency on anything related to Reedline. In a sense that it is an independent module that any line editor can use.

src/history/file_backed.rs
```diff
@@ -89,8 +91,8 @@ impl HistoryView for FileBackedHistory {
 
     fn forward(&mut self) {
         match self.query.clone() {
-            HistoryNavigationQuery::Normal => {
-                if (self.cursor as isize) < self.entries.len() as isize - 1 {
+            HistoryNavigationQuery::Normal(_) => {
+                if self.cursor < self.entries.len() {
```
sherubthakur, line 95: let's say we have a vector of length 3, then the index value `3` is invalid. Right? We only have 0, 1, 2.

So shouldn't this be `self.entries.len() - 1` and not `self.entries.len()`?

sholderbach: We use the state `self.cursor == self.entries.len()` to denote we are not currently browsing the history. Thus this allows to leave the history and also fixes the issues of skipping the most recent entry on the first go. This is safe as `VecDeque::get()` returns `None` for the resulting out of bound access, which is desired to denote leaving the list of history entries.

src/history/base.rs
sholderbach: Sub-crate level modularity is currently not high on the list of design requirements. JT and I talked about this being probably the most elegant way to get the `bash` style recovery of the previously typed line buffer (if your cursor is not at the end for prefix search). This way we avoid additional state in the engine that might get out of sync. Also like this, we have a single authoritative line buffer in the `Reedline` engine all the other plugins can rely on. We probably don't have unlimited mental "state budget" for the engine, so we cheated that into this enum.

src/history/file_backed.rs
sholderbach: Will be necessary anyways to fix the consecutive match behavior back in ;)

sholderbach:
> I rebased onto that. But stepped onto new weird interactions

Post-mortem on that one: 

The interaction observed that reset the history browsing all the time was the introduction of the separate `Up` and `Down` edit commands for multiline, that were missed by the guard at the beginning of `run_edit_commands`

jntrnr:
Looks good!

src/history/base.rs
sherubthakur: > Sub-crate level modularity is currently not high on the list of design requirements.

Ok, if by sub-carte level modulatiry you mean modular design inside this crate then won't you say that this has the potential to hinder the development later down the road?
If you were hinting about history as a separate create then I did not mean it that way, I was just using that as an example on how I think we should envision the history module API.

Although I would say that I am not really sure what is the exact problem we are trying to solve here and if there even is a better way to do so.

PR #111, opened 2021-07-20
Outside world does not know about insertion point

sherubthakur:
A very small PR that hides the implementation details of line_buffer. Now no one knows how line_buffer is tracking insertion point.

### DIFF
--- a/src/completer.rs
+++ b/src/completer.rs
@@ -66,26 +66,25 @@ impl TabHandler for DefaultTabHandler {
             self.initial_line = LineBuffer::new();
             self.initial_line.set_buffer(line.get_buffer().into());
             self.initial_line
-                .set_insertion_point(line.insertion_point());
+                .set_insertion_point(line.line(), line.offset());
         } else {
             line.set_buffer(self.initial_line.get_buffer().into());
-            line.set_insertion_point(self.initial_line.insertion_point())
+            line.set_insertion_point(self.initial_line.line(), self.initial_line.offset())
         }
-        let completions = self.completer.complete(
-            self.initial_line.get_buffer(),
-            self.initial_line.insertion_point().offset,
-        );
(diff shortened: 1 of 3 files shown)


mzanrosso:
@sherubthakur - For me there is no problem with this change; let's say it's a stylistic change that can be there, like no. Let's see what others say

sholderbach:
What are the intended semantics for the `offset()` getter? The cursor over the total string or the column? The line buffer is definitely subject to change especially for a working vi mode.

sherubthakur:
> What are the intended semantics for the `offset()` getter? The cursor over the total string or the column?

`offset` here is the column. Do you think column/row would be better names than line/offset?

I am of the opinion that the total-string-offset idea should not really exist (at least not in the public API).

> The line buffer is definitely subject to change especially for a working vi mode.

Not really sure what you mean by this. Can you elaborate on this?

sherubthakur:
> @sherubthakur - For me there is no problem with this change; let's say it's a stylistic change that can be there, like no. Let's see what others say

I am not sure I follow this. Are you saying that there was no problem with the earlier code? or are you saying that this is a stylistic change? If you can elaborate on this, then I can answer this better.

The rationale behind this was to hide the internal details that we use in the line buffer, as I did not really feel like there was any reason to expose them. Here are some of the benefits that I can think of that we can gain here
- the probability that any change to line buffer that impacts use changing the other parts of the codebase decreases.
- You can swap out the internal details with a completely different implementation and (if the interface does not change) then you will not need to worry about changing anything in the other parts of the codebase.
- Fewer merge conflicts.

jntrnr:
Sure, sounds good. I'm tempted to totally remove InsertionPoint, but until then at least we can hide it more like this PR does. Thanks!

Issue #113, opened 2021-07-20
[Bug] history up arrow doesn't update cursor position


mzanrosso:
![Registrazione schermo 2021-07-20 alle 20 31 08](https://user-images.githubusercontent.com/51322105/126377167-add26d84-2ae3-4940-80b7-1a1202ffd5ba.gif)

sholderbach:
This is intended as the prefix search indicator, but if it causes a usability issue it is an easy fix in `Reedline::update_buffer_from_history()`

mzanrosso:
> This is intended as the prefix search indicator, but if it causes a usability issue it is an easy fix in `Reedline::update_buffer_from_history()`

I think it's not the behaviour one expect, if you complete the line with the history maybe you want insert other things but with this implementation you have to press x times the right arrow to be able to go further with the line

PR #115, opened 2021-07-20
Change cursor pos in prefix search to the end

sholderbach:
Approach discussed in #113


mzanrosso:
@sholderbach  - I tested it and it works, thanks for the quick fix

jntrnr:
Looks good!

PR #125, opened 2021-07-26
[StyleFix] change hinter color

mzanrosso:
How about changing the default hinter color ?

From:

Schermata 2021-07-26 alle 10 09 49
 To:
 
Schermata 2021-07-26 alle 10 10 28

I think that the default hinter should be less visible than the actual one.
We can choose any of this, actually I proposed the 249 --> https://upload.wikimedia.org/wikipedia/commons/1/15/Xterm_256color_chart.svg

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -64,7 +64,7 @@ fn main() -> Result<()> {
             DefaultHinter::default()
                 .with_completer(completer) // or .with_history()
                 // .with_inside_line()
-                .with_style(Style::new().italic().fg(Color::LightGray)),
+                .with_style(Style::new().italic().fg(Color::Fixed(249))),
         ));
 
     let prompt = DefaultPrompt::new(1);


sholderbach:
How about https://docs.rs/nu-ansi-term/0.34.0/nu_ansi_term/struct.Style.html#method.dimmed instead of setting an explicit color? This should also work in light terminals and is according to https://docs.rs/crossterm/0.20.0/crossterm/style/enum.Attribute.html#platform-specific-notes generally supported.

mzanrosso:
> How about https://docs.rs/nu-ansi-term/0.34.0/nu_ansi_term/struct.Style.html#method.dimmed instead of setting an explicit color? This should also work in light terminals and is according to https://docs.rs/crossterm/0.20.0/crossterm/style/enum.Attribute.html#platform-specific-notes generally supported.

I think that your idea is valid, but at this point, since we use the style in multiple occasion; why don't put some examples for customize the style object ? if you agree I can do that even today

fdncred:
I made this configurable in engine-q.

mzanrosso:
@fdncred Ok. Can you brief me quickly on engine-q? thanks

fdncred:
> @fdncred Ok. Can you brief me quickly on engine-q? thanks

Sure. [engine-q](https://github.com/nushell/engine-q) is our new engine for Nushell. it's nearly a complete rewrite of Nushell to fix some limitations and hard-to-fix bugs we ran into. It's the "engine" but the fact that internals changed quite a bit made us need to port all the old commands into it. See the [pinned issues](https://github.com/nushell/engine-q/issues) for a status update on it.

PR #128, opened 2021-07-28
Core editor

sherubthakur:
This is a very basic PR that pulls out the core editing functionality into a struct.

Currently, it handles
- LineBuffer
- CutBuffer

In the future, this opens up possibilities to
- pull undo/redo functionality out into this struct too #129 
- more cohesive units which will help the engine perform up and down commands (currently all the logic is in the engine)

# Caveats
Some decisions with the structure led me to use `pub line_buffer` and I have not tried to restructure those pieces in this PR. (I am not sure how to approach them - but will try to get to this in the near future)


sherubthakur:
Now all the functions related to editing just call editor from the engine. I think we can remove that indirection level as it does not add a lot of value. Initially when we were pulling these out then it added value as there was some logic in a bunch of these methods, now it is not the case.

src/core_editor/editor.rs
```diff
@@ -0,0 +1,189 @@
+use super::{Clipboard, LineBuffer};
+
+pub struct Editor {
+    // Hack: Fix this once coupling in the code is taken care of
+    pub line_buffer: LineBuffer,
```
sherubthakur, line 5: As mentioned in the caveat section: I don't really think this should be public at all, or that engine should even know if there is a line buffer under the hood (editor).

src/engine.rs
```diff
@@ -434,30 +428,30 @@ impl Reedline {
...
 
             // Move left until we're looking at the newline
             // Observe what column we were on
             while position > 0 && &buffer[(position - 1)..position] != "\n" {
-                self.line_buffer.move_left();
+                self.editor.move_left();
                 num_of_move_lefts += 1;
-                position = self.line_buffer.offset();
+                position = self.editor.offset();
             }
```
sherubthakur, line 455: I intend to pull some cohesive pieces out of this method into the line_buffer and the editor. Then use them here. Hopefully, we will be able to simplify this logic.

src/engine.rs
```diff
@@ -507,11 +501,11 @@ impl Reedline {
...
     fn set_history_navigation_based_on_line_buffer(&mut self) {
-        if self.line_buffer.is_empty()
-            || self.line_buffer.offset() != self.line_buffer.get_buffer().len()
-        {
-            self.history
-                .set_navigation(HistoryNavigationQuery::Normal(self.line_buffer.clone()));
+        if self.editor.is_empty() || self.editor.offset() != self.editor.get_buffer().len() {
+            self.history.set_navigation(HistoryNavigationQuery::Normal(
+                // Hack: Hight coupling point
+                self.editor.line_buffer.clone(),
```
sherubthakur, line 507: @sholderbach can you reiterate all the cases as to why we decided to use `LineBuffer` here? Let's see if we can find an alternative here

src/engine.rs
```diff
@@ -900,12 +854,12 @@ impl Reedline {
                     Event::Key(KeyEvent { code, modifiers }) => {
                         match (modifiers, code, self.edit_mode) {
                             (KeyModifiers::NONE, KeyCode::Tab, _) => {
-                                self.tab_handler.handle(&mut self.line_buffer);
+                                self.tab_handler.handle(&mut self.editor.line_buffer);
```
sherubthakur, line 857: I have to look into why line_buffer is required here too. I mean does it need it? Can I use editor instead? Can Tab handler become completely independent of line_buffer or editor and engine can wire up the required functionality?

src/core_editor/editor.rs
sholderbach: I would think of `LineBuffer` as the representation that is allowed to be `Clone` and `Send` whilst the engine with its undo capabilities would get out of step if that would be the case. Adding a setter and getter is certainly an option (Setting the line buffer to something new from outside (plugin etc.) would also have implications for the undo behavior)

sholderbach:
Which one of the PR's do you want to merge at the end. With the option to revert or pick commits the logically linear history from your refactoring efforts shouldn't be as big of an issue.

sherubthakur:
The order that I had in mind was #128 #129 then #130

src/core_editor/editor.rs
sherubthakur: I intended to pull undo/redo as editor specific features as you can see in #129 . Given all these pieces are encapsulated inside the editor. I don't really think there is the scope of getting anything get out of step.

sholderbach: A completion/suggestion might want to set the content of the line buffer to a new value, including a useful cursor position. In this case supplying a `LineBuffer` instance might be an easy api. Yet the undo stack managed by `Editor` would need to keep track that the linebuffer changed in a certain way to undo this completion. So a setter might be a better choice than the `pub line_buffer`

src/engine.rs
sholderbach: This is for the bash style navigation, if you start on an empty line and go up through the history, and go back down, you end up on the empty line. If you started on a line with content it would restore that if the history is exhausted (moving beyond the most recent entry). In the current implementation, this behavior can be observed if you have content on the buffer but the cursor not at the end for prefix search. By storing it in the `HistoryNavigationQuery::Normal` Variant we won't loose track of that previous state. The navigation mode can only reset if you start editing the buffer.

jntrnr:
This looks good

PR #129, opened 2021-07-28
Pull undo redo to editor

sherubthakur:
I wanted to create this on the branch that I have requested to merge but looks like I can't really do that as that branch does not exist in this repo. (Sigh: github should really allow for creating PRs over other PRs :P)

Anyway, this is a simplification of pulling out the undo and redo functionality into the editor struct whilst also simplifying it a bit.

**NOTE TO Reviewers: #128 does most of the work on which this build on. Only take a look at the last commit in this PR**

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -1,18 +1,29 @@
+use crate::core_editor::get_default_clipboard;
+
 use super::{Clipboard, LineBuffer};
 
 pub struct Editor {
     line_buffer: LineBuffer,
     cut_buffer: Box<dyn Clipboard>,
+
+    edits: Vec<LineBuffer>,
+    index_undo: usize,
+}
+
+impl Default for Editor {
+    fn default() -> Self {
+        Editor {
+            line_buffer: LineBuffer::new(),
+            cut_buffer: Box::new(get_default_clipboard()),
+
+            // Note: Using list-zipper we can reduce these to one fild
+            edits: vec![],
+            index_undo: 2,
+        }
+    }
 }
 
 impl Editor {
-    pub fn new(line_buffer: LineBuffer, clip_buffer: Box<dyn Clipboard>) -> Editor {
-        Editor {
-            line_buffer,
-            cut_buffer: clip_buffer,
-        }
-    }
-
     pub fn line_buffer(&mut self) -> &mut LineBuffer {
         &mut self.line_buffer
     }
@@ -128,20 +139,53 @@ impl Editor {
         self.line_buffer.is_empty()
     }
 
-    pub fn undo(&mut self) -> Option<()> {
-        self.line_buffer.undo()
+    pub fn reset_olds(&mut self) {
+        self.edits = vec![LineBuffer::new()];
+        self.index_undo = 2;
+    }
+
+    fn get_index_undo(&self) -> usize {
+        if let Some(c) = self.edits.len().checked_sub(self.index_undo) {
+            c
+        } else {
+            0
+        }
     }
 
     pub fn redo(&mut self) -> Option<()> {
-        self.line_buffer.redo()
+        if self.index_undo > 2 {
+            self.index_undo = self.index_undo.checked_sub(2)?;
+            self.undo()
+        } else {
+            None
+        }
     }
 
-    pub fn reset_olds(&mut self) {
-        self.line_buffer.reset_olds()
+    pub fn undo(&mut self) -> Option<()> {
+        self.line_buffer = self.edits.get(self.get_index_undo())?.clone();
+
+        if self.index_undo <= self.edits.len() {
+            self.index_undo = self.index_undo.checked_add(1)?;
+        }
+        Some(())
     }
 
     pub fn set_previous_lines(&mut self, is_after_action: bool) -> Option<()> {
-        self.line_buffer.set_previous_lines(is_after_action)
+        self.reset_index_undo();
+
+        if self.edits.len() > 1
+            && self.edits.last()?.word_count() == self.line_buffer.word_count()
+            && !is_after_action
+        {
+            self.edits.pop();
+        }
+        self.edits.push(self.line_buffer.clone());
+
+        Some(())
+    }
+
+    pub fn reset_index_undo(&mut self) {
+        self.index_undo = 2;
     }
 
     pub fn cut_from_start(&mut self) {
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -16,7 +16,7 @@ impl Default for Editor {
             line_buffer: LineBuffer::new(),
             cut_buffer: Box::new(get_default_clipboard()),
 
-            // Note: Using list-zipper we can reduce these to one fild
+            // Note: Using list-zipper we can reduce these to one field
             edits: vec![],
             index_undo: 2,
         }
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -22,9 +22,6 @@ impl Default for InsertionPoint {
 /// In memory representation of the entered line(s) to facilitate cursor based editing.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct LineBuffer {
-    old_lines: Vec<Vec<String>>,
-    old_insertion_point: Vec<InsertionPoint>,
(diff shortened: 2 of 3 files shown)


src/core_editor/editor.rs
```diff
@@ -125,20 +147,66 @@ impl Editor {
...
 
     pub fn set_previous_lines(&mut self, is_after_action: bool) -> Option<()> {
-        self.line_buffer.set_previous_lines(is_after_action)
+        self.reset_index_undo();
+
+        // NOTE: Not sure what we are trying to do here. @mazarano help?
+        if self.old_lines.len() > 1
+            && self.old_lines.last()?.concat().trim().split(' ').count()
+                == self.lines.concat().trim().split(' ').count()
+            && !is_after_action
```
sherubthakur, line 196: @mzanrosso Can you help me a bit here. Like can you explain what this check is for? I havent' really applied any brains in this PR. I just did one-to-one mapping of what was already present and what I came up with.

src/core_editor/editor.rs
```diff
@@ -1,16 +1,38 @@
...
+
+impl Default for Editor {
+    fn default() -> Self {
+        Editor {
+            line_buffer: LineBuffer::new(),
+            cut_buffer: Box::new(get_default_clipboard()),
+
+            // Note: Using list-zipper we can reduce these to one fild
+            edits: vec![],
+            index_undo: 2,
```
sherubthakur, line 22: I wasn't too sure why we are using 2 as an index here?

src/core_editor/editor.rs
```diff
@@ -125,20 +147,66 @@ impl Editor {
...
 
-    pub fn reset_olds(&mut self) {
-        self.line_buffer.reset_olds()
+    pub fn undo(&mut self) -> Option<()> {
+        // NOTE: Leaving the old logic to help with the review
+        // self.lines = self.old_lines.get(self.get_index_undo())?.clone();
+        // let insertion_point = *self.old_insertion_point.get(self.get_index_undo())?;
+        // self.set_insertion_point(insertion_point.line, insertion_point.offset);
+
+        self.line_buffer = self.edits.get(self.get_index_undo())?.clone();
```
sherubthakur, line 181: It does a bit of simplification IMO. (Instead or 2 arrays we only deal with one).

@sholderbach (https://discord.com/channels/601130461678272522/855886335980994600/868896677883420733) I think we can push it a bit more and have a list-zipper that way we will not even need to have the index tracking field. 
IMO ListZipper will remove the possibility of getting the index into an inconsistent state. Though I don't really know of any rust libs which have a list zipper.

(P.S. - I am a huge proponent of making impossible states unrepresentable, If I am going too far in this direction please do point it out)

src/core_editor/editor.rs
mzanrosso: @sherubthakur Imagine that you write "te", "tes" and finally "test"; basically we don't want to have 3 lines in the old_lines state vector but we want to update only the last word since we decided to have history by words and not by single character.

src/core_editor/editor.rs
mzanrosso: @sherubthakur Because of this line:

`self.lines = self.old_lines.get(self.get_index_undo())?.clone();`

The get requires 1 more and since when we initialize everything we insert a blank string as initial state for undo the index will result initialized as 2

src/core_editor/editor.rs
sherubthakur: Ok, so here is my take. if we introduce the following 
- `word count` in `LineBuffer` which gives the count of words in that buffer
- `last_buffer` which gives the last element of the edit buffer
- `current_buffer` which gives handle to the current line buffer
then this will look something like

```rust
if self.has_undo_history() && self.last_buffer()?.word_count() == self.current_buffer().word_count()
```

This would be equivalent to what you intended. Right?

Note: Thoughts on trying to pull the strategy of having swappable logic for how we decide to do history. e.g. word-based, line-based, character-based, etc. Though I am not sure how this will look like.

mzanrosso: @sherubthakur You miss the last condition `&& !is_after_action` but in theory should be fine

src/core_editor/editor.rs
sholderbach: How should the API for a list zipper look like in rust? Enum with legal tuples or Struct with datastructure and index encapsulated with the legal transitions as methods?

sherubthakur: ### Defining the List Zipper
```rust
struct ListZipper<A> {
    prev: Stack<A>,
    current: A,
    next: Stack<A>
}
```
This is the formulation that I had in mind. 
Note: This requires at-least one element to work. We can obviously use an `Option<A>` for the `current` field so that it can be empty too.

It can now support operations like 
`back` -> which `push`es `current` onto `prev` and `pop`s top from `next` to `current`
`forward` -> which `push`es `current` onto `next` and `pop`s top from `prev` to `current`

### Usage in relation to undo-redo
And `undo` basically corresponds to going `back` and redo corresponds to going `forward`.
Another API we would need is invalidating the `next` `Stack` when a user `undo`s and then starts typing new stuff.

> Enum with legal tuples or Struct with datastructure and index encapsulated with the legal transitions as methods?
Note: Given this structure, we should need an index 

### Pros
- State management is not happening in 2 places which can go out of sync
- No need for addition/subtraction hacks that we are doing currently in order to handle cases where index rules fail etc
- We don't really need the version which works with empty. So no need to handle the `Option<A>` field

### Cons
- There is no crate that has this (?), So having to implement this doesn't seem worthwhile enough to me (given it is just used in one place, if there was a lib that we could leverage then I think it would have been something worth exploring)

sholderbach: This struct could also be a simple wrapper over `Vec` and `index` with accessors and the valid operations, that we implement and test separately. If we consider that performance critical, this would probably suffice. The undo stack is most likely not a hot path and wouldn't require cache locality compared to the history where similar movement is performed but we want the search to be fast.

sherubthakur: > This struct could also be a simple wrapper over Vec and index with accessors and the valid operations, that we implement and test separately.

Yes, that will work too. Though we do miss out on the "impossible states being unrepresentable" bit, which is sort of the main argument in favour of using a List Zipper.

I want to reiterate that I would be fine with doing it the present way too where the Editor maintains the pointer. Moreover in the particular PR, I would suggest that we don't involve the ListZipper. And if in future the Editor becomes unwieldy then we can extract this abstraction out. 
The major reason I wanted to use ListZipper was it simplifies API + it is something that I assume existed and we would only have to import the lib. But given the lib does not exist I don't really find it worthwhile to have this abstraction. Given that Editor is simple enough right now.
Thoughts?

src/core_editor/line_buffer.rs
```diff
@@ -298,6 +240,10 @@ impl LineBuffer {
         }
     }
 
+    pub fn word_count(&self) -> usize {
+        self.lines.concat().trim().split(' ').count()
+    }
```
sherubthakur, line 245: For readability sake, I extracted this out. Though the exact name that is given here is a bit misleading as it does not handle all such cases and I have added a failing test that showcases this. Let me know what do we want to do here. 
Here are a few ways that I see we can tackle this scenario.

- Is this a worthwhile abstraction then I focus on fixing this using the test as a guide
- it isn't a worthwhile abstraction then we can just simply use it's internal detail at the caller's site. Given this is just an expression just replacing it as is should be sufficient.
- We don't worry about it as much at let is stay as is. (and open a good-first-issue to fix this)

jntrnr: You could use `split_whitespace` instead of `.split(' ')` and I think that should be enough to pass the test you mention.

jntrnr:
Looks good!

src/core_editor/editor.rs
jntrnr: My two cents: zippers are a cool technique, but I think I'm hesitant to use things like that because it assumes that other contributors will be familiar with them. The only context I've seen them used is functional programming, and I think not many people come from a functional programming background. Similar to how monads would assume functional programming background.

It's okay if a technique isn't "perfect" when it's easy for most contributors to go in, understand how it works quickly, and can fix issues with it pretty quickly.

Again, just my two cents.

PR #130, opened 2021-07-29
Cleanup

sherubthakur:
Building on top of previous PRs this just does some basic cleanup

- Remove un-necessary indirections from engine to editor functions
- Remove dead code - partial_command
- Make undo/redo interface consistent with other editor functions
- abstract up/down command logic to appropriate units

Reviewer Note: This is meant to be reviewed when the base is #129

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -350,80 +350,21 @@ impl Reedline {
 
     fn up_command(&mut self) {
         // If we're at the top, then:
-        if !self.editor.get_buffer()[0..self.editor.offset()].contains('\n') {
+        if !self.editor.is_cursor_at_first_line() {
             // If we're at the top, move to previous history
             self.previous_history();
         } else {
-            // If we're not at the top, move up a line in the multiline buffer
-            let mut position = self.editor.offset();
-            let mut num_of_move_lefts = 0;
-            let buffer = self.editor.get_buffer().to_string();
-
-            // Move left until we're looking at the newline
-            // Observe what column we were on
-            while position > 0 && &buffer[(position - 1)..position] != "\n" {
-                self.editor.move_left();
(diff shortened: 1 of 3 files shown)


src/engine.rs
```diff
@@ -698,40 +514,24 @@ impl Reedline {
             ]
```
sholderbach, line 514: (Also pinging @mzanrosso) should the addition to the undo stack be completely encapsulated by the `Editor` or do we envision cases where it would be better that the commands on the editor don't do it automatically to not get confused when plugins/completions start to edit?

src/engine.rs
```diff
@@ -355,130 +345,27 @@ impl Reedline {
...
-
-            self.set_offset(position);
-
-            // Move right from this position to the column we were at
-            while &buffer[position..(position + 1)] != "\n" && num_of_move_lefts > 0 {
-                self.line_buffer.move_right();
-                position = self.line_buffer.offset();
-                num_of_move_lefts -= 1;
-            }
+            self.editor.move_line_down()
```
sholderbach, line 368: Moving down is broken this way!
We only want to invoke the history move if we are on the last line.

src/engine.rs
mzanrosso: I quickly saw the changes and I think that currently nothing changes if we keep it on linebuffer or editor;  I should spend more time analyzing the code well to respond with a bit of foresight

src/engine.rs
sherubthakur: ahh yes! totally copy-pasted from `up` with a brain fog it seems and forgot to modify the condition which should read something like `!self.editor.is_cursor_at_last_line()`

jntrnr:
Nice to see it all come together. Looks good

PR #133, opened 2021-08-14
Less intrusive tab completion

sherubthakur:
This is a PR to implement an alternate approach to handle tab-handler resetting of completions.

The idea is that tab-handler maintains an internal state which it can compare the next completion to and based on that reset index (or not). This way we can avoid all the calls from engine to it.


sherubthakur:
Note: 
- The first commit just moves some files around (this is more of a taste preference) if others have anything against it I can merge this into a single file. I just like one unit and it's unit test in the file and have it imported in other places.
- The second commit adds tests to what I understood about the existing cases.
- The third commit adds a failing test that supports the functionality in question (internal state is cleared based on value in line buffer)
- The last commit implements the functionality and also uses it in engine. (I can split it as I was under the impression that I made two commits there )

src/completion/tab_handler.rs
```diff
@@ -146,4 +154,22 @@ mod test {
...
+        tab.handle(&mut buf);
+        let mut expected_buffer = buffer_with("that is my test th");
+        expected_buffer.set_insertion_point(0, 4);
+        assert_eq!(buf, expected_buffer);
+
+        // updating the cursor to end should reset the completions
+        buf.set_insertion_point(0, 18);
+        tab.handle(&mut buf);
+        assert_eq!(buf, buffer_with("that is my test that"));
+    }
```
sherubthakur, line 176: @jntrnr does this test capture the requirement you had in mind?

jntrnr: Yup, exactly

src/completion/tab_handler.rs
```diff
@@ -50,15 +50,14 @@ impl ComplationActionHandler for DefaultCompletionActionHandler {
...
-            self.initial_line = LineBuffer::new();
-            self.initial_line.set_buffer(line.get_buffer().into());
-            self.initial_line
-                .set_insertion_point(line.line(), line.offset());
+            self.initial_line = present_buffer.clone();
         } else {
-            line.set_buffer(self.initial_line.get_buffer().into());
-            line.set_insertion_point(self.initial_line.line(), self.initial_line.offset())
+            *present_buffer = self.initial_line.clone();
         }
```
sherubthakur: @mzanrosso Can you help me understand this logic better. Why do we need to do this?

jntrnr:
Looks good

PR #134, opened 2021-08-15
Modularization of the main struct

sherubthakur:
This is like a 2nd take at the ambitious ticket https://github.com/jntrnr/reedline/issues/61 which was opend way back. But this time the results are more close to what we (me, @jntrnr and @sholderbach ) had in mind. (Note: Well what I inferred what @jntrnr and @sholderbach meant).

## High level summary
- EditorMode related stuff is completely decoupled from the main stuct and is behind a contract set by a trait (InputParser)
- Adds another layer of Events (enum) called Reedline (subject to change). The idea is simple that the InputParser's parse the input we get from the users and turn that into events that Reedline understands.

## Potential benefits
- Completely decouples EditMode stuff and we can now develop them completely independently of what happens to the main struct (as long as we keep the contract)
- The addition of events can help us in creating stuff like multiple streams from different sync producers that are eventually parsed by the main struct. This should help in easy dev. of widgets like clock etc. (Though that is still a far fetched dream)

## Todo
- Reintegration of vi mode is still pending
- Tests

## Note to reviewers
I  think the best way to review this pull request is
- Look at the commit summary to get an idea of what is going on.
- Look at the whole thing together. Ignore most of the stuff and focus on the `src/engine.rs` and see if all the changes it's the internal structure and how it interacts with the new interfaces, a reasonable position to be in.
- After that focusing on the individual aspects in detail. 
I will try to add helpful comments on places where most attention is required.

Note: This is a massive PR, it would be completely understandable if you feel like it does not align with the project vision or if there is a possibility it can be broken down into smaller ones. I have opened this to get opinions from the maintainers.


src/engine.rs
```diff
@@ -32,6 +32,18 @@ struct PromptWidget {
     origin: (u16, u16),
 }
 
+enum ReedlineEvent {
```
sherubthakur: this is the main part of the design that I was thinking every Event generates this enum (definitely needs a better name). This will eventually help in decluttering the main struct yet be functionally equivalent. 

Clock poll generates an events
emacs mode generates events
vi mode generates events

then all these are handled in the same function. I want to make distinction between this and EditCommand. This is line-editor events (i.e. everything) where as I was going to reduce `EditCommand`s to core-editor specific operations only.

src/engine.rs
```diff
@@ -705,110 +713,110 @@ impl Reedline {
...
+
+                    let (new_column, _) = cursor::position()?;
+
+                    if new_column < original_column && original_row + 1 == self.terminal_rows() {
+                        // We have wrapped off bottom of screen, and prompt is on new row
+                        // We need to update the prompt location in this case
+                        let (prompt_offset_columns, prompt_offset_rows) = self.prompt_widget.offset;
+                        let (prompt_origin_columns, prompt_origin_rows) = self.prompt_widget.origin;
+                        self.set_prompt_offset((prompt_offset_columns, prompt_offset_rows - 1));
+                        self.set_prompt_origin((prompt_origin_columns, prompt_origin_rows - 1));
```
sherubthakur, line 766: This looks ugly. I will circle back and update the way this is handled.

src/engine.rs
```diff
@@ -32,6 +32,18 @@ struct PromptWidget {
...
 
+enum ReedlineEvent {
+    HandleTab,
+    CtrlD, // Don't know a better name for this
+    CtrlC, // Don't know a better name for this
+    ClearScreen,
+    Enter,
+    Mouse, // Fill in details later
+    Resize(u16, u16),
+    EditInsert(EditCommand), // HACK: Special handling for insert
```
sherubthakur: As mentioned it's a simple hack for now, I would not like a special case for InsertCommand

src/engine.rs
```diff
@@ -32,6 +32,18 @@ struct PromptWidget {
     origin: (u16, u16),
 }
 
+enum ReedlineEvent {
+    HandleTab,
+    CtrlD, // Don't know a better name for this
+    CtrlC, // Don't know a better name for this
```
sherubthakur: Let me know what the names for these operations are.

src/engine.rs
```diff
@@ -705,110 +713,110 @@ impl Reedline {
...
+                self.tab_handler.handle(&mut line_buffer);
+                Ok(None)
+            }
+            ReedlineEvent::CtrlD => {
+                if self.editor.is_empty() {
+                    self.editor.reset_olds();
+                    Ok(Some(Signal::CtrlD))
+                // } else if let Some(binding) = self.find_keybinding(modifiers, code) {
+                //     self.run_edit_commands(&binding);
+                //     Ok(None)
```
sherubthakur: The rationale to remove this was that input_parsing will handle conversion into what operation we want to perform and we care only about the end result here (i.e. what to do with the given action). 
Handling if keybindings are different is the responsibility of the input_parsing stage.

src/engine.rs
```diff
@@ -65,23 +65,25 @@ impl PromptWidget {
     }
 }
 
-trait EventParser {
+trait InputParser {
     fn parse_event(&self, event: Event) -> ReedlineEvent;
+    fn update_keybindings(&mut self, keybindings: Keybindings);
+    fn edit_mode(&self) -> EditMode;
 }
```
sherubthakur: This is a makeshift interface for now. It should remain the same in spirit only (i.e. it parses `Event`s to `ReedlineEvent`s) but the interface will definitely need updating.

src/engine.rs
```diff
@@ -883,31 +866,7 @@ impl Reedline {
...
-                        ReedlineEvent::Edit(vec![])
-                    }
-                }
-            },
-
-            Event::Mouse(_) => ReedlineEvent::Mouse,
-            Event::Resize(width, height) => ReedlineEvent::Resize(width, height),
-        };
-
-        Ok(editor_event)
```
sherubthakur: And finally, we achieve complete segregation of even parsing into a separate struct.

src/engine.rs
sherubthakur: Note: Now the user of this lib can develop their own InputParser's without any help from the lib. All they need to do is adhere to this interface and we will be done. (prompt handling might pose a caveat in this though)

sherubthakur:
Next Steps:

## Resolve hacks
In order to achieve this segregation, I took a few shortcuts that need resolution
- Special handling for the InsertChar command
- Setup proper keybindings for VI (This IMO requires some thought as with vi there is stuff like nmap, imap, vmap, etc what all do we want to support here?)

These hacks don't have any impact on end-user. Have a little bit of impact on code quality but is easily resolvable, once we have an agreed-upon solution.

## Pull non-editor commands out of `EditCommand` to `ReedlineEvent`
The following values of the enum `EditCommand` make more sense in `ReedlineEvent`
```rust
EditCommand::Clear
EditCommand::Up
EditCommand::Down
EditCommand::AppendToHistory
EditCommand::PreviousHistory
EditCommand::NextHistory
EditCommand::SearchHistory
```
Moreover, I think `PreviousHistory` and `NextHistory ` are not even required now with the introduction of `Up` and `Down` commands.

jntrnr:
@sherubthakur - did you want to do those next steps in this PR or a follow-up?

If you want to do them in a follow-up I can review this one with those remaining items in mind.

sherubthakur:
> @sherubthakur - did you want to do those next steps in this PR or a follow-up?
> 
> If you want to do them in a follow-up I can review this one with those remaining items in mind.

I would suggest reviewing this assuming that I will do a follow-up PR.
- My intention is that the hacks that I mentioned are resolved in this PR only
- Pulling the non-editor commands is up for discussion with the wider team and I will pick this up as a follow-up.

src/engine.rs
```diff
@@ -67,17 +69,13 @@ pub struct Reedline {
...
-    fn default() -> Self {
-        Self::new()
-    }
+    terminal_size: (u16, u16),
 }
 
 impl Reedline {
     /// Create a new [`Reedline`] engine with a local [`History`] that is not synchronized to a file.
-    pub fn new() -> Reedline {
+    pub fn new() -> io::Result<Reedline> {
```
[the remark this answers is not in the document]
jntrnr: a rule of thumb for `new()` is that it should be infallible. You might want to use a different name for the constructor here to show that it doesn't return Self but instead a fallible version of Self.

jntrnr (commented):
Looks good - just had some nits about constructor return types

jntrnr:
Looks good

PR #137, opened 2021-09-02
Single responsibility painter

sherubthakur:
This small PR pulls out the highlighter and hinter out of Painter.

- I am fairly confident that the generation of hints should not handle in the Painter.
- Highlighter I am not so sure about. I was thinking if we can create an interface for highlighter that is more general, then it should become part of the Painter. 

Let me know your thoughts on this one.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -81,14 +81,19 @@ pub struct Reedline {
     // Stdout
     painter: Painter,
 
+    // Edit Mode: Vi, Emacs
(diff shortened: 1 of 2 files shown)


jntrnr:
@sherubthakur - did you not want to go this direction?

sherubthakur:
> @sherubthakur - did you not want to go this direction?

Included in https://github.com/jntrnr/reedline/pull/138#issuecomment-913015309

PR #138, opened 2021-09-03
Pull some events from EditCommand to ReedlineEvent

sherubthakur:
I have done some refactors along the way. Let me know your thoughts on this.

This could have been 4-5 separate PRs, getting them reviewed/merged individually would have taken some time, so I just clubbed them together to save some time. (The squashing merge of PRs also makes it hard to rebase, otherwise, that could have saved a lot of trouble and I would have managed to do multiple PRs - This might just be a lack of knowledge on my side, if there is an easy way to rebase with the squashed stuff please do tell me.). 

Meanwhile here are all the things that this PR tries to do

## Extraction of `Hinter` and `Highlighter` out of `Painter`

`Hinter` has logic that is more than just display specific so it does not make sense for it to be a part of `Painter` (A component that is supposed to only care about screen painting operations). Though the same can not be said about the `Highlighter`. The `Highlighter` does deal with display elements, though currently, it understands what a `LineBuffer` is, which makes me skeptical about its presence in `Painter`.
Maybe we can have an interface with `Painter` where we supply 3-tuple (kinda) object `(before_cursor_string, virtual_text/hint, after_cursor_string)`, then the `Highlighter` can do anything with this info whilst residing inside the `Painter`.

## Change `Keybinding` structure to use use  `ReedlineEvent` instead of `Vec<EditCommand>` 

Given that the `Keybinding` operates on more than the editor stuff (e.g. history, etc)  it should follow that its internal structure reflects this.  `ReedlineEvent` (in need of a better name) is the `enum` that fits this bill IMO.

## Pull non-editor commands out of EditCommand to ReedlineEvent

As mentioned in https://github.com/jntrnr/reedline/pull/134#issuecomment-904327380  the following 
```rust
EditCommand::Clear
EditCommand::Up
EditCommand::Down
EditCommand::AppendToHistory
EditCommand::PreviousHistory
EditCommand::NextHistory
EditCommand::SearchHistory
```
make more sense in `ReedlineEvent` than they do in `EditCommand`, for the reason that these aren't really editor operations which `EditCommand` should be interested in.

## Pull history search-related specific decisions to the top level.

History search operations are too deep down in the code which makes it a high coupling piece. I wanted to try and reduce that coupling and as a result, I have pulled this up.
Another way to think about this is that when we start history search by pressing Ctrl-R (or equivalent) we essentially start another editor with limited functionality and the result from this is carried over to any line-editors buffer. (Kinda like how shell and line-editor interact)
Maybe eventually we can have a shared editor between struct (different instances) thus providing best in class history search experience with minimal effort from our side. Thoughts?
But currently, The distinction is made between how the handling of events happens at the top level (instead of the old way where it was taking place nestled down deep)

## Resolve `InsertChar` hack in `ReedlineEvent`

Initially, I added a hack (special brach in `ReedlineEvent` enum to handle `InsertChar`) to simplify pulling handle-event into a separate function. So this just fixes that hack.

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -36,7 +36,7 @@ impl EditMode for Emacs {
                 (KeyModifiers::NONE, KeyCode::Enter) => ReedlineEvent::Enter,
                 _ => {
                     if let Some(binding) = self.keybindings.find_binding(modifiers, code) {
-                        ReedlineEvent::Edit(binding)
+                        binding
                     } else {
                         ReedlineEvent::Edit(vec![])
                     }
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -31,7 +31,7 @@ impl EditMode for Emacs {
                 (KeyModifiers::CONTROL, KeyCode::Char('l')) => ReedlineEvent::ClearScreen,
                 (KeyModifiers::NONE, KeyCode::Char(c))
                 | (KeyModifiers::SHIFT, KeyCode::Char(c)) => {
-                    ReedlineEvent::EditInsert(EditCommand::InsertChar(c))
+                    ReedlineEvent::Edit(vec![EditCommand::InsertChar(c)])
                 }
                 (KeyModifiers::NONE, KeyCode::Enter) => ReedlineEvent::Enter,
                 _ => {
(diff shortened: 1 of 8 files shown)


jntrnr:
Looks good.

Make me wonder what other things we can put into ReedlineEvent.

PR #139, opened 2021-09-04
Modifier for altgr

elferherrera:
Added clause to consider special modifier for non english keyboards


src/edit_mode/emacs.rs
```diff
@@ -33,6 +33,9 @@ impl EditMode for Emacs {
                 | (KeyModifiers::SHIFT, KeyCode::Char(c)) => {
                     ReedlineEvent::EditInsert(EditCommand::InsertChar(c))
                 }
+                (m, KeyCode::Char(c)) if m == KeyModifiers::CONTROL | KeyModifiers::ALT => {
```
jntrnr, line 36: Since we look for Ctrl+c a few lines above this one, should we just look for ALT here?

jntrnr:
Just fixed CI for reedline, so in theory the next change should pick it up

src/edit_mode/emacs.rs
elferherrera: The thing is that in my keyboard the key `alt gr` appear as a mixed KeyModifier called `CONTROL | ALT`. This isnt a Control or Alt situation, it has to be both

jntrnr: Oooh as a bitwise operation. Got it. I totally misread that this morning.

PR #140, opened 2021-09-04
Fix double chars

jntrnr:


src/lib.rs
```diff
@@ -13,17 +13,20 @@
 //!  let prompt = DefaultPrompt::default();
 //!
 //!  loop {
-//!      let sig = line_editor.read_line(&prompt).unwrap();
+//!      let sig = line_editor.read_line(&prompt);
 //!      match sig {
```
mzanrosso, line 17: How about compact this match avoiding creating sig variable ?

PR #143, opened 2021-09-04
[BugFix] undo function

mzanrosso:
This commit solves the problem that when the undo function is performed,
for some reason the buffer remained with 1 character.

The FIX consists in correcting the logic of the current implementation
by bringing it to the initial design state; Where the vector of the
buffer had been initialized with an empty entry so that when performing
many times the undo brought you back to an initial condition with
the blank buffer rather than show you the first character.


jntrnr:
@mzanrosso - can you add a test to ensure this keeps working in the future?

mzanrosso:
@jntrnr I'm trying

mzanrosso:
@jntrnr Done.

src/core_editor/editor.rs
```diff
@@ -196,6 +200,31 @@ impl Editor {
         Some(())
     }
 
+    /// ```
+    /// use reedline::{Editor, LineBuffer};
+    /// let mut editor = Editor::default();
+    /// editor.line_buffer().set_buffer(String::from("a"));
+    /// editor.set_previous_lines(false);
+    ///
```
mzanrosso, line 208: If you want we can reduce the space there, avoiding blank lines.

src/core_editor/editor.rs
```diff
@@ -17,13 +17,17 @@ impl Default for Editor {
...
             // Note: Using list-zipper we can reduce these to one field
-            edits: vec![],
+            edits: vec![LineBuffer::new()],
             index_undo: 2,
         }
     }
 }
 
 impl Editor {
+    pub fn get_edits(self) -> Vec<LineBuffer> {
```
sherubthakur, line 27: Is this method only used for testing? or are there any other use cases for this?

If it is the former then I would recommend not adding this method at all. The reason for that being: any addition for the public interface ties us to a contract which can make it harder to change the internal details of the particular sturct/class/object/etc. 

For testing we can simply write a unit test in this module which will not require us to add any method to the public interface. 

Thoughts?

mzanrosso: Propose your change in a branch and if it's a better solutions i will modify this PR, thanks.

src/core_editor/editor.rs
```diff
@@ -196,6 +200,28 @@ impl Editor {
...
+    ///
+    /// assert_eq!(
+    ///     vec![
+    ///         LineBuffer::from(""),
+    ///         LineBuffer::from("ab "),
+    ///         LineBuffer::from("ab c")
+    ///     ],
+    ///     editor.get_edits()
+    /// );
+    /// ```
```
sherubthakur, line 224: @mzanrosso all I am proposing is that this doc test can be made into a unit test. That way we don't need to add the `editor.get_edits()` method to the public interface that `Editor` exposes. 
I will try to also add a PR as soon as I am able to.

sholderbach:
I made the changes to the test according to the suggestions by @sherubthakur and made sure to merge in the main branch to pass the tests.

PR #146, opened 2021-09-08
Add tests to emacs and vi edit mode

sherubthakur:
This PR intends to add basic tests to the emacs edit mode.

There are two tests that are marked ignore, which I am unsure what they should be. Please add details on the desired behaviuor.

I personally think that 
- we should be able to override keybinds so it will require some work on the `KeyBinding` struct.
- I am not too sure what the default behaviour should be in-case of Events that don't have any definition. I was thinking of returning a default IGNORE event. Ideally, we should log the details along with doing nothing for that. Would like to know what others think about this.

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -138,4 +138,20 @@ mod test {
         // Right now this returns ReedlineEvent::Edit(vec![]), but what should we really return
         assert_eq!(result, ReedlineEvent::HandleTab);
     }
+
+    #[test]
+    fn inserting_capital_character_for_non_ascii_remains_as_is() {
+        let mut emacs = Emacs::default();
+
+        let uppercase_l = Event::Key(KeyEvent {
+            modifiers: KeyModifiers::SHIFT,
+            code: KeyCode::Char('😀'),
+        });
+        let result = emacs.parse_event(uppercase_l);
+
+        assert_eq!(
+            result,
+            ReedlineEvent::Edit(vec![EditCommand::InsertChar('😀')])
+        );
+    }
 }
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -35,7 +35,7 @@ impl EditMode for Emacs {
                 _ => self
                     .keybindings
                     .find_binding(modifiers, code)
-                    .unwrap_or_else(|| ReedlineEvent::Edit(vec![])),
+                    .unwrap_or(ReedlineEvent::None),
             },
 
             Event::Mouse(_) => ReedlineEvent::Mouse,
@@ -123,9 +123,8 @@ mod test {
         );
     }
 
-    #[ignore = "Unsure what the desired behaviour is"]
     #[test]
-    fn return_IDONT_KNOW_WHAT_when_keybinding_is_not_found() {
+    fn return_none_reedline_event_when_keybinding_is_not_found() {
         let keybindings = Keybindings::default();
 
         let mut emacs = Emacs::new(keybindings);
@@ -135,8 +134,7 @@ mod test {
         });
         let result = emacs.parse_event(ctrl_l);
 
-        // Right now this returns ReedlineEvent::Edit(vec![]), but what should we really return
-        assert_eq!(result, ReedlineEvent::HandleTab);
+        assert_eq!(result, ReedlineEvent::None);
     }
 
     #[test]
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -72,7 +72,6 @@ mod test {
         assert_eq!(result, ReedlineEvent::ClearScreen);
     }
 
-    #[ignore = "Unsure what the desired behaviour is"]
     #[test]
     fn overriding_default_keybindings_works() {
         let mut keybindings = default_emacs_keybindings();
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -25,8 +25,11 @@ impl EditMode for Emacs {
     fn parse_event(&mut self, event: Event) -> ReedlineEvent {
         match event {
             Event::Key(KeyEvent { code, modifiers }) => match (modifiers, code) {
-                (m, KeyCode::Char(c)) if m == KeyModifiers::NONE | KeyModifiers::SHIFT => {
+                (KeyModifiers::NONE, KeyCode::Char(c)) => {
                     ReedlineEvent::Edit(vec![EditCommand::InsertChar(c)])
+                }
+                (KeyModifiers::SHIFT, KeyCode::Char(c)) => {
+                    ReedlineEvent::Edit(vec![EditCommand::InsertChar(c.to_ascii_uppercase())])
                 }
                 (KeyModifiers::NONE, KeyCode::Enter) => ReedlineEvent::Enter,
                 _ => self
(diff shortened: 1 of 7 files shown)


src/edit_mode/vi/mod.rs
```diff
@@ -158,3 +162,96 @@ impl Vi {
...
+    #[case('k', ReedlineEvent::Up)]
+    #[case('l', edit_event(EditCommand::MoveRight))]
+    #[case('u', edit_event(EditCommand::Undo))]
+    #[case('p', edit_event(EditCommand::PasteCutBuffer))]
+    #[case('w', edit_event(EditCommand::MoveWordRight))]
+    #[case('b', edit_event(EditCommand::MoveWordLeft))]
+    #[case('0', edit_event(EditCommand::MoveToStart))]
+    #[case('$', edit_event(EditCommand::MoveToEnd))]
+    #[case('A', edit_event(EditCommand::MoveToEnd))] // Not checking if it also moves to end
+    #[case('D', edit_event(EditCommand::CutToEnd))]
```
sherubthakur, line 192: I really like how neatly and easily we can write tests with rstest. Let me know what others think.

Aside: One nice thing about these tests is that even if we switch to a different vi parsing logic (as outlined in parser.rs) this interface is not going to change and these tests will all remain valid and hopefully might help in dome debugging.

src/core_editor/line_buffer.rs
```diff
@@ -541,26 +541,67 @@ mod test {
...
-    #[test]
-    #[ignore] // Note: Not sure if this is the intended behaviour
-    fn uppercase_word_works_when_one_last_index() {
-        let mut line_buffer = buffer_with("This is a test");
-        line_buffer.uppercase_word();
+    #[rstest]
+    // FIXME: capitalize_char move insertion point to the end
+    #[case("This is a test", (0, 13), "This is a tesT", (0, 14))]
+    // FIXME: capitalize_char move insertion point to the end
+    #[case("This is a test", (0, 10), "This is a Test", (0, 14))]
```
sherubthakur, line 548: Should the cursor position change when doing capitalization?

src/core_editor/line_buffer.rs
```diff
@@ -541,26 +541,67 @@ mod test {
...
 
-    #[test]
-    fn uppercase_word_works() {
-        let mut line_buffer = buffer_with("This is a test");
-        line_buffer.move_word_left();
+    #[rstest]
+    // FIXME: uppercase_word does not work when on penultimate location
+    // #[case("This is a test", (0, 13), "This is a TEST", (0, 13))]
+    // FIXME: uppercase_word updates the insertion point location
+    #[case("This is a test", (0, 10), "This is a TEST", (0, 14))]
```
sherubthakur, line 569: - Should the cursor position change
- This updates the words after the point instead of the word

src/core_editor/line_buffer.rs
```diff
@@ -541,26 +541,67 @@ mod test {
...
 
-        assert_eq!(expected_line_buffer, line_buffer);
+        assert_eq!(expected, line_buffer);
+    }
+
+    #[rstest]
+    // FIXME: uppercase_word does not work when on penultimate location
+    // #[case("This is a TEST", (0, 13), "This is a test", (0, 13))]
+    // FIXME: uppercase_word updates the insertion point location
+    #[case("This is a TEST", (0, 10), "This is a test", (0, 14))]
```
sherubthakur, line 590: Same comments as for uppercase word

src/core_editor/line_buffer.rs
```diff
@@ -578,4 +619,87 @@ mod test {
 
         assert_eq!(4, line_buffer.word_count());
     }
+
+    #[rstest]
+    #[case("This is a test", (0, 13), "This is a tets", (0, 14))]
+    #[case("This is a test", (0, 14), "This is a test", (0, 14))]
+    // FIXME: swap_graphemes does not work with first index
+    // #[case("This is a test", (0, 0), "hTis is a test", (0, 0))]
```
sherubthakur, line 627: - This function is a no-op on the first location?
- Which direction should this work in? (left/right or have different methods for each)?
- How should the cursor behave on this one?

src/core_editor/line_buffer.rs
```diff
@@ -578,4 +619,87 @@ mod test {
...
+        let mut expected = buffer_with(output);
+        expected.set_insertion_point(out_location.0, out_location.1);
+
+        assert_eq!(line_buffer, expected);
+    }
+
+    #[rstest]
+    #[case("This is a test", (0, 9), "This is test a", (0, 14))]
+    // FIXME: Shouldn't this be a no-op
+    // #[case("This is a test", (0, 14), "This is a test", (0, 14))]
```
sherubthakur, line 648: same as swap graphenes!

src/core_editor/line_buffer.rs
```diff
@@ -578,4 +619,87 @@ mod test {
...
+    //     let mut line_buffer = buffer_with(input);
+    //     line_buffer.set_insertion_point(in_location.0, in_location.1);
+    //
+    //     line_buffer.move_line_down();
+    //
+    //     let mut expected = buffer_with(output);
+    //     expected.set_insertion_point(out_location.0, out_location.1);
+    //
+    //     assert_eq!(line_buffer, expected);
+    // }
```
sherubthakur, line 704: I think we have only done a partial move to only string representation which is causing tests to fail. We should outline what all is the expectation here.

src/edit_mode/emacs.rs
```diff
@@ -62,3 +54,104 @@ impl Emacs {
...
+        let ctrl_l = Event::Key(KeyEvent {
+            modifiers: KeyModifiers::CONTROL,
+            code: KeyCode::Char('l'),
+        });
+        let result = emacs.parse_event(ctrl_l);
+
+        assert_eq!(result, ReedlineEvent::ClearScreen);
+    }
+
+    #[ignore = "Unsure what the desired behaviour is"]
```
sherubthakur, line 75: I think we should be able to override default keybindings

src/edit_mode/emacs.rs
```diff
@@ -62,3 +54,104 @@ impl Emacs {
...
+        });
+        let result = emacs.parse_event(uppercase_l);
+
+        assert_eq!(
+            result,
+            ReedlineEvent::Edit(vec![EditCommand::InsertChar('L')])
+        );
+    }
+
+    #[ignore = "Unsure what the desired behaviour is"]
```
sherubthakur, line 126: either we can 
- return Option<ReedlineEvent>
- add some default (like NoOp) branch to ReedlineEvent
- return Result<ReedlineEvent> with a custom error type detailing what the problem is

I am not sure what the approach here should be. Thoughts?

src/edit_mode/vi/mod.rs
```diff
@@ -17,15 +17,24 @@ enum Mode {
...
+    Keybindings::new()
+}
+
 impl Default for Vi {
     fn default() -> Self {
         Vi {
-            // FIXME: Setup proper keybinds
-            keybindings: default_emacs_keybindings(),
+            insert_keybindings: default_vi_insert_mode_keybindings(),
+            normal_keybindings: default_vi_normal_mode_keybindings(),
```
sherubthakur, line 37: I think we always had different kebinds for normal and insert mode. So, readded this. Using emacs binds in insert mode, not sure if this is the best way or of any alternatives.

src/core_editor/line_buffer.rs
jntrnr: I was thinking it should, like `~` in vim

src/core_editor/line_buffer.rs
jntrnr: the best way is to check against emacs and/or a shell with emacs bindings

src/core_editor/line_buffer.rs
jntrnr: Can you say a bit more about what's missing? I'm not up to speed

src/edit_mode/emacs.rs
jntrnr: we should bake in a default keybinding. For things where there isn't a keybinding, I think we should return None.

src/edit_mode/vi/mod.rs
jntrnr: insert mode is just for vim. Emacs has its own mode - I wonder why it used to use that for vim

jntrnr (commented):
I like the test cleanups. just left a few comments

src/core_editor/line_buffer.rs
sherubthakur: In vim `~` capitalizes then move left 1 step, whereas this function capitalizes and then move to the **end**. 

Also, I am not so sure what approach should the editor take like we can have one where we do two things 
- capitalize and then move left 
- (or) only capitalize and leave the moving left to the caller.

Thoughts on why we should pick one over the other?

src/edit_mode/vi/mod.rs
sherubthakur: I think this was something that was introduced in the massive refactor I did, and I missed this piece, I will extract the default vi map and get it here.

jntrnr:
looks good

PR #148, opened 2021-09-23
Multiline improvements

jntrnr:
This makes working at the last line during a multiline a bit cleaner. Both up/down and new lines should now flow to the next line and properly scroll the input if you're at the bottom of the terminal.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -143,8 +143,16 @@ impl Editor {
         self.line_buffer.line()
     }
 
+    pub fn num_lines(&self) -> usize {
+        self.line_buffer.num_lines()
+    }
+
     pub fn is_empty(&self) -> bool {
         self.line_buffer.is_empty()
+    }
+
+    pub fn ends_with(&self, c: char) -> bool {
+        self.line_buffer.ends_with(c)
     }
 
     pub fn is_cursor_at_first_line(&self) -> bool {
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -108,8 +108,8 @@ impl Editor {
         self.line_buffer.swap_graphemes();
     }
 
-    pub fn set_insertion_point(&mut self, line: usize, pos: usize) {
-        self.line_buffer.set_insertion_point(line, pos)
+    pub fn set_insertion_point(&mut self, pos: usize) {
+        self.line_buffer.set_insertion_point(pos)
     }
 
     pub fn get_buffer(&self) -> &str {
@@ -239,8 +239,7 @@ impl Editor {
             self.cut_buffer
                 .set(&self.line_buffer.get_buffer()[cut_range.clone()]);
             self.clear_range(cut_range);
-            self.line_buffer
-                .set_insertion_point(self.line_buffer.line(), left_index);
+            self.line_buffer.set_insertion_point(left_index);
         }
     }
 
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -3,13 +3,12 @@ use {std::ops::Range, unicode_segmentation::UnicodeSegmentation};
 /// Cursor coordinates relative to the Unicode representation of [`LineBuffer`]
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 struct InsertionPoint {
-    line: usize,
     offset: usize,
 }
 
 impl InsertionPoint {
     pub fn new() -> Self {
-        Self { line: 0, offset: 0 }
+        Self { offset: 0 }
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/core_editor/editor.rs
```diff
@@ -108,8 +108,8 @@ impl Editor {
         self.line_buffer.swap_graphemes();
     }
 
-    pub fn set_insertion_point(&mut self, line: usize, pos: usize) {
-        self.line_buffer.set_insertion_point(line, pos)
+    pub fn set_insertion_point(&mut self, pos: usize) {
+        self.line_buffer.set_insertion_point(pos)
```
sherubthakur, line 112: What is the interface here?

- do we have the idea of a present line and on that present line we are adding an insertion point?
- (or) are we saying that the outside world knows that internally we store this as a String and we get an index in that string?

src/core_editor/line_buffer.rs
```diff
@@ -35,50 +34,49 @@ impl Default for LineBuffer {
...
+
+    /// Calculates the current the user is on
+    pub fn line(&self) -> usize {
+        let offset = self.insertion_point.offset;
+        let count = self.lines[0..offset].lines().count();
+        if count == 0 {
+            0
+        } else {
+            count - 1
+        }
```
sherubthakur, line 72: How about
```rust
max(0, count - 1)
```

src/core_editor/line_buffer.rs
```diff
@@ -165,21 +160,20 @@ impl LineBuffer {
...
-        self.lines.push(String::new());
+        self.lines = String::new();
         self.insertion_point = InsertionPoint::new();
     }
 
     /// Clear everything beginning at the cursor to the right/end.
     /// Keeps the cursor at the end.
     pub fn clear_to_end(&mut self) {
-        self.lines[self.insertion_point.line].truncate(self.insertion_point.offset);
+        self.lines.truncate(self.insertion_point.offset);
```
sherubthakur, line 190: Won't this delete the whole thing? Can we add a test for this one?

src/core_editor/line_buffer.rs
```diff
@@ -206,11 +200,11 @@ impl LineBuffer {
     where
```
sherubthakur, line 214: Given that this function operates on the current line maybe its name should reflect that. Thoughts?
`replace_range_in_current_line`.
Well ideally this comment applies to all the functions that operate on current line.

src/core_editor/line_buffer.rs
```diff
@@ -479,36 +470,24 @@ mod test {
...
+        let before_operation_location = InsertionPoint { offset: 11 };
         assert_eq!(before_operation_location, line_buffer.insertion_point());
 
         line_buffer.set_buffer("new line 1\nnew_line 2".to_string());
 
-        let after_operation_location = InsertionPoint {
-            line: 1,
-            offset: 10,
-        };
+        let after_operation_location = InsertionPoint { offset: 10 };
```
sherubthakur: This does not seem right. The old test was saying that the line is first and the location on that line is 10. Maybe let's assert for the line being 1 here too. Thoughts?

src/core_editor/line_buffer.rs
```diff
@@ -72,6 +72,20 @@ impl LineBuffer {
...
     }
 
+    pub fn num_lines(&self) -> usize {
+        let count = self.lines.split('\n').count();
+
+        if count == 0 {
+            1
+        } else {
+            count
+        }
```
sherubthakur, line 82: `max(1, count)` ?

src/core_editor/line_buffer.rs
sherubthakur: Ignore this one, looks like the last commit fixes this.

src/core_editor/editor.rs
jntrnr: My assumption is that lines is a concept of the editor but not really of the outside world. We add a call that lets you set the insertion point to the closest it can find a row+col, also. This would be best effort, since the line may not have the column requested (or we may not have the line requested), but it could be one way of allowing someone to use it and build something more editor-like from it.

src/core_editor/line_buffer.rs
jntrnr: If count is 0 that would panic

src/core_editor/line_buffer.rs
jntrnr: Not sure what you mean here

src/core_editor/line_buffer.rs
jntrnr: I moved the functionality away from thinking in lines to thinking in ranges in the buffer. Offsets then are from the start of the buffer rather than the start of the line.

I wasn't able to find anywhere where the logic in the line editing part of reedline assumed a kind of 2-dimentional edit, but seemingly almost always just worked on the buffer given.

PR #149, opened 2021-09-23
Incomplete brackets

elferherrera:
Added check for incomplete brackets

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -50,6 +50,11 @@ impl EditMode for Vi {
                 (Mode::Insert, KeyModifiers::NONE, KeyCode::Char(c)) => {
                     ReedlineEvent::Edit(vec![EditCommand::InsertChar(c)])
                 }
+                (Mode::Insert, m, KeyCode::Char(c))
+                    if m == KeyModifiers::CONTROL | KeyModifiers::ALT =>
+                {
+                    ReedlineEvent::Edit(vec![EditCommand::InsertChar(c)])
+                }
                 (Mode::Insert, KeyModifiers::SHIFT, KeyCode::Char(c)) => {
                     ReedlineEvent::Edit(vec![EditCommand::InsertChar(c.to_ascii_uppercase())])
                 }
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -50,6 +50,9 @@ impl EditMode for Vi {
                 (Mode::Insert, KeyModifiers::NONE, KeyCode::Char(c)) => {
                     ReedlineEvent::Edit(vec![EditCommand::InsertChar(c)])
                 }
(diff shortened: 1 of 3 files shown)


src/validator.rs
```diff
@@ -19,10 +19,32 @@ pub struct DefaultValidator;
...
-        if line.split('"').count() % 2 == 0 {
+        if line.split('"').count() % 2 == 0 || incomplete_brackets(line) {
             ValidationResult::Incomplete
         } else {
             ValidationResult::Complete
         }
     }
 }
+
+fn incomplete_brackets(line: &str) -> bool {
```
sherubthakur, line 30: let's add some tests for this function.

src/validator.rs
```diff
@@ -19,10 +19,32 @@ pub struct DefaultValidator;
...
+    let mut balance: Vec<char> = Vec::new();
+
+    for c in line.chars() {
+        if c == '{' {
+            balance.push('}')
+        } else if c == '[' {
+            balance.push(']')
+        } else if c == '(' {
+            balance.push(')')
```
sherubthakur, line 39: how about -
```rust
if ['{', '[', '('].contains(&c) {
   balance.push(c)
else if ['}', ']', ')'].contains(&c) {
   ...
}
```
I just find that the if and else blocks look similar (complimentary?) in this one.

src/edit_mode/vi/mod.rs
```diff
@@ -50,6 +50,11 @@ impl EditMode for Vi {
                 (Mode::Insert, KeyModifiers::NONE, KeyCode::Char(c)) => {
                     ReedlineEvent::Edit(vec![EditCommand::InsertChar(c)])
                 }
+                (Mode::Insert, m, KeyCode::Char(c))
+                    if m == KeyModifiers::CONTROL | KeyModifiers::ALT =>
```
sherubthakur, line 54: What is this bitwise or of CONTROL and ALT here mean? Do you folks think it would be helpful if we add a comment explaining this? or a test case.

sherubthakur (commented):
I am not sure I understand the purpose of validator right now, but I had a few thoughts on the PR, if you can address them, that would be super helpful.

src/edit_mode/vi/mod.rs
sherubthakur: I think I might have stupidly deleted this during a refactor. A comment or a test case would have stopped me from doing this :P

elferherrera: Yeap, it was deleted and making non american keyboards not work correctly. The issue is that some keyboards need the special key `alt gr` which happens to be a combination of ALT and CTR modifiers. That's why it was there before

src/validator.rs
elferherrera: They are not similar. If you look at the cases, each insert the opposite pair to the condition. Probably it would be a good idea to use a enum and move those cases to the implementations of the enum. However, I wanted to keep the default validator simple and very explicit

jntrnr:
Looks good, thanks!

PR #154, opened 2021-09-27
Experimenting with Undo redo

sherubthakur:
The idea here is to make engine not have to worry about the internals of undo-redo (`set_previous_lines`). 
and 2nd have an abstraction that encapsulates index handling etc, out of editor
and 3rd to enable various pluggable strategies for what to track in the undo stack. 

NOTE: THIS IS NOT TESTED YET!

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -5,10 +5,33 @@ use super::{
     Clipboard, LineBuffer,
 };
 
+#[allow(dead_code)]
+fn undo_strategy_allow_all(
+    undo_stack: &mut Box<dyn UndoStack<LineBuffer>>,
+    latest_entry: &LineBuffer,
+) {
+    undo_stack.insert(latest_entry.clone());
+}
+
+fn undo_strategy_club_similar(
+    undo_stack: &mut Box<dyn UndoStack<LineBuffer>>,
+    latest_entry: &LineBuffer,
+) {
+    // TODO: Add more intelligence
+    if undo_stack.current().word_count() == latest_entry.word_count() {
+        undo_stack.undo();
+    }
+    undo_stack.insert(latest_entry.clone())
+}
+
+type UndoStrategy =
+    for<'r, 's> fn(&'r mut Box<(dyn UndoStack<LineBuffer> + 'static)>, &'s LineBuffer);
+
 pub struct Editor {
     line_buffer: LineBuffer,
     cut_buffer: Box<dyn Clipboard>,
     undo_stack: Box<dyn UndoStack<LineBuffer>>,
+    undo_strategy: UndoStrategy,
 }
 
 impl Default for Editor {
@@ -17,6 +40,7 @@ impl Default for Editor {
             line_buffer: LineBuffer::new(),
             cut_buffer: Box::new(get_default_clipboard()),
             undo_stack: Box::new(BasicUndoStack::new()),
+            undo_strategy: undo_strategy_club_similar,
         }
     }
 }
@@ -200,21 +224,6 @@ impl Editor {
         self.line_buffer = val.clone();
     }
 
-    // pub fn set_previous_lines(&mut self, is_after_action: bool) -> Option<()> {
-    //     self.reset_index_undo();
-    //
-    //     if self.edits.len() > 1
-    //         && self.edits.last()?.word_count() == self.line_buffer.word_count()
-    //         && !is_after_action
-    //     {
-    //         self.edits.pop();
-    //     }
-    //     self.edits.push(self.line_buffer.clone());
-    //
-    //     Some(())
-    // }
-    //
-
     pub fn cut_from_start(&mut self) {
         let insertion_offset = self.line_buffer.offset();
         if insertion_offset > 0 {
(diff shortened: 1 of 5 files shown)


src/core_editor/editor.rs
```diff
@@ -1,24 +1,46 @@
...
+) {
+    // TODO: Add more intelligence
+    if undo_stack.current().word_count() == latest_entry.word_count() {
+        undo_stack.undo();
+    }
+    undo_stack.insert(latest_entry.clone())
+}
+
+type UndoStrategy =
+    for<'r, 's> fn(&'r mut Box<(dyn UndoStack<LineBuffer> + 'static)>, &'s LineBuffer);
```
sherubthakur, line 28: The compiler spits out this type. I don't really understand this tbh!

sholderbach:
I have not yet taken a deeper dive into the implementation parts (so I don't grok the ugly looking type signature screaming indirection)

I would list my personal priorities for the undo implementation as follows:

1. How easy is it for someone not intricately familiar with reedline to add or change some editing action (e.g. new completion system, helpful editing shortcut) without subtly breaking the undo machinery?
2. How simple is the surface of the undo machinery if I want to debug or change some behavior with it. (meaning that the implementation it self might be complicated but not magically doing weird things)
3. How easy is it to make changes to what is undoable based on preferences e.g. @mzanrosso implementation tracks the non-editing moves of the cursor as well, most implementations would only move the cursor on text-changing edits.
4. Performance on the insert-side (especially single keypress insert/delete/cursor moves)
5. Overall performance when pressing undo/redo
6. How easy it is to completely change the implementation/ Hiding all the interactions behind abstractions for that

In generally I am all for renaming `Editor::reset_olds` and `Editor::set_previous_lines` to your more descriptive names. 
I am unsure if it is fully helpful if every method is responsible for the undo tracking step (see Prio 1).

sherubthakur:
Let me try to take a stab at answering all the mentioned concerns.

> ugly looking type signature screaming indirection

The type signature is ugly but what part makes you say that it is an indirection? 
- Is it the part where I pick a name for the type signature? 
- or the part where it takes `UndoStack<LineBuffer>` and `LineBuffer` as argument and somehow this could have been a method on the `UndoStack`?
- or something else?

> How easy is it for someone not intricately familiar with reedline to add or change some editing action (e.g. new completion system, helpful editing shortcut) without subtly breaking the undo machinery?

I would like the callers of `Editor` struct not to be bothered by the internals of `Editor` (including how it does undo/redo). The caller can worry about `resetting` the whole editor but given that the undo-redo stack is an internal detail to the `Editor` I don't think the caller needs to worry about it. 
Assuming we can achieve this (which is what I hope we can) 
- it should simplify the job of the caller
- (on the other hand) one will have to be familiar with the undo machinery if they introduce a core edit action in the `Editor` itself. 

Aside on edit actions: I think these are of two types
- Core/Atomic/etc (Don't really know what to call them) -> these are edit actions that we offer as single method at the end of the editor.
- Composed: given the core ones we can string them together to achieve more complex ones. 
I don't think there should be a lot of the core edit actions.

> How simple is the surface of the undo machinery if I want to debug or change some behaviour with it. (meaning that the implementation itself might be complicated but not magically doing weird things)

The approach used in this PR has two components
1. `UndoStack` - This is essentially an abstraction so that we don't have to fiddle around with index management etc in the `Editor` struct and 
2. `UndoStrategy` - This part looks at the old undo actions (the UndoStack) and the current `LineBuffer` and decides whether or not to put this item on the `UndoStack`.

Here I would like to pull the `run_edit_commands` stuff into the `Editor` itself (not sure about it atm). With that, we can have a choice of using the `Vec<EditCommand>` to take a call about undo-redo without exposing the internals of the `Editor`.

 > How easy is it to make changes to what is undoable based on preferences e.g. @mzanrosso implementation tracks the non-editing moves of the cursor as well, most implementations would only move the cursor on text-changing edits.
 
 Given that we have access to current `LineBuffer` and old `LineBuffers`s (via `UndoStack`) when taking this call (the function adhering to the `UndoStrategy` type). It should be easy to achieve.
 
 > Performance on the insert-side (especially single keypress insert/delete/cursor moves)
 
 Given that both the pieces are pluggable (`UndoStack` and `UndoStrategy`) we can tweak performance in them. If that does not suffice (which I doubt) one consideration might be that we are doing this on a single operation if we want to avoid that then we can batch multiple operations by using some sort of a producer-consumer setup (again I doubt performance would become a concern).
 Also note, if we do something like this it will probably go against point 2 (that you raised).
 
> Overall performance when pressing undo/redo

As mentioned above we can have more efficient implementors of `UndoStack` trait or `UndoStrategy` type)

> How easy it is to completely change the implementation/ Hiding all the interactions behind abstractions for that

Refer to response to points 4 and 5.

> In generally I am all for renaming Editor::reset_olds and Editor::set_previous_lines to your more descriptive names.
I am unsure if it is fully helpful if every method is responsible for the undo tracking step (see Prio 1 and 3).

I agree with the point about each and every method being responsible to call the insertion to undo-stack, though I disagree that the internals should be exposed and then `Reedline` struct has to understand the internals of `Editor`. 
One experiment I have wanted to do is try and pull `run_edit_commands` inside the editor, but I am still unsure about.

mzanrosso:
Why do we need right now to change an already working functionality ?

PR #160, opened 2021-10-03
Clarify method names for undo-interface

sholderbach:
Minimal change to make undo machinery of `Editor` more explicit.


jntrnr:
@sholderbach between this and #154 - are they independent or is there an order between them?

sholderbach:
This would be a minimal impact alternative to #154, naming could be reflected in a more ambitious refactor as well (but might cause minimal conflict at first)

jntrnr:
got it. yeah these names are better

PR #164, opened 2021-10-10
Attempt at consolidating offset calculations and streamlining the painting

sholderbach:
### Motivation:
Keeping correct track of coordinates on the terminal screen is paramount to not cause artifacts. Those calculations have to take into account wrapping, new lines and resizing of the terminal. 

### Current situation:
In several locations in `src/engine.rs` such computations are performed and then executed in the `Painter`. Some magical comparisons and offsetting operations are not clearly documented, making it difficult to debug problems and adding new functionality that might require moving around multiple lines.

# Changes for now
* Rename `PromptWidget` to `PromptCoordinates`
* Explicit setters for `PromptCoordinates`
* Set offset implicitly on `Reedline::full_repaint`
* move the terminal size and prompt coordinate information into `Painter`
* Account for some issues with history search
* Make sure that the prompt is painted on the next free line and no spare lines are printed

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -13,9 +13,8 @@ use {
         text_manipulation, DefaultHighlighter, DefaultValidator, EditCommand, Highlighter, Prompt,
         Signal, ValidationResult, Validator,
     },
-    crossterm::{cursor, event, event::Event, terminal, Result},
+    crossterm::{event, event::Event, terminal, Result},
     std::{io, time::Duration},
-    unicode_width::UnicodeWidthStr,
 };
 
 /// Determines if inputs should be used to extend the regular line buffer,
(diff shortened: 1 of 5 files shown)


src/engine.rs
```diff
@@ -381,8 +381,27 @@ impl Reedline {
...
+                            last_edit_commands = Some(ec);
+                        }
+                        (None, other_event) => {
+                            reedline_events.push(other_event);
+                        }
+                        (Some(ref mut last_ecs), ReedlineEvent::Edit(ec)) => {
+                            last_ecs.extend(ec);
+                        }
+                        (ref mut a @ Some(_), other_event) => {
+                            reedline_events.push(ReedlineEvent::Edit(a.take().unwrap()));
```
sholderbach, line 397: It is rather unfortunate to incur this check for being able to avoid a Vec::clone

src/engine.rs
```diff
@@ -381,8 +381,27 @@ impl Reedline {
...
+
+                            reedline_events.push(other_event);
+                        }
+                    }
+                }
+                if let Some(ec) = last_edit_commands {
+                    reedline_events.push(ReedlineEvent::Edit(ec));
```
sholderbach: This is certainly not the most readable piece of code. `itertools::Itertools::coalesce` would be probably a nicer replacement (Writing it out myself might be sign of premature optimization)

PR #167, opened 2021-10-17
add Suggestion struct for use in dialoguer

rsteube:


fdncred:
I think his was great inspiration with dialoguer and we've created something similar but that maybe fits our needs better.

PR #168, opened 2021-11-04
Builder to customize and disable idle animations

josh-59:
Appended an Option<u64> to Reedline struct that allows library users to customize the time waited between repaints.  It is interpreted as milliseconds, and if it is 'None', then no repainting occurs.  Added corresponding builder function.

Ideally, it would be friendlier if we could disable repainting dynamically, especially, when the prompt is not visible, or on 'MouseEventKind::ScrollUp'.  Unfortunately, my terminal (xfce4) would not report mouse events, so I couldn't get that to work.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -11,7 +11,7 @@ use {
         text_manipulation, DefaultHighlighter, DefaultValidator, EditCommand, Highlighter, Prompt,
         Signal, ValidationResult, Validator,
     },
-    crossterm::{cursor, event, terminal, Result},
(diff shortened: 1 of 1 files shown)


jntrnr:
I like it! Nice.

Can you run `cargo fmt` on it? That and the tests (and maybe also `cargo clippy`) are going to get checked by the CI.

jntrnr:
Looks good!

josh-59:
Thanks!  Just wanted to say thanks for your Reedline livestreams- I've
learned a ton from them.

Issue #169, opened 2021-11-05
Cannot type curly brackets in engine-q


aslynatilla:
After building and running `engine-q`, I cannot type `{` or `}`. 
I use an italian keyboard, where opening and closing curly brackets implies pressing: Shift + Alt Gr + Key.

Note that even just typing `[` or `]` seems to present a problem, and it implies pressing: Alt Gr + Key.
In this case, the square brackets are printed, but they can only be cancelled by pressing Backspace twice; it looks like "two characters were pushed" instead of one.

elferherrera:
In theory, this [line](https://github.com/nushell/reedline/blob/main/src/edit_mode/emacs.rs#L31) should take care of it. 

Do you mind running reedline with the flag -k (cargo run -- -k)? 
This flag will show you the events that are being pressed in reedline

aslynatilla:
Building and compiling last commit on reedline's main branch prints:
```
cargo run -- -k
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `target\debug\reedline.exe -k`
Ready to print events (Abort with ESC):
Event::Key(KeyEvent { code: Char('a'), modifiers: NONE })
Event::Key(KeyEvent { code: Char('s'), modifiers: NONE })
Event::Key(KeyEvent { code: Char('['), modifiers: CONTROL | ALT })
Event::Key(KeyEvent { code: Char('{'), modifiers: SHIFT | CONTROL | ALT })
Event::Key(KeyEvent { code: Esc, modifiers: NONE })
```
Everything seems ok, there!

elferherrera:
Oohh... you see that your input requires SHIFT CONTROL ALT

can you try adding that case to the line I showed you? And run again reedline

aslynatilla:
Adding the following case allows me to write `{` or `}`.
```
(m, KeyCode::Char(c)) if m == KeyModifiers::CONTROL | KeyModifiers::ALT | KeyModifiers::SHIFT => {
   ReedlineEvent::Edit(vec![EditCommand::InsertChar(c)])
}
```
However, the "phantom character" problem is still there.
Moreover, since this might be related to this issue... I could see the cursor/caret flickering, going from the start to the end of the line.

sholderbach:
Would you mind opening a PR for the change?

For the phantom character problem could you check if the non brace `AltGr` characters (not `[]{}`)also cause this issue? I would like to exclude an interaction with the brace matching logic used to decide if enter moves to a new line or executes.

[...]

PR #170, opened 2021-11-05
Provide more details in the keyevent debugger

sholderbach:
Inform about the character code and the bit pattern of the modifier to
debug more subtle problems with key events

Prompted by #169


fdncred:
nice! can we land this?

jntrnr:
Nice

Issue #166, opened 2021-10-15
Resizing window causes lag/animation of resize


jntrnr:
similar to #157 

When the window is resized, all the resize events are used, causing the resize to lag and animate from one size to another. 

We may want to check the event stream for other resize events and only use the latest one.

sholderbach:
#179 should improve the performance somewhat but does not deal with the correctness of the ouput after the resize. 

See #42 #157 for other issues arising with resizes

PR #179, opened 2021-11-09
Improvements to handling of incoming events

sholderbach:
To avoid issues with incoming resize events (#166) and improve performance in certain situations (pasting text via terminal facilities) restructure the event input processing:

- Drain all crossterm events currently available
- Prioritize the last resize and discard intermmediates
- Join all EditCommands to be processessed together before handing them to the internal event handler which will trigger a repaint


sholderbach:
Note: previously lived unrelated under #164

jntrnr:
This looks good to me. And trying it out, it's definitely a big improvement over what we had

PR #180, opened 2021-11-10
`fn queue_buffer` probably didn't mean to flush; Simplified engine->insert_char and removed the need to create a String at each insert.

josh-59:
Sorry: Still new to all this Github machinery!  

But, have eliminated the need to create a new String at each character insert.
Since we have separated the manipulation of the underlying buffer from the act of painting it, there's no harm in inserting the character into the buffer, then inspecting that.

Moreover, tried to make the code for 'insert char' more readable, and to simplify the interfaces.


jntrnr:
Thanks for the PR! looks good

Issue #112, opened 2021-07-20
Multiline needs continuation prompt


jntrnr:
For example:
```
> start of line
: continuation of multiline
```

sholderbach:
MVP addressed by #184 and #185

PR #184, opened 2021-11-20
Simple multiline continuation prompt

aslynatilla:
As requested in #112, I tried implementing a simple way of printing the multiline continuation prompt.

Note that this works for multiline blocks only. Opening a block by using a parenthesis, a bracket or a brace, and going to the new line prints the continuation prompt. Text wrapping on the next line does not.
This might actually be considered a feature, since it allows to visualize how the input was actually inserted in the buffer; however, it is visually inconsistent.

As another issue of the implementation, the prompt style is re-created and used in the `render_around_insertion_point` function (see `styled_text.rs`, line 38). This is currently hard-coded since the prompt styling is handled differently when printing the main prompt: it uses `crossterm`'s `Color` instead of `nu_ansi_term`'s one.
Some comments in `full_repaint` (see: `engine.rs`, line 1010) propose a simple way of plugging the style into the aforementioned function, if we were to find a proper way of converting the prompt style's representation across crates.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1007,6 +1007,7 @@ impl Reedline {
         prompt_origin: (u16, u16),
     ) -> Result<(u16, u16)> {
         let prompt_mode = self.prompt_edit_mode();
+        // let prompt_style = Style::new().fg(nu_ansi_term::Color::LightBlue);
         let buffer_to_paint = self.editor.get_buffer();
         let cursor_position_in_buffer = self.editor.offset();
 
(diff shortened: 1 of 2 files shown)


jntrnr:
Nice! very cool

fdncred:
works well. thanks @aslynatilla ! I just noticed that it doesn't use the predefined line continuation character. maybe you can circle back and put this in place?
https://github.com/nushell/reedline/blob/75d9b4b52d0ea60334ef96d49cd00002acc5720c/src/prompt.rs#L14

aslynatilla:
I can try and fix it as soon as it is possible for me...!

PR #185, opened 2021-11-24
Forwarding Prompt's multiline indicator

aslynatilla:
As requested [here](https://github.com/nushell/reedline/pull/184#issuecomment-977800119) and as a direct follow-up to #184, I tried to "forward" a Prompt parameter to the rendering functions, so that the `render_prompt_multiline_indicator` was available.

I am not sure about how I used `Cow<str>, or that,  from a design standpoint, passing in the multiline prompt is the best idea. I am available to implement any variation, in order to make it better, of course!

### DIFF
--- a/src/styled_text.rs
+++ b/src/styled_text.rs
@@ -1,3 +1,5 @@
+use std::borrow::{Borrow, Cow};
+
 use nu_ansi_term::{Color, Style};
 
 /// A representation of a buffer with styling, used for doing syntax highlighting
@@ -31,6 +33,7 @@ impl StyledText {
         &self,
         insertion_point: usize,
         // prompt_style: &Style,
+        multiline_prompt: Cow<str>,
     ) -> (String, String) {
         let mut current_idx = 0;
         let mut left_string = String::new();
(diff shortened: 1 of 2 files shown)


jntrnr:
Ah nice. The code looks good. Can you also take a look at the merge conflicts?

aslynatilla:
Within Saturday night, I will be able for sure...! Is that ok?

jntrnr:
Yup, no worries

aslynatilla:
Conflicts should be solved, now!

src/styled_text.rs
```diff
@@ -31,24 +33,41 @@ impl StyledText {
         &self,
         insertion_point: usize,
         // prompt_style: &Style,
```
sholderbach, line 35: Should we yank that, or do we want to transfer that through?

src/styled_text.rs
```diff
@@ -57,11 +76,16 @@ impl StyledText {
...
+    renderable: &(Style, String),
+    prompt_style: &Style,
+    multiline_prompt: &str,
+) -> String {
     let mut rendered = String::new();
+    let formatted_multiline_prompt = format!("\n{}", multiline_prompt);
     for (line_number, line) in renderable.1.split('\n').enumerate() {
         if line_number != 0 {
-            rendered.push_str(&prompt_style.paint("\n: ").to_string());
+            rendered.push_str(&prompt_style.paint(&formatted_multiline_prompt).to_string());
```
sholderbach, line 88: Allocation `.to_string()` is unnecessary as you immediately consume them. Can be replaced by either `.deref()`, `.as_ref()`, or prepending with `&*` (which I personally find less readable).

src/styled_text.rs
```diff
@@ -57,11 +76,16 @@ impl StyledText {
...
         }
         rendered.push_str(&renderable.0.paint(line).to_string());
```
sholderbach, line 90: see above

sholderbach (commented):
Functionality LGTM :+1: 

Small comments regarding the code

src/styled_text.rs
aslynatilla: I think that is a design decision that should be done "collectively"; does it make sense that the prompt style is forwarded like that? Do we want to take it out? Or should `StyledText` have access to it through other means?

Note that the Style is currently created inside the function because there was no way to obtain a compatible `Style` from the call sites, and the conversion between `nu_ansi_term::Style` and `crossterm::Style` wasn't easy to perform.

src/styled_text.rs
aslynatilla: I can fix those asap!

src/styled_text.rs
sholderbach: Future separation of concern alternative would be to move the integration of the prompt strings out of the `render_around_insertion_point()` function and do that in the painter as the prompt is passed into that one already. Writing a translation between the `nu-ansi-term` and `crossterm` styles was on my list of coding tasks to mindlessly do with rust analyzer :stuck_out_tongue_winking_eye:

aslynatilla: If you're okay with that, then, I would remove the "comment-parameter" and leave the conversion task to you! ;P

sholderbach: The conversion turns out to be not completely 1 to 1 as nu-ansi-term seems to use the full xterm256 8-bit color palette to define its main colors while crossterm seems to rely on the 4-bit basic set (that might be more cross plattform) for the named colors. But we might have to check in with the nu old guard what their experience was so far.

src/styled_text.rs
sholderbach: My bad, the deref is implemented differently and this breaks formatting

PR #188, opened 2021-11-30
Fix busy polling when animation is disabled

sholderbach:
Addresses https://github.com/nushell/engine-q/issues/386

Breaks the API for variable refresh times, standardizes on one second
and simply optionally disables the automatic repaint.


sholderbach:
#189 would be an alternate fix that would not break the API and would not necessitate small changes in engine-q

fdncred:
Looks good to me. We just need to get it into e-q and i'll have to change how the animate works there.

fdncred (approved):
Looks good.

PR #189, opened 2021-11-30
Minimal fix for poll busy loop

sholderbach:
Alternative to #188 which would not break API

Addresses nushell/engine-q#386

When disabling the clock animation by not wanting to repaint give poll a
1 sec timeout instead of immediately polling again. Option with time is
still used, using small millisecond timeouts can still turn it into a
busy loop.
Would not necessistate a change on the engine-q side and allow for
playing around with this timeout for debugging but might not be the most
useful feature for general use

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -356,7 +356,7 @@ impl Reedline {
         let mut reedline_events: Vec<ReedlineEvent> = vec![];
 
         loop {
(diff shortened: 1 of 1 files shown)


fdncred (commented):
Either way is fine with me. Let's run with the other one and if we don't like it we can try this one.

sholderbach:
The behavior is virtually the same, with this one leaving more options to the user. Will merge #188 and make the fix for engine-q

Issue #190, opened 2021-12-01
Incomplete coverage of operations by Undo/Redo functionality


sholderbach:
At the moment not all available editing operations that affect the content of the line buffer are covered by the Undo/Redo logic.

e.g. using `Alt+c` to capitalize a character

The check for operations that require saving of the state is currently not exhaustive

https://github.com/nushell/reedline/blob/82bc3acb64e4d3a3a13b57d19dc014652e7daee2/src/engine.rs#L846-L865

This should be replaced by a function using a fully specified `match` to catch any future additions of `EditCommand`s.

PR #191, opened 2021-12-01
Improve undo of EditCommands

sholderbach:
Address #190

Formalize the undo role of different edit commands via exhaustive
matching

Open questions:

- Undo behavior of pure moves #193 
    * Do we want to make them undoable
    * Do we have to register them for accurate undos of true edits
- Manual backspace and delete #194 
    * Do we want to coalesce them if a whole word is manually deleted?

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -3,7 +3,7 @@ use {
         completion::{CircularCompletionHandler, CompletionActionHandler},
         core_editor::Editor,
         edit_mode::{EditMode, Emacs},
-        enums::ReedlineEvent,
+        enums::{ReedlineEvent, UndoBehavior},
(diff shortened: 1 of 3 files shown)


sholderbach:
Excluding the moves as in 85dab62 has the weird side effect that after an undo the cursor jumps back to where it was after the last text changing operation
undo_test
(Deletion of `is` with `Ctrl+w` and then undo via `Ctrl+z`)

sholderbach:
Conversely when moves are recorded, undos will follow the cursor (Current behavior on main)

undo_moving

sholderbach:
Would decide to merge this first (with undoable moves) and let somebody address the moves later with #193 

CI checks failed due to more sensitive rustc/clippy warnings with latest toolchain

PR #197, opened 2021-12-09
add the ability to optionally turn off ansi coloring

fdncred:
`with_ansi_colors(false)`
image

`with_ansi_colors(true)`
image


sholderbach:
Looks good to me, users without ANSI terminal support probably want to turn of the hints anyways as they become hard to distinguish from the actual input

Issue #183, opened 2021-11-19
Problems with input requiring more lines than available in the terminal window screen


fdncred:
using engine-q, i was testing out pasting json and was presented with this panic.

this is the json i pasted.
```
{
  "name": "inc",
  "usage": "Increment a value or version. Optionally use the column of a table.",
  "extra_usage": "",
  "required_positional": [],
  "optional_positional": [],
  "rest_positional": {
    "name": "",
    "desc": "",
    "shape": "Any",
    "var_id": null
  },
  "named": [
    {
      "long": "help",
      "short": "h",
      "arg": null,
      "required": false,
      "desc": "Display this help message",
      "var_id": null
    },
    {
      "long": "major",
      "short": "M",
      "arg": null,
      "required": false,
      "desc": "increment the major version (eg 1.2.1 -> 2.0.0)",
      "var_id": null
    },
    {
      "long": "minor",
      "short": "m",
      "arg": null,
      "required": false,
      "desc": "increment the minor version (eg 1.2.1 -> 1.3.0)",
      "var_id": null
    },
    {
      "long": "patch",
      "short": "p",
      "arg": null,
      "required": false,
      "desc": "increment the patch version (eg 1.2.1 -> 1.2.2)",
      "var_id": null
    }
  ],
  "is_filter": false,
  "creates_scope": false,
  "category": "Default"
}
```

this is the backtrace

```
thread 'main' panicked at 'attempt to subtract with overflow', /Users/fdncred/.cargo/git/checkouts/reedline-e42026a78d91c510/c11aef2/src/engine.rs:928:60
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::panicking::panic
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:50:5
   3: reedline::engine::Reedline::wrap
             at /Users/fdncred/.cargo/git/checkouts/reedline-e42026a78d91c510/c11aef2/src/engine.rs:928:60
   4: reedline::engine::Reedline::run_edit_commands
             at /Users/fdncred/.cargo/git/checkouts/reedline-e42026a78d91c510/c11aef2/src/engine.rs:822:25
[...]

PR #201, opened 2021-12-19
Disable raw mode with `Drop` for panics

sholderbach:
Panicking inside reedline (or bubbling up an Err inside reedline's
critical sections that gets unwrapped/expected by the hosting
application) leaves the terminal in a bad state. Workaround that should
also work without cooperation by the consumer would be to include a call
to `disable_raw_mode` in the `Drop` trait. With standard stack-unwinding
panics will call the drop.

Calling `crossterm::terminal::disable_raw_mode` twice seems to be fine
if the program started in non-raw mode (under linux)

@jntrnr or @fdncred could you check if running the demo does not cause a problem due to disabling the raw mode twice

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -108,6 +108,14 @@ pub struct Reedline {
     use_ansi_coloring: bool,
 }
 
+impl Drop for Reedline {
+    fn drop(&mut self) {
+        // Ensures that the terminal is in a good state if we panic semigracefully
+        // Calling `disable_raw_mode()` twice is fine with Linux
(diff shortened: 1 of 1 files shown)


sholderbach:
Looking at the crossterm Windows source, calling it twice should be fine. If you want to cause a panic to check for proper restore a good way would currently be to paste a json with more lines than available (#183)

fdncred (approved):
I tested it out and it looks good to me. I couldn't make it break.

jntrnr:
Looks great!

PR #211, opened 2021-12-25
vi mode commands

elferherrera:
Adds new functionality to vi mode using vi parser


src/core_editor/line_buffer.rs
```diff
@@ -454,6 +454,92 @@ impl LineBuffer {
     pub fn is_cursor_at_last_line(&self) -> bool {
         !self.get_buffer()[self.offset()..].contains('\n')
     }
+
+    /// Finds index for the first occurrence of a char to the right of offset
+    pub fn find_char_right(&self, c: &char) -> Option<usize> {
+        if self.offset() + 1 > self.lines.len() {
```
sholderbach, line 460: I think this misses the case of being on the last (potentially multibyte character that can not move any further). But the search might fail already.

src/core_editor/line_buffer.rs
```diff
@@ -454,6 +454,92 @@ impl LineBuffer {
...
+            return None;
+        }
+
+        let search_str = &self.lines[self.offset() + 1..self.lines.len()];
+        search_str.find(*c).map(|index| index + self.offset() + 1)
```
sholderbach, line 465: The `+1` has to be replaced with the `char::len_utf8` for the character starting at offset. `LineBuffer::grapheme_right_index()` finds the next index of a grapheme (potentially wider than the codepoint)

```suggestion

        let search_str = &self.lines[self.grapheme_right_index()..];
        search_str.find(*c).map(|index| index + self.grapheme_right_index())
```

src/core_editor/line_buffer.rs
```diff
@@ -454,6 +454,92 @@ impl LineBuffer {
...
+            self.insertion_point.offset = index
+        }
+
+        self.insertion_point.offset
+    }
+
+    /// Moves the insertion point before the next char to the right
+    pub fn move_right_before(&mut self, c: &char) -> usize {
+        if let Some(index) = self.find_char_right(c) {
+            self.insertion_point.offset = index - 1
```
sholderbach, line 490: Bit tricky. 
We have to also make sure that we end up with the cursor on valid graphemes (e.g. Emojis multiple code points for a single column on the display)
```suggestion
            self.insertion_point.offset = index;
            self.insertion_point.offset = self.grapheme_left_index();
```

sholderbach (changes requested):
Unicode issues will cause panics for non-ascii characters

elferherrera:
@sholderbach I added your suggestions. Thanks, that helped a lot with characters larger than 1 char. 

There is one issue that I haven't figured. When you are in a multi line edit, something like this
```
let a = {
sentence 1
sentence 2
sentence 3
}
```
if I delete the word sentence with `dw` the word doesnt disappear, but it isnt there any more
I checked the emacs bindings and it as well uses the same edit command (DeleteWord)
Is there something I am missing?

sholderbach (approved):

src/core_editor/line_buffer.rs
sholderbach: The second `self.grapheme_right_index()` was missing and causing panics on `f<non-ascii>` moves

sholderbach (approved):

PR #212, opened 2021-12-26
Improve history internals

sholderbach:
- Move history newline escape to the file IO

This avoids that the newline escape character can be found via search and thus mess upt the search results, also makes the API internally simpler

- Remove unnecessary optional history allocation

Very minor potential improvement


sholderbach:
Checked engine-q for potential interaction with API assumptions, no direct link, found thus going ahead with the merge.
`history` command is made aware of the escaping with nushell/engine-q#592

PR #213, opened 2021-12-27
vi insert mode default

elferherrera:
Making vi insert mode default when starting terminal


sholderbach:
Yeah that's more convenient and the behavior I remember of zle's vi mode

PR #215, opened 2021-12-27
Disable hints during history traversal

sholderbach:
Additional autosuggestion (history) hints are confusing while
navigating through the history.

Example history:
```
hello test
hello
something else
hello world
whatever
```

- Type `h`
- Press up arrow
- Prefix based history search will return `hello world`
- Press up arrow again
- Previously `hello <hint>world</hint>` now just `hello`

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -515,7 +515,7 @@ impl Reedline {
                 let line_buffer = self.editor.line_buffer();
 
                 let current_hint = self.hinter.current_hint();
-                if !current_hint.is_empty() {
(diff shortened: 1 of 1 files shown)


fdncred (approved):
it seems to work the way you describe but it's hard to tell with such a limited history. I'd like to commit it and pull the update into e-q to see how this works.

sholderbach:
Just let me know if you revert it, so I can repeat the deduplication

fdncred:
ok. I can't mess with it for a little while. I need to install one of those video doorbell things I got for Christmas.

fdncred:
I'm really liking this so far.

PR #217, opened 2021-12-28
prompt moves based on hint

elferherrera:
The prompt moves and gives space to hint when it is located at the bottom. 

image
the prompt moves upwards based on the suggestion
image

Note: This also removes the issue when the suggestion was being redrawn every second

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -811,9 +811,9 @@ impl Reedline {
                 EditCommand::InsertChar(c) => {
                     self.editor.insert_char(*c);
 
-                    if self.painter.require_wrapping(&self.editor) {
-                        self.handle_wrap(prompt)?;
-                    }
+                    // if self.painter.require_wrapping(&self.editor) {
+                    //     self.handle_wrap(prompt)?;
+                    // }
 
                     self.repaint(prompt)?;
                 }
@@ -822,9 +822,9 @@ impl Reedline {
                         self.editor.insert_char(c);
                     }
 
-                    if self.painter.require_wrapping(&self.editor) {
-                        self.handle_wrap(prompt)?;
-                    }
+                    // if self.painter.require_wrapping(&self.editor) {
+                    //     self.handle_wrap(prompt)?;
+                    // }
 
                     self.repaint(prompt)?;
                 }
@@ -905,7 +905,7 @@ impl Reedline {
         if self.input_mode == InputMode::HistorySearch {
             self.history_search_paint(prompt)
         } else {
-            // self.buffer_paint(prompt)?;
+            // self.buffer_paint(prompt)
             self.full_repaint(prompt)
         }
     }
@@ -936,7 +936,6 @@ impl Reedline {
                 Some(string) => {
                     self.painter
                         .queue_history_search_result(&string, string.len())?;
-                    self.painter.flush()?;
                 }
 
                 None => {
@@ -986,7 +985,7 @@ impl Reedline {
     // fn buffer_paint(&mut self, prompt: &dyn Prompt) -> Result<()> {
     //     let lines = self.prepare_buffer_content(prompt);
 
-    //     if lines.required_lines() > self.painter.remaining_lines() {
+    //     if lines.required_lines()? > self.painter.remaining_lines() {
     //         let prompt_mode = self.prompt_edit_mode();
     //         self.painter
     //             .repaint_everything(prompt, prompt_mode, lines, self.use_ansi_coloring)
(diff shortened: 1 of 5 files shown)


src/painter.rs
```diff
@@ -35,6 +35,54 @@ impl PromptCoordinates {
...
+            #[allow(clippy::single_char_pattern)]
+            after_cursor.split("\n").collect::<Vec<&str>>()
+        };
+
+        let hint_lines = if cfg!(windows) {
+            hint.split("\r\n").collect::<Vec<&str>>()
+        } else {
+            #[allow(clippy::single_char_pattern)]
+            hint.split("\n").collect::<Vec<&str>>()
+        };
```
sholderbach, line 72: For better readability how about making the switch in the definition of a single `const NEW_LINE_SEPARATOR: &str` ?

src/engine.rs
```diff
@@ -978,7 +980,25 @@ impl Reedline {
...
+                hint,
+                self.use_ansi_coloring,
+            )?;
+        } else {
+            self.painter.queue_buffer(
+                &lines.before_cursor_lines,
+                &lines.after_cursor_lines,
+                &hint,
+            )?;
+        }
```
sholderbach, line 1000: This logic is probably also required in the full_repaint and history paint cases. Can we lift that into `Painter::queue_buffer`?

src/engine.rs
```diff
@@ -978,7 +980,25 @@ impl Reedline {
     fn buffer_paint(&mut self, prompt: &dyn Prompt) -> Result<()> {
         let (highlighted_line, hint) = self.prepare_buffer_content(prompt);
 
-        self.painter.queue_buffer(highlighted_line, hint)?;
+        let lines = PromptLines::from_strings(&highlighted_line.0, &highlighted_line.1, &hint);
```
sholderbach, line 983: How about letting `prepare_buffer_content()` directly returning the prompt lines?

src/painter.rs
```diff
@@ -35,6 +35,54 @@ impl PromptCoordinates {
...
+
+        Self {
+            before_cursor_lines,
+            after_cursor_lines,
+            hint_lines,
+        }
+    }
+
+    pub fn required_lines(&self) -> u16 {
+        (self.before_cursor_lines.len() + self.hint_lines.len()) as u16
```
sholderbach, line 82: Is this exhaustive if there are multiple lines after the cursor?

src/painter.rs
sholderbach: Alternatively... is `str::lines` doing the right thing already?

src/painter.rs
```diff
@@ -102,47 +155,38 @@ impl Painter {
...
-            #[allow(clippy::single_char_pattern)]
-            after_cursor.split("\n")
-        };
-
-        let mut commands = self.stdout.queue(MoveTo(
+    pub fn queue_buffer(
+        &mut self,
+        before_cursor_lines: &[&str],
+        after_cursor_lines: &[&str],
+        hint: &str,
```
sholderbach, line 162: Can this simply be `PromptLines`?

src/painter.rs
```diff
@@ -194,6 +238,21 @@ impl Painter {
         hint: String,
```
sholderbach, line 238: Also simplification via `PromptLines`

src/painter.rs
```diff
@@ -35,6 +35,54 @@ impl PromptCoordinates {
     }
 }
 
+pub struct PromptLines<'prompt> {
```
sholderbach, line 38: Bikeshed: `BufferView` as it does not care about the `Prompt` directly and `View` may reflect the reference nature. Same for the lifetime name

sholderbach (commented):
Awesome that you are tackling that! I like the simplification. Will try to give it a thorough testing

jntrnr:
Seeming some places where we need to output /r/n to keep things aligned:

image

elferherrera:
@jntrnr can you check with the new commit?

jntrnr:
@elferherrera same issue -

image

src/painter.rs
```diff
@@ -15,6 +17,8 @@ use {
     unicode_width::UnicodeWidthStr,
 };
 
+const END_LINE: &str = if cfg!(windows) { "\r\n" } else { "n" };
```
sholderbach, line 20: Oopsie

elferherrera:
@jntrnr @sholderbach @fdncred do you mind checking the new commit?

jntrnr:
@elferherrera - same issue

image

fdncred:
ya, same for me too. 
Screen Shot 2021-12-29 at 4 16 14 PM

elferherrera:
@fdncred @sholderbach do you mind checking again this commit?

sholderbach:
Looking good to me. Even wrapping lines seem to be good at the bottom of the screen. The only thing i noticed that if the last line is reached with actual content, it jumps up a line to leave a blank line after a full repaint.

fdncred:
I'd land this one and continue to work on other issues like 'very long buffer' and 'jumps up a line'.

elferherrera:
@fdncred I think that is a good idea. I dont want to make the PR larger

jntrnr:
Tried out the latest - it doesn't seem to let you work with the last line of the screen:
image

I also noticed that if you have a multiline history hint at the bottom, each character scrolls:
image

jntrnr:
Here's an animation to see these a little bit better:
Peek 2021-12-31 08-00

fdncred:
it's interesting that I don't get this behavior on Mac.

jntrnr:
@elferherrera - thanks for banging this into shape!

PR #219, opened 2021-12-29
Make the editing operations aware of multiline

sholderbach:
- Moves to start and beginning of the line
- Make start and end of buffer available for move via `Ctrl-Home` and
`Ctrl-End`
- Support vi style cutting of whole lines in the cut buffer.
- add vi style `p` after paste in addition to before cursor paste with
`P` or in emacs mode
- Enables vi style `dd` and paste
- Make sure the line to end cut and clear operations work only on the
line


jntrnr:
Looks good!

PR #222, opened 2021-12-29
Turn off raw when drawing

jntrnr:
This enables raw mode only when we're taking in input from the user, otherwise it turns it off so drawing doesn't have to worry about it


fdncred:
seems to straighten out my multi-line history.
Screen Shot 2021-12-29 at 4 27 39 PM

but wow, on the last line of the terminal, fuggedabowdit.

fdncred:
and until you hit enter
Screen Shot 2021-12-29 at 4 30 34 PM

PR #223, opened 2021-12-29
fix: make traits `Send`

### LINKED ISSUE #220: traits are not thread safe
Would it be possible to have the traits to be thread safe? This is a blocker for Deno to switch away from rustyline to reedline, as we spawn a new tokio thread for each read_line call

crowlKats:
Closes #220

### DIFF
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -31,13 +31,13 @@ impl Span {
 /// The handler for when the user begins a completion action, often using the tab key
 /// This handler will then present the options to the user, allowing them to navigate the options
 /// and pick the completion they want
-pub trait CompletionActionHandler {
+pub trait CompletionActionHandler: Send + Sync {
     /// Handle the completion action from the given line buffer
     fn handle(&mut self, line: &mut LineBuffer);
 }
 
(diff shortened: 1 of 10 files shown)


sholderbach:
Thanks! Looks good to me. CI just failed on `cargo fmt`. 

@jntrnr can we think of any reason why we would want something not necessarily `Sync`?

bartlomieju:
Maybe a test case in `src/engine.rs` could be useful? Something like:
```rust
#[test]
fn thread_safe() {
  fn f<S: Send + Sync>(_: S) {}
  f(Reedline::create().unwrap());
}
```

fdncred:
can someone add that test case please?

jntrnr:
@sholderbach - that's a good question

@crowlKats - do you need both Send + Sync or just Send? I'm not as familiar with tokio and its requirements.

crowlKats:
Test added, and it seems just `Send` is required, adjusted the PR

sholderbach:
Thanks!

jntrnr:
Looks good. Thanks @crowlKats!

bartlomieju:
Thanks folks, much appreciated! We'll get back to you after preliminary tests.

PR #224, opened 2021-12-30
Test and fix line buffer utils

sholderbach:
- current line detection was off on the first position of a following
line
- find_current_line_end was not correctly aware of CRLF platforms
- Increase of coverage for some old functions


jntrnr:
nice

PR #226, opened 2021-12-31
adjust position to insert in middle

elferherrera:
There were repetition when inserting text in the middle of a multiline buffer. This should correct it


jntrnr:
nice!

PR #234, opened 2022-01-01
Improve and highlight history search

sholderbach:
Fixes `history_search_paint`:
- Fix unnecessary clear of single line
- Ensure LF to CRLF in search results

Fixes `Painter`:
- Make sure that Color is cleared after prompt (Currently depends on
Highlighter to provide a new color)
- Remove unnecessary fn

Introduce simple highlighting of the search term in the search result
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ let mut line_editor = Reedline::create()?
 ```rust,no_run
 // Create a reedline object with highlighter support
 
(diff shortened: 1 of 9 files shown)


fdncred:
that looks cool!!

sholderbach:
Replaced the example highlighter with a proper implementation that will highlight all matches

PR #240, opened 2022-01-05
supports double prompt

elferherrera:


fdncred:
this is on my Mac. looks like the measuring maybe be off a few chars.
rprompt

jntrnr:
Linux isn't quite there yet:

image

fdncred:
here's the latest commit on Mac.
rprompt_f

jntrnr:
did some estimation instead of reading the cursor position directly: https://github.com/nushell/reedline/tree/double-prompt

jntrnr:
Merged as part of #241

PR #243, opened 2022-01-09
Context menu

elferherrera:
Introduces context menu for reedline. 

To activate the menu use ALT+1. This is temporal while we decide how to call the menu
Animation1


fdncred:
at some point we'll need to come back to this and disable the color if use_ansi_coloring is $false.

elferherrera:
@fdncred sure. I just want to merge it so I can use it in e-q and test the ideas I have

PR #252, opened 2022-01-14
fix windows type-o

fdncred:


sholderbach:
Thanks! I missed that one

PR #257, opened 2022-01-17
Parsing keybindings

elferherrera:
exposes keybindings to be used in engine q


sholderbach:
Are you sure everything went fine with the merges? The final diff of the PR looks somewhat small? or is it just the final commit?

elferherrera:
It's just the last commit. I don't know why this new branch added those commits. I need to fix my git

PR #262, opened 2022-01-19
Remove superfluous insertion of carriage return

sholderbach:
In the hinter logic `LF` was always replaced by `CRLF` thus inserting unnecessary `CR`'s into the linebuffer that accumulate if a multiline entry is recalled via hint autosuggestion from history multiple times.

image
- Hexdump of history file shows contamination with `CR` (`0x0D`)

Carriage returns, that are not introduced at the end of the line as part of the painter or the OS convention, break the painting of multilines, as the cursor position might jump to the start of the multiline continuation prompt and start painting from there!

### DIFF
--- a/src/hinter.rs
+++ b/src/hinter.rs
@@ -54,8 +54,6 @@ impl Hinter for DefaultHinter {
                 let span = completions[0].0;
                 hint.replace_range(0..(span.end - span.start), "");
 
-                let hint = hint.replace("\n", "\r\n");
-
                 self.current_hint = hint.clone();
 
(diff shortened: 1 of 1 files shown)


fdncred:
my guess is that crlf was here because hints were broken on windows, which uses crlf, for a while and they looked crazy until the crlf vs lf issue was resolved. I'm hoping that fix was somewhere else and not here.

sholderbach:
The raw mode requires CR independent of platform and thus the painting has to ensure them but that was probably a relict from the non-centralized painting coming together with the woes of making the history/hinter multiline capable

PR #263, opened 2022-01-20
Remove `Reedline::print_line` from the API

sholderbach:
As this function can only be called outside running
`Reedline::read_line` it is not relevant for raw mode reasons and can be
replaced by standard `println!()`


elferherrera:
Looks good

PR #264, opened 2022-01-20
Show content after cursor when completing

sholderbach:
Currently the content after the cursor was completely omitted when
entering the completion menu

Addresses it for the simple case of everything fitting on the screen

Question what should take priority if the entry and menu together exceed
the height of the screen remains open!


sholderbach:
Merged with #265

PR #266, opened 2022-01-20
multiple options for keybindings

elferherrera:
decouples event that were hard coded and gives the opportunity to define multiple events to the same keybinding


sholderbach:
Continued with alternative to `8cc1ff9` in #267

PR #269, opened 2022-01-21
Support "token by token" hint completion

sholderbach:
Hints can now be completed partially

Keybindings: `Alt-right` `Alt-f` `Ctrl-right`

Uses whitespace as token/word boundary for now

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -96,7 +96,7 @@ mod test {
         keybindings.add_binding(
             KeyModifiers::CONTROL,
             KeyCode::Char('l'),
-            ReedlineEvent::Complete,
+            ReedlineEvent::HistoryHintComplete,
         );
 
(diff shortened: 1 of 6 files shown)


elferherrera:
Do you mind adding the same keybinding for vi insert mode?

sholderbach:
Vi insert is no problem, for Vi normal we probably have to be a bit more elegant

elferherrera:
Im wondering how detailed should vi mode be

sholderbach:
Yeah our current vi behavior is not compliant when it comes to word boundaries, but I consider that a problem for another day. Either procrastination or complaints by vi purists will determine the timeline on that :D

jntrnr:
Let's go ahead and merge this and play with it. We might want to compare against fish as we do, just see if folks coming from other shells will have something familiar

fdncred:
just lovely. thanks stefan!

PR #270, opened 2022-01-21
Remove outdated CompletionActionHandler(s)

sholderbach:
Thanks to @elferherrera, we now have a nice, interactive completion option with the completion list, that supports arrow navigation and tab completion.

Thus, we have to consider consolidating the completion behavior of reedline:

**Updated** pinging @fdncred @jntrnr @elferherrera 

The `ListCompletionActionHandler` implementation is incompatible with the line estimation of the painter and superseded by the selection supporting menu and can thus be completely replaced/removed.
The public trait `CompletionActionHandler` is removed by this PR as correct implementation of painting and line buffer integration requires some care and we should provide the UI options.
The CircularCompletionHandler is now available statically and the`Reedline` can be built `with_completer()` to pass the completion logic.

The **open question** is how to best configure different completion UI  options and configure the event handling in the best way:

The event handling is currently separated into `ActionHandler` for the basic bash style completion and `Menu_` events to handle the interactive completion menu.

- Option A) Just use different key to event mappings to change the behavior. Downside: requires utmost care when changing keybindings
- Option B) Have a completion UI config struct/builder pattern to define the preferences and perform the event dispatch internally based on the available components


elferherrera:
When working on the menu I had no idea what to do with those. I guess they could be used by an implementation that doesn't want a menu and only do completion. I would remove them and only leave a simple implementation of the trait for people that may want to use Reedline

fdncred:
I don't think we want only one way to do things. We want to allow people to have options. If they don't want to use @elferherrera's completion thing, what options do they have right now? As long as they have some other option, I'll be fine with a little house cleaning. Maybe we should have some rudimentary examples that show our different implementations.

sholderbach:
> I don't think we want only one way to do things. We want to allow people to have options. 
Fully agree

Currently we have an implementation of bash style completions that just change in place on tab. (the `CircularCompletionHandler`)

I assume as we can not really infer a reasonable ordering of multiple completion sources on a case by case basis, it is probably reasonable for each user be it nushell or deno to implement one `Completer` that provides suggestions (and maybe some metadata on how to preferrably display the case) 

I think with the technical subtleties around painter and how to keep the line buffer consistent, I would prefer to have the UI elements all as part of reedline. 

The question becomes how do we want to expose the configurabiltiy? Just with the keybinding of `Tab`? 
Have a builder pattern as follows?
```
Reedline::create()
.with_completer(the_single_completer)
.with_completion_style(CompletionStyle::Interactive)
```

elferherrera (approved):

PR #271, opened 2022-01-21
Avoid panic if completion exceeds screen width

sholderbach:
Painter panicked, if the menu entry shown is wider than the screen.

This commit just avoids the panic but causes painter miscalculations
leading to wrapping at the bottom of the screen and oversized entries

TODO fix this on the painter side as well

### DIFF
--- a/src/context_menu.rs
+++ b/src/context_menu.rs
@@ -2,6 +2,8 @@ use nu_ansi_term::{ansi::RESET, Color, Style};
 
 use crate::{Completer, DefaultCompleter, LineBuffer, Span};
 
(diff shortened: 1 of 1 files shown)


elferherrera:
How can I reproduce this error?

sholderbach:
Make the window narrow, with the reedline demo type `h<TAB>` as there is a long completion.

elferherrera:
Funny. That should trigger a large buffer state and deal with it differently

sholderbach:
Closing for now as some of the fixes seem to be reproduced in #273

PR #276, opened 2022-01-27
Generic menus

elferherrera:
Introduces a Menu trait that can be used to create menus for reedline

### DIFF
--- a/src/edit_mode/keybindings.rs
+++ b/src/edit_mode/keybindings.rs
@@ -99,20 +99,12 @@ pub fn add_common_keybindings(kb: &mut Keybindings) {
     kb.add_binding(
         KM::NONE,
         KC::Up,
-        ReedlineEvent::UntilFound(vec![
-            ReedlineEvent::HistoryMenuPrevious,
-            ReedlineEvent::MenuUp,
-            ReedlineEvent::Up,
-        ]),
+        ReedlineEvent::UntilFound(vec![ReedlineEvent::MenuUp, ReedlineEvent::Up]),
     );
     kb.add_binding(
         KM::NONE,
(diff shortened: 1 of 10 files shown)


sholderbach:
That change is pretty gnarly to get compatible with #270 as the trait makes explicit assumptions about info available at `update_values` time. 

Second aspect I am not too fond of is that each implementer has to provide the row/column calculations and implement the movements themselves, with a lot of opportunities to mess stuff up

sholderbach:
Alternative idea:

``` rust
struct Menu {
  options: Box<dyn MenuOptions>
// Implements the details of layouting and making sure that selection can be managed with keybindings
}

trait MenuOptions {
  fn get_entries(&self, line_buffer: &LineBuffer) -> Vec<(String, Style)>; // Implementer has to take care of gathering the necessary info separately
  fn execute_entry(&self, index: usize, line_buffer: &LineBuffer) -> LineBuffer; // Could in theory also be some other kind of operation if the menu should not just affect the `LineBuffer`. e.g. menu for changing settings
  fn clear(&mut self);
}

struct HistoryMenu; // can update itself based on the history (call to that could be special cased in `Reedline`)
struct CompletionMenu; // Can update itself with the completer being part of `Reedline`

impl MenuOptions for HistoryMenu {}
impl MenuOptions for CompletionMenu {}

trait UpdateOptions {
  fn update_values(&mut self, line_buffer: &LineBuffer, ...) // For external menus provide the information that they reasonably might need
}

type CustomMenu = MenuOptions + UpdateOptions;

fn Reedline::add_menu(&mut self, menu_content: CustomMenu, layout_preferences: MenuStyle)

```

PR #280, opened 2022-01-27
chronological order search

elferherrera:
Change how the history filters data. The filtered values now are sorter chronologically 

Also the keybinding for the history menu is changed to match engine q


sholderbach:
Looks good to me!

PR #281, opened 2022-01-28
Coerce to CRLF lazily

sholderbach:
Will only introduce a CR if LF is not part of CRLF.

Only allocates if necessary


sholderbach:
Will go ahead and merge it to see if this also addresses nushell/engine-q#575

PR #283, opened 2022-01-28
Make `strip_ansi` lazy allocating

sholderbach:
If supplied string does not contain control sequences avoids allocation.
Performs simple additional check beforehand.

True performance impact on painting is hard to realistically benchmark

### DIFF
--- a/src/styled_text.rs
+++ b/src/styled_text.rs
@@ -97,17 +97,18 @@ impl StyledText {
 ///
 /// If parsing fails silently returns the input string
 pub(crate) fn strip_ansi(string: &str) -> Cow<str> {
+    // Check if any ascii control character except LF(0x0A = 10) is present,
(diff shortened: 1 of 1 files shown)


sholderbach:
In the nushell table and grid layout machinery, this change noticably improved performance as the most likely codepath was input without ANSI control sequences or ASCII control chars. (https://github.com/nushell/nushell/pull/4378, https://github.com/nushell/nushell/pull/4380)
In reedline this is probably more of a mixed bag as the syntax highlighted text gets processed frequently and syntax highlighting is currently in-band ANSI at the painter stage. The impact of the additional check should in the best case have a minimal impact when the string to be stripped already starts with ANSI styling.
I currently don't have a good way to run a benchmark script on the real use case of reedline.

sholderbach:
Don't have the indication that it matters and the setup is only working around limitations of `strip_ansi_escapes`

PR #287, opened 2022-01-30
Init prompt without superfluous line at the bottom

sholderbach:
The prompt initialization was still hardcoded for the two line prompt.
Now all adaptation to account for the number of prompt lines is done
during repaint


jntrnr:
I wondered what caused that

PR #285, opened 2022-01-29
list things with `reedline --list`

fdncred:
This is a quick hack to list things. It's totally meant as a conversation starter and not a finished implementation. Feel free to hack on and submit to this PR.

If this was to be used exclusively to pass information to e-q we could `impl Display for blah_enum` and output the text that the keybindings in e-q want to see when parsed. I prototyped that here too.

I'm thinking something like this needs to be turned into several exposed functions so that, in e-q, we can call `get_modifiers()`, `get_modes()`, `get_keycodes()`, etc. and show them to the user so they know what is map-able via keybindings.

```rust
impl Display for PromptEditMode {
    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
        match self {
            PromptEditMode::Default => write!(f, "default"),
            PromptEditMode::Emacs => write!(f, "emacs"),
            PromptEditMode::Vi(_) => write!(f, "vi_normal\nvi_insert"),
            PromptEditMode::Custom(s) => write!(f, "{}", format!("custom_{}", s)),
        }
    }
}
```

```
    ~/src/forks/reedline   list_things  cargo r -- --list                                               ✔  01:40:41 PM 
   Compiling reedline v0.2.0 (/Users/fdncred/src/forks/reedline)
    Finished dev [unoptimized + debuginfo] target(s) in 1.33s
     Running `target/debug/reedline --list`
--Key Modifiers--
Alt
Control
Shift
None

--Modes--
Default
Emacs
Vi_Normal
Vi_Insert
Custom_

--Key Codes--
Backspace
Enter
Left
Right
Up
Down
Home
End
PageUp
PageDown
Tab
BackTab
Delete
Insert
F(1)
F(2)
F(3)
F(4)
F(5)
F(6)
F(7)
F(8)
F(9)
F(10)
F(11)
F(12)
Char('a')
Null
Esc

--Reedline Events--
None
HistoryHintComplete
HistoryHintWordComplete
ActionHandler
CtrlD
CtrlC
ClearScreen
Enter
Esc
Mouse
Resize(0, 0)
Edit([])
Repaint
PreviousHistory
Up
Down
Right
Left
NextHistory
SearchHistory
Paste([])
Multiple([])
UntilFound([])
Menu("")
MenuNext
MenuPrevious
MenuUp
MenuDown
MenuLeft
MenuRight
MenuPageNext
MenuPagePrevious

--Edit Commands--
MoveToStart
MoveToLineStart
MoveToEnd
MoveToLineEnd
[...]

PR #291, opened 2022-02-01
Strum enums

elferherrera:
Implementing a trait that allows to print enum cases


fdncred:
See my thoughts in #285

PR #292, opened 2022-02-02
Polishing in the painter

sholderbach:
- Correct allocation behavior of `coerce_crlf`
- Consolidate names to `screen_width/height`
- Simplify the `PromptCoordinate` down to the row
- Encapsulate the large buffer state readonly
- Consolidate line requirement calculion

### DIFF
--- a/src/painter.rs
+++ b/src/painter.rs
@@ -353,7 +353,7 @@ impl Painter {
             self.stdout
                 .queue(Print(format!(" [h{}:", screen_height)))?
                 .queue(Print(format!("w{}] ", screen_width)))?
-                .queue(Print(format!("y{}] ", self.prompt_start_row)))?
+                .queue(Print(format!("y:{} ", self.prompt_start_row)))?
                 .queue(Print(format!("rm:{} ", remaining_lines)))?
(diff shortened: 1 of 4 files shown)


src/painter.rs
```diff
@@ -262,16 +250,23 @@ impl Painter {
...
 
-    pub(crate) fn terminal_cols(&self) -> u16 {
+    pub(crate) fn screen_width(&self) -> u16 {
         self.terminal_size.0
     }
 
-    pub fn remaining_lines(&self) -> u16 {
-        self.terminal_size.1 - self.prompt_coords.prompt_start.1
+    fn remaining_lines(&self) -> u16 {
+        self.screen_height() - self.prompt_start_row
```
elferherrera, line 262: Im wondering if only for sanity this should be a saturated sub. I dont know what behaviour we get during screen resizing

elferherrera (approved):

src/painter.rs
sholderbach: Good point, but should be safe as resize touches both at the same time. This wasn't the case for the initialization step, next commit unifies that

PR #297, opened 2022-02-04
Keybindings functions

elferherrera:
Simplifying the keybinding functions and correcting default keybindings

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -558,7 +558,7 @@ impl Reedline {
                             self.completer.as_ref(),
                         );
 
-                        if menu.get_values().len() == 1 {
+                        if menu.get_values().len() == 1 && self.quick_completions {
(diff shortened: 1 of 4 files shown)


fdncred:
is this working right? I only have 1 keybinding for `vi_normal`.

sholderbach:
The vi normal mode parser is handwritten and not available as a mapping. Q how we want to document it. Exposing it to configuration is definitely non trivial

fdncred:
@sholderbach is there no way `vi_normal` can be more keybind-y? I'm looking at edit_mode/vi/commands.rs parse_command() and a lot of that looks like keybindings. But maybe these shouldn't be changeable. I mean, having a thing in vi mode means it will work like vi where $ moves to the end, etc. so, maybe we should just leave some things hard-coded here.

PR #310, opened 2022-02-10
Bump dependencies and edition to 2021

sholderbach:


sholderbach:
For us it's not a problem and most projects (e.g. deno) that build with recent compilers shouldn't take issue with that. Some other library crates seem to mainly strive towards low minimal compiler versions for things that use age-old compilers in distros with long release cycles

fdncred:
ya, that's a @jntrnr question. i know things can get messed up easily with this.

jntrnr:
@sholderbach - yeah I was wondering if we should update both at the same time. So, if you want to bump the version for reedline, let's also plan to do a PR for nushell, too, so they upgrade together.

sholderbach:
Gonna merge this one, we can do a `nu-ansi-term` bump with the release of nushell 0.60

Issue #315, opened 2022-02-17
How to custom reedline history input msg


baoyachi:
I need to know the history input and when. Can this be customized ?

fdncred:
a datetime stamp isn't currently stored in the history, so that would take some design and a pr i think. a datetime stamp will definitely be there if we ever get around to implementing the sqlite version of history.

baoyachi:
I see now reedline record in file. [https://github.com/nushell/reedline#integrate-with-custom-history](https://github.com/nushell/reedline#integrate-with-custom-history)

baoyachi:
May be can solve the problem of custom display first.

PR #317, opened 2022-02-17
add  history input and when

baoyachi:
# work for https://github.com/nushell/reedline/issues/315

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ history.txt
 target-coverage/
 tarpaulin-report.html
 .vscode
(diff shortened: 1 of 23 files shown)


fdncred:
I'm guessing that we're going to save this change for when we implement the historydb.

sholderbach:
Closing this for now as the history API changed with #401 which implements timestamping etc. for the SQlite history backend. Feel free to implement an alternative file based encoding on that API. For compatibility in nushell we will maintain the entry only file format for a while.

Issue #312, opened 2022-02-16
`InjectCommand` event/action for simpler keybinding of full commands


sholderbach:
With the new engine-q driven nushell, keybindings to nushell commands for convenience and not destined for the entry in reedline are currently convoluted hacks of several events.
This will affect the state machine of the editor in ways that are hard to recover if you want to actually keep some state from the unfinished manually typed entry:
e.g.
 ```
 keybindings: [
    {
        name: reload_config
        modifier: alt
        keycode: char_p
        mode: emacs
        event: [
            { edit: { cmd: clear } }
            { edit: { cmd: insertString value: $"source ($nu.config-path)" } }
            { send: Enter }
        ]
    }
  ]
```

Instead we could have a single event `ReedlineEvent::InjectCommand` 
which will return from the `Reedline::read_line` with the given command string but keep the current entry plus additional state (up for debate: e.g. history search, active completions etc.) in the reedline object (change on the nushell side necessary to persist that!!!). 

Design decision would be how to handle draw position after returning to `Reedline::read_line`. Either act like a full line was submitted and draw the sent command on submission or try to preserve the prompt position as if nothing happened (tricky if command were to generate output)

sholderbach:
Closed by #323 

Nushell integration pending

PR #323, opened 2022-02-23
Add event to keybind a command in the hosting app

sholderbach:
Instead of affecting the editor this lets `Reedline::read_line` return
with `Ok(Signal::Success)` containing a predefined command to be executed or
consumed by the app using `reedline` as a line editor.
The state of the editor is preserved. After the command execution in a
REPL the host would call `Reedline::read_line` again and the line editor
would be redrawn on the next free line.
Command will not be added to the history in this config.

Open for debate if this should not expect printed output from the
command and try to redraw on top or reset some state.

Addresses #312

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -499,6 +499,10 @@ impl Reedline {
                 self.input_mode = InputMode::Regular;
                 Ok(EventStatus::Handled)
             }
+            ReedlineEvent::ExecuteHostCommand(host_command) => {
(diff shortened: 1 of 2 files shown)


elferherrera:
I like the idea. It will make running commands smoother. I think in nushell the prompt will be cleaned anyway but I would like to test it there

sholderbach:
@elferherrera right with nushell it will probably nuke the content of the typed line as it creates a separate on every repl loop, the last time I checked

PR #324, opened 2022-02-23
Provide cursor position for highlighting

sholderbach:
Changes the `Highlighter` trait to include the position of the cursor in
the buffer. This allows it to make cursor aware highlighting such as
highlighting matching braces or code blocks

Prerequisite to address nushell/nushell#4325


sholderbach:
Gonna merge and follow up with the corresponding fixes to nushell

PR #326, opened 2022-02-26
Revert crossterm: key modifier issue on win

sholderbach:
crossterm 0.23 seems to have introduced a regression on windows for
decoding keybindings with `ctrl` that uppercases the char.

Tracking issue crossterm-rs/crossterm#636

Might be fixed by crossterm-rs/crossterm#629

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "crossterm"
-version = "0.23.0"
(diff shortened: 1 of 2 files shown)


sholderbach:
@elferherrera can you give it a try if this resolves the ctrl-uppercase issue for you?

fdncred:
I wonder if this needs to be done in nushell too since it's v23?

PR #334, opened 2022-03-06
History improved

elferherrera:
Improvements on the history commands:
- The `!!` can be undone
- The `!10` will select the entry number 10 chronologically (matches history command)
- Values from history menu can be inserted in the prompt 
Animation3


jntrnr:
Looks good - going to land this so we can play with it

PR #336, opened 2022-03-07
Bang command expands only alone in buffer

elferherrera:
Simplifies the logic for the `!` expansion
When more expansions are required, the history menu offers more flexibility


jntrnr:
Does `sudo !!` still work?

PR #354, opened 2022-03-23
end/control-e selects the current completion if there is one

jbr:
this is also the behavior in fish's bindings

thanks!


sholderbach:
Awesome, thanks!

PR #358, opened 2022-03-25
Descriptions for completion suggestions

elferherrera:
Descriptions are added to the suggestions received from the completer. This allows to have help for the suggestion menu
 
image


fdncred:
this should be fun to play with

PR #363, opened 2022-03-27
Fix `menu::menu_functions::find_common_string`

uasi:
Fix `menu::menu_functions::find_common_string` returning wrong index when given non-ASCII strings. This should fix https://github.com/nushell/nushell/issues/4971

### DIFF
--- a/src/completion/default.rs
+++ b/src/completion/default.rs
@@ -348,3 +348,40 @@ impl CompletionNode {
         completions
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn default_completer_with_non_ansi() {
+        use super::*;
+
+        let mut completions = DefaultCompleter::default();
+        completions.insert(
+            vec!["ｎｕｓｈｅｌｌ", "ｎｕｌｌ", "ｎｕｍｂｅｒ"]
+                .iter()
+                .map(|s| s.to_string())
+                .collect(),
+        );
+
+        assert_eq!(
+            completions.complete("ｎ", 3),
+            vec![
+                Suggestion {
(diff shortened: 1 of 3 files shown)


elferherrera:
Nice. Thanks for the help. Have you tested by adding strings to the default completer?

uasi:
I haven't yet. Now I added the following to completion/default.rs and it passed. Should I commit  it as well?

```rs
#[cfg(test)]
mod tests {
    #[test]
    fn default_completer_with_non_ansi() {
        use super::*;

        let mut completions = DefaultCompleter::default();
        completions.insert(vec!["ｎｕｓｈｅｌｌ", "ｎｕｌｌ", "ｎｕｍｂｅｒ"].iter().map(|s| s.to_string()).collect());

        assert_eq!(
            completions.complete("ｎ", 3),
            vec![
                Suggestion { value: "ｎｕｌｌ".into(), description: None, span: Span { start: 0, end: 3 } },
                Suggestion { value: "ｎｕｍｂｅｒ".into(), description: None, span: Span { start: 0, end: 3 } },
                Suggestion { value: "ｎｕｓｈｅｌｌ".into(), description: None, span: Span { start: 0, end: 3 } },
            ]);
    }
}
```

elferherrera:
Yes please, add it. It will help a lot.

By the way, I normally add test strings [here](https://github.com/nushell/reedline/blob/main/src/main.rs#L39) as well. So that you can test it live and check if it feels right for you. I dont know how to write non ansi characters.

uasi:
Updated. For writing full-width letters, this translator might come in handy https://lingojam.com/FullWidthTextGenerator

elferherrera:
Thanks. This looks good

PR #364, opened 2022-03-27
Fix `menu::menu_functions::string_difference`

uasi:
Fix `menu::menu_functions::string_difference` returning wrong substring or even panicking when given non-ASCII strings.


elferherrera:
Those pesky non ansi characters. They got us again :P 
Thanks for your help

PR #366, opened 2022-03-27
Add event to synchronize history explicitly

sholderbach:
This can be used to only synchronize the history with a keybinding

Resets the traditional Ctrl-R search. leaves the basic up/down
navigation in the previous state. TODO: Check effects on menu

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -549,6 +549,13 @@ impl Reedline {
                 }
                 Ok(EventStatus::Handled)
             }
+            ReedlineEvent::SynchronizeHistory => {
+                // this will internally also reset the cursor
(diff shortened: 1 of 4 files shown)


elferherrera:
When is the history updated now?

sholderbach:
This is just an additional event to update the history with a keybinding. To make the sync only happening on closing a session the change to nushells loop is necessary. This would be an escape hatch.

elferherrera:
I see. I dont see any problem with this event. Regarding the menu, I dont think it will be affected if you call this event and then the menu.

sholderbach:
Will close this as the history API has changed significantly. Might serve as inspiration.

Issue #193, opened 2021-12-02
Ignore cursor movements in the undo stack


sholderbach:
At the moment cursor moves are stored along with edits that affect the text in the undo stack.
This is probably not the desired behavior as you typically only want to undo edits (vi style replay of moves would probably a different advanced feature).

## Current state

Moves are recorded, undos will follow the cursor (Current behavior on main)

![undo_moving](https://user-images.githubusercontent.com/15833959/144318272-e4a12930-8cc9-4028-8b78-5a0f8892eec6.gif)

(Moving backwards with left arrow, undo (`Ctrl-z`) a few times and redo (`Ctrl-g`) again)

## Challenge

Simply ignoring the moves with the current architecture of the undo stack causes jumping of the cursor to unexpected positions.

### Example discovered when excluding moves from tracking:

> Excluding the moves as in [85dab62](https://github.com/nushell/reedline/commit/85dab620434f557297be8ec51076f31aab08164d) has the weird side effect that after an undo the cursor jumps back to where it was after the last text changing operation 
> ![undo_test](https://user-images.githubusercontent.com/15833959/144317452-759ec485-2509-4b10-84af-9db942aa9cd7.gif) 
> (Deletion of `is` with `Ctrl+w` and then undo via `Ctrl+z`)

### Potentially problematic part of the current implementation

Currently full line buffer states after edits are stored.
https://github.com/nushell/reedline/blob/82bc3acb64e4d3a3a13b57d19dc014652e7daee2/src/core_editor/editor.rs#L5-L11
https://github.com/nushell/reedline/blob/82bc3acb64e4d3a3a13b57d19dc014652e7daee2/src/core_editor/editor.rs#L166-L223

PR #368, opened 2022-03-28
Edit stack

sherubthakur:
Hey, I was looking at the #193 and wanted to pick it up. Though before I went into fixing it I decided to clean up the undo-redo logic by adding an edit stack in the mix. The benefits that I envision from this approach are 
- We were maintaining edits and a cursor in the `Editor` which seemed like a responsibility that it should not have to deal with 
- Simplifies the undo-redo logic (now we only need to think about maintaining a single stack then keeping track of edits vector and cursor 

Let me know if this is acceptable.

### DIFF
--- a/src/core_editor/edit_stack.rs
+++ b/src/core_editor/edit_stack.rs
@@ -1,0 +1,144 @@
+/// This outlines the operations (interface) that any data strcuture will need to obey to be useful
+/// to the core_editor as undo/redo stack
+pub trait EditStack<T>: Send
+where
+    T: Default,
+    T: Clone,
+    T: Send,
+{
+    /// Go back one point in the undo stack.
+    /// If present on first edit do nothing
+    fn undo(&mut self) -> &T;
+
+    /// Go forward one point in the undo stack.
+    /// If present on the last edit do nothing
+    fn redo(&mut self) -> &T;
+
+    /// Insert a new entry to the undo stack.
+    /// NOTE: (IMP): If we have hit undo a few times then discard all the other values that come
+    /// after the current point
+    fn insert(&mut self, value: T);
+
+    /// Reset the stack to the initial state
+    fn reset(&mut self);
+
+    /// Return the entry currently being pointed to
+    fn current(&mut self) -> &T;
+
+    /// List out all the entries on the undo stack
+    /// Mostly used for debugging. Might remove this one
+    fn edits<'a>(&'a self) -> Box<dyn Iterator<Item = &'a T> + 'a>;
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct BasicEditStack<T> {
+    internal_list: Vec<T>,
+    index: usize,
+}
+
+impl<T> BasicEditStack<T> {
+    pub fn new() -> Self
+    where
+        T: Default,
+    {
+        BasicEditStack {
+            internal_list: vec![T::default()],
+            index: 0,
+        }
+    }
+}
+
+impl<T> EditStack<T> for BasicEditStack<T>
+where
+    T: Default,
+    T: Clone,
+    T: Send,
+{
+    fn undo(&mut self) -> &T {
+        self.index = if self.index == 0 { 0 } else { self.index - 1 };
+        &self.internal_list[self.index]
+    }
+
+    fn redo(&mut self) -> &T {
+        self.index = if self.index == self.internal_list.len() - 1 {
+            self.index
+        } else {
+            self.index + 1
+        };
+        &self.internal_list[self.index]
+    }
+
+    fn insert(&mut self, value: T) {
+        if self.index < self.internal_list.len() - 1 {
+            self.internal_list.resize_with(self.index + 1, || {
+                panic!("Impossible state reached: Bug in UndoStack logic")
+            });
+        }
+        self.internal_list.push(value);
+        self.index += 1;
+    }
+
+    fn reset(&mut self) {
+        self.index = 0;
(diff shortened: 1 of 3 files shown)


src/core_editor/edit_stack.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Reset the stack to the initial state
+    fn reset(&mut self);
+
+    /// Return the entry currently being pointed to
+    fn current(&mut self) -> &T;
+
+    /// List out all the entries on the undo stack
+    /// Mostly used for debugging. Might remove this one
+    fn edits<'a>(&'a self) -> Box<dyn Iterator<Item = &'a T> + 'a>;
+}
```
sholderbach, line 31: This should not be a trait. There is no requirement for multiple coexisting impls and it makes encapsulation by visibility harder. Concrete types are also easier when testing alternative implementations as testing and reviewing a patch without changes to the unit tests is nicer (Instead of having to copy and slightly change the test suite and being unsure if it is testing the same thing. Dang that traits are not concepts in this case :P)

src/core_editor/editor.rs
```diff
@@ -431,7 +396,12 @@ mod test {
                 LineBuffer::from("ab "),
                 LineBuffer::from("ab c")
             ],
-            editor.edits
+            editor
+                .edit_stack
+                .edits()
+                .into_iter()
+                .cloned()
+                .collect::<Vec<_>>()
```
sholderbach, line 404: I am not that fond of that existing test as it doesn't test the behavior we care about and instead the implementation. More valuable would be to see that the correct line buffer state results from doing the actual undo ops

src/core_editor/editor.rs
```diff
@@ -176,60 +174,27 @@ impl Editor {
...
             && !is_after_action
         {
-            self.edits.pop();
+            self.edit_stack.undo();
         }
-        self.edits.push(self.line_buffer.clone());
+        self.edit_stack.insert(self.line_buffer.clone());
 
         Some(())
     }
```
sholderbach, line 197: Q: Should this operation be encapsulated as well? and just take the UndoBehavior enum for the op?

src/core_editor/editor.rs
```diff
@@ -176,60 +174,27 @@ impl Editor {
...
     pub fn remember_undo_state(&mut self, is_after_action: bool) -> Option<()> {
-        self.reset_index_undo();
-
-        if self.edits.len() > 1
-            && self.edits.last()?.word_count() == self.line_buffer.word_count()
+        if self.edit_stack.current().word_count() == self.line_buffer.word_count()
             && !is_after_action
         {
-            self.edits.pop();
+            self.edit_stack.undo();
```
sholderbach, line 191: In this case that naming reads a little confusing at first :D

sholderbach (changes requested):
Awesome that you are picking up #193!

Do you want to land it first with behavior parity or wait for the fix to 193?

src/core_editor/editor.rs
sherubthakur: Was trying to keep things as they were. Now that you mention this I will add some more tests to test the logic. Will not track cursor behaviour in these new tests as I plan to add that with the PR that will fix #193 .

sherubthakur:
Let's land this first with behaviour parity first. Then I will pick the fix in a separate PR.

sherubthakur:
@sholderbach the latest two commits should address the comments that you had. Let me know if there is anything else that needs change.

src/core_editor/edit_stack.rs
```diff
@@ -57,11 +57,6 @@ where
         self.internal_list = vec![T::default()];
     }
 
-    /// List out all the entries on the undo stack Mostly used for debugging. Might remove this one
-    pub(super) fn edits<'a>(&'a self) -> Box<dyn Iterator<Item = &'a T> + 'a> {
-        Box::new(self.internal_list.iter().take(self.index + 1))
-    }
-
     /// Return the entry currently being pointed to
     pub(super) fn current(&mut self) -> &T {
```
sherubthakur, line 61: I am not sure what the convention is in rust land but my assumption is that this particular module should only be accessible in the editor module, so I added this constraint.

src/core_editor/editor.rs
```diff
@@ -375,30 +375,37 @@ impl Editor {
...
+
+    fn str_to_edit_commands(s: &str) -> Vec<EditCommand> {
+        s.chars().map(|c| EditCommand::InsertChar(c)).collect()
+    }
+
+    #[test]
+    fn test_undo_works_on_work_boundries() {
+        let mut editor = editor_with("This is a");
+        for cmd in str_to_edit_commands(" test") {
+            editor.run_edit_command(&cmd);
```
sherubthakur, line 392: I am only making use of run_edit_command here as this is the primary interface for the editor. Rest is all (more or less) implementation detail.

sholderbach: Nice!

src/core_editor/edit_stack.rs
sholderbach: Don't have opinions on that. I tend to spam `pub(crate)` for convenience

src/core_editor/editor.rs
sholderbach: OK, do you want to hold of @sherubthakur? I would be happy enough to merge

sherubthakur: Yup let's hold off that discussion (i.e. not make it part of this PR). I think this particular PR is mergable too.

PR #371, opened 2022-03-29
Fix the first example in the readme so it can be run

wmstack:
The example in the read me doesn't actually run until you add a return value for propagating io results.


sholderbach:
Thanks for the catch! Signature might change with #367

PR #372, opened 2022-03-29
Ignore cursor while undoing

### LINKED ISSUE #193: Ignore cursor movements in the undo stack

sherubthakur:
Fixes #193

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -1,6 +1,65 @@
 use super::{edit_stack::EditStack, Clipboard, ClipboardMode, LineBuffer};
 use crate::{core_editor::get_default_clipboard, EditCommand, UndoBehavior};
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct UndoNode {
+    line_buffer: LineBuffer,
+    edit_commands: Vec<EditCommand>,
+}
+
+impl Default for UndoNode {
+    fn default() -> Self {
+        Self {
+            line_buffer: LineBuffer::new(),
+            edit_commands: Vec::new(),
+        }
+    }
+}
+
+impl UndoNode {
+    fn new(line_buffer: LineBuffer, edit_commands: Vec<EditCommand>) -> Self {
+        Self {
+            line_buffer,
+            edit_commands,
+        }
+    }
+
+    fn has_only_movements(&self) -> bool {
+        self.edit_commands
+            .iter()
+            .all(|edit_command| match edit_command {
+                EditCommand::MoveToStart
+                | EditCommand::MoveToEnd
+                | EditCommand::MoveToLineStart
+                | EditCommand::MoveToLineEnd
+                | EditCommand::MoveToPosition(_)
+                | EditCommand::MoveLeft
+                | EditCommand::MoveRight
+                | EditCommand::MoveWordLeft
+                | EditCommand::MoveWordRight
+                | EditCommand::MoveRightUntil(_)
+                | EditCommand::MoveRightBefore(_)
+                | EditCommand::MoveLeftUntil(_)
+                | EditCommand::MoveLeftBefore(_) => true,
+
+                _ => false,
+            })
+    }
+
+    fn add_edit_command(&mut self, edit_command: EditCommand) {
+        self.edit_commands.push(edit_command);
+    }
+
+    fn has_only_insert_chars(&self) -> bool {
+        self.edit_commands
+            .iter()
+            .all(|edit_command| match edit_command {
+                EditCommand::InsertChar(_) => true,
+                _ => false,
+            })
+    }
+}
+
 /// Stateful editor executing changes to the underlying [`LineBuffer`]
 ///
 /// In comparison to the state-less [`LineBuffer`] the `Editor` keeps track of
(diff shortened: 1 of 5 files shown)


sholderbach:
This does not solve the weirdness of the cursor positions after undoing/redoing if there were moves between operations (see #191 for how I tried the same assignment of `UndoBehavior` it before)

sholderbach:
No worries. Another edge case you might want to write a test for when ignoring moves:

* Type two words
* Move the cursor inside the first word.
* Add a char
* Move into the second word 
* Add chars inside

Undo should:
* Revert edit on second word
* Revert char addition on first word
* Revert the typed words

sherubthakur:
Based on the cases that you(@sholderbach) have outlined here it appears that state-only info will be insufficient to produce the undo behavior that is desired.

Let's consider all the approaches that we can take by only using state-only info from the undo-stack followed by an approach where we store more info.

## Case 1
**Using word count** (The present approach): We can only detect if there is a new word, we can't detect if there is a change to any of the words.
### Example
```
this is an examle|
<jump to m's position>
this is an exam|le
<insert character `p`>
this is an examp|le
```
## Case 2
Using word differences: We can only detect if a word was changed. In this case, we will track everything. Which we don't want.

## Case 3
Using more sophisticated techniques like detecting where the change was made to the word. 
```
if the word count for the present buffer is the same as that of the current buffer on undo stack
     check if the character was appended to the last word
     if the character was appended
        merge this edit to the current entry on the undo stack
    else
        add a new entry to the stack
```

IMO if we follow this approach we will end up with extremely complicated logic in our hands to maintain. 
Though the bigger problem is that it will still fail to get rid of the cursor jumping case, which has been outlined in https://github.com/nushell/reedline/pull/191#issuecomment-984074822.

## Case 4 (Storing more info)
We can approach this problem by storing more info in the UndoStack. Namely all the EditCommad that are chunked together in a particular node on the undo-stack.

This keeps the logic extremely simple and easy to maintain. Though on the other hand, we store a lot more info. I am not sure how big a concern this would be. I haven't done any form of testing to see if there is any perceivable degradation in performance. 

Not outlining the logic for this in the comment as this is what the PR has gone with. One can see how this is implemented by looking at the PR. All the changes pertaining to it are in the commit with the title "Rough sketch of fixing the undo bug".

Once a reviewer is convinced that this approach is admissible. I will edit history to have a nicer commit log and also incorporate any feedback. 

If there are alternate apporaches to solve this, please feel free to outline them.

src/core_editor/editor.rs
```diff
@@ -181,10 +181,15 @@ impl Editor {
         self.line_buffer = val.clone();
     }
 
-    pub fn remember_undo_state(&mut self, is_after_action: bool) -> Option<()> {
```
sherubthakur, line 184: Having a function do two different things based on a boolean that is passed usually results in unreadable and
unmaintainable code and this is no different, in my opinion.

We reap the following advantages (if this is done)
- users of the interface don't have to worry about what that boolean flag is.
   - Given that Rust does not have kwargs the readability of passing booleans into a function results in having to read the 
      interface and documentation (or implementation) of the particular function. Which we can also avoid if we don't do it.
- removes the possibility of creating a bug because a wrong boolean value was selected.

src/core_editor/editor.rs
```diff
@@ -374,6 +414,50 @@ impl Editor {
 
         self.line_buffer.insert_str(string);
     }
+
+    fn setup_edit_stack(&mut self, command: EditCommand) {
+        println!("\n\n{:?}", command);
+        println!("Current (edit stack): {:#?}", self.edit_stack);
+        println!("Current (Line Buffer): {:#?}", self.line_buffer);
```
sherubthakur, line 421: Will remove these. This commit is more like a draft PR.

src/core_editor/editor.rs
sholderbach: I think the already existing `UndoBehavior` enum would be a prime candidate to make that explicit and exhaustive (what `is_after_action` means is elusive to me now)

sholderbach:
I think that the end result will be a combination of 1 and 4 (I think word units to undo seem comfortable). I don't know if for the full test compliance we actually need to store the full history of `EditCommand`s (would theoretically allow to expose names of changes for a menu), Could suffice to keep the last `EditCommand` around and have state machine (which basically just needs the `UndoBehavior` (maybe needs more variants)) if changes are necessary, but for the cursor movements on undo redo we probably need to know, when to move to the position after an action or before an action.

sholderbach:
@sherubthakur thanks for giving this a try! I think #460 finds a way to remove the existing snag points, so we can close this one.

PR #374, opened 2022-04-01
bashisms feature for reedline

elferherrera:
Moved all bashisms behind a feature flag
Added `!$` to get the last token from the last command run


fdncred:
Thanks for the `!$` - my life is complete :)

PR #375, opened 2022-04-01
format descriptions to the left

elferherrera:
aligns suggestions descriptions to the left 

image


fdncred:
wow, that looks a lot harder than I though it would be. thanks for this!

PR #378, opened 2022-04-02
generic reedline menus

elferherrera:
Changes the completion and history menu to become generic templates that can be used with different sources. These templates can be used in nushell and they could be fed using a nushell block


fdncred:
I predict this will be fun to play with. :)

PR #388, opened 2022-04-07
change up prompt with traditional nushell colors

fdncred:
This is totally up for debate. Just looking for something less boring, and less ugly blue. :) I'm up for just about any colors other than everything being the same boring blue.

image
that's in windows - left prompt is `green bold` (like old nushell), indicator is `cyan bold` (like old nushell), right prompt is `dark orange`. left prompt replaces `home dir` with `~` if appropriate and also changes `\\` to `/`.

### DIFF
--- a/src/prompt/default.rs
+++ b/src/prompt/default.rs
@@ -86,7 +86,7 @@ fn get_working_dir() -> Result<String, std::io::Error> {
         },
     };
     let new_path = if path_str != homedir {
-        path_str.replace(&homedir, "~").replace("\\", "/")
+        path_str.replace(&homedir, "~")
     } else {
         path_str
     };
--- a/src/prompt/default.rs
+++ b/src/prompt/default.rs
@@ -77,7 +77,20 @@ impl DefaultPrompt {
 
 fn get_working_dir() -> Result<String, std::io::Error> {
     let path = env::current_dir()?;
-    Ok(path.display().to_string())
+    let path_str = path.display().to_string();
+    let homedir: String = match env::var("USERPROFILE") {
+        Ok(win_home) => win_home,
+        Err(_) => match env::var("HOME") {
+            Ok(maclin_home) => maclin_home,
+            Err(_) => path_str.clone(),
+        },
+    };
+    let new_path = if path_str != homedir {
+        path_str.replace(&homedir, "~").replace("\\", "/")
+    } else {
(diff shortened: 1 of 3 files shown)


src/prompt/default.rs
```diff
@@ -77,7 +77,20 @@ impl DefaultPrompt {
 
 fn get_working_dir() -> Result<String, std::io::Error> {
     let path = env::current_dir()?;
-    Ok(path.display().to_string())
+    let path_str = path.display().to_string();
+    let homedir: String = match env::var("USERPROFILE") {
+        Ok(win_home) => win_home,
+        Err(_) => match env::var("HOME") {
+            Ok(maclin_home) => maclin_home,
+            Err(_) => path_str.clone(),
```
sholderbach, line 85: This would in the failure case always result in `~`?

src/prompt/default.rs
```diff
@@ -77,7 +77,20 @@ impl DefaultPrompt {
...
+        },
+    };
+    let new_path = if path_str != homedir {
+        path_str.replace(&homedir, "~").replace("\\", "/")
```
sholderbach, line 89: Let the Windows users live in their world with the backwards slashes :P

sholderbach (commented):
Like the styling. Do we want to keep `get_prompt_color()` around? Feels like an unnecessary special case

src/prompt/default.rs
fdncred: ya, i'm good with whatever.

src/prompt/default.rs
fdncred: I tested it, by changing `USERPROFILE` to `USERPROFILE2` which doesn't exist and on WSL I changed `HOME` to `HOME2` which doesn't exist. Help me understand what you mean please.

src/prompt/default.rs
fdncred: removed the replace \ with / part

fdncred:
> Like the styling. Do we want to keep `get_prompt_color()` around? Feels like an unnecessary special case

We're still using `get_prompt_color()` for getting the left prompt color.

fdncred:
confirmed that it looks good in apple's terminal.app.
Screen Shot 2022-04-07 at 6 36 56 PM
that orange may be too much - other suggestions?

PR #392, opened 2022-04-11
Hack to allow disabling hints.

sholderbach:
With the current architecture the hints aren't optional and always
active per default.
This can be annoying in environments without ANSI colors as the hint
gets indistinguishable from the normal text.
Simple workaround: have a NoOpHinter

Alternative make the hinter an option throughout its use


sholderbach:
Let's do this properly

PR #399, opened 2022-04-17
Add explicit `EditCommand::InsertNewline`

sholderbach:
Uses the system newline separator

Example in main with `Alt+Enter` (according to @fdncred `cmd-option-enter` or `shift-option-enter` on macOS to get the same key event)


fdncred:
this is with `shift-option-enter` on macOS
Screen Shot 2022-04-20 at 6 39 30 AM

PR #400, opened 2022-04-17
Add better options for screen clearing

### LINKED ISSUE #120: Move the clear screen/`Ctrl-L` logic inside the line editor engine
We currently handle the `Ctrl-L` screen clearing command by exiting the `read_line()` method with the `Signal::CtrlL` and keeping the line buffer state as is. This requires the `read_line()` method to respect the previous state of `Reedline` and collaboration from the side of the consumer. 
This might be less than optimal. We might consider moving the `Ctrl-L` handling into the `Reedline` engine with a call to the painter and maybe provide the Option to disable this. A public method to clear from the side of the consumer is already existing.

sholderbach:
**Warning API break:** removes a variant from `Signal`

- Remove the `Signal::CtrlL` as it requires explicit handling and
cooperation but we offer configurable keybindings already
- Default keybinding of `Ctrl-l` is `ReedlineEvent::ClearScreen` that
directly performs the scrollback buffer preserving clear of the current
screen.
- Add `ReedlineEvent::ClearScrollback` to perform screen clearing with
removal of the scrollback buffer. (Demo will now do that on clear as
seen in bash, zsh etc.)

Fixes #120

Helps with nushell/nushell#5089

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,6 @@ fn main() -> io::Result<()> {
                 println!("\nAborted!");
                 break Ok(());
             }
-            Signal::CtrlL => {
-                line_editor.clear_screen().unwrap();
-            }
         }
     }
 }
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -384,6 +384,13 @@ impl Reedline {
         Ok(())
     }
 
+    /// Clear the screen and the scollback buffer of the terminal
+    pub fn clear_scrollback(&mut self) -> Result<()> {
+        self.painter.clear_scrollback()?;
+
+        Ok(())
+    }
+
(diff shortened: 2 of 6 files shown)


stormasm:
@sholderbach Thanks for working on this PR...

It will be interesting to see how it works on the 

https://support.apple.com/en-asia/guide/terminal/welcome/mac

Regarding #5089

https://github.com/nushell/nushell/issues/5089

I will test this after this PR lands...

And let you know what  I find out....

Cc: @fdncred

sholderbach:
Will land and make the changes in nushell to accept the new api/make the keybindings accessible.

It sends the same clearing commands in ANSI that are part of @fdncreds `clsterm` command (minus the style reset).

fdncred:
beautiful!

PR #402, opened 2022-04-19
Use a default terminal size if reported terminal size is 0, 0

DhruvDh:
I have am using reedline on an remote shell rendering on the web. The terminal size is incorrectly reported as 0, 0 (NOT due to a bug in crossterm), which leads to divide by zero panics.

Instead of panicking, I thought maybe having a default size makes sense. 
Feel free to not merge this, I can just use this as a patch. But I think any terminal size is better than pacicking.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -95,7 +95,16 @@ impl Painter {
     /// [`Painter::handle_resize()`] instead
     pub(crate) fn initialize_prompt_position(&mut self) -> Result<()> {
         // Update the terminal size
-        self.terminal_size = terminal::size()?;
+        self.terminal_size = {
+            let size = terminal::size()?;
+            // if reported size is 0, 0 -
+            // use a default size to avoid divide by 0 panics
+            if size == (0, 0) {
+                (80, 24)
+            } else {
+                size
+            }
+        };
         // Cursor positions are 0 based here.
         let (column, row) = cursor::position()?;
         // Assumption: if the cursor is not on the zeroth column,


sholderbach:
Without accurate size/position reporting there will likely be drawing artifacts. Which remote terminal is affected by that so we can keep track of possible issues? Does it implement the resize protocol (or is it in fact fixed to VT100 80x24)? 
But panics due to div by zero are definitely less welcome than a little weird painting.
One thing I am unsure about, if certain terminals/multiplexers might collapse to `0,0` when minimizing. In this case we probably want to hold off on painting in general instead of assuming the typical `(80,24)`.

src/painting/painter.rs
```diff
@@ -95,7 +95,16 @@ impl Painter {
     /// [`Painter::handle_resize()`] instead
     pub(crate) fn initialize_prompt_position(&mut self) -> Result<()> {
         // Update the terminal size
-        self.terminal_size = terminal::size()?;
+        self.terminal_size = {
+            let size = terminal::size()?;
+            // if reported size is 0, 0 - 
+            // use a default size to avoid divide by 0 panics
+            if size == (0, 0) {
+                (24, 80)
```
sholderbach, line 103: Should be columns, rows
https://docs.rs/crossterm/latest/crossterm/terminal/fn.size.html
```suggestion
                (80, 24)
```

DhruvDh:
Resize doesn't seem to reset to 0, 0 - https://capture.dropbox.com/FtxJZ9jcptGZdEZt

I have opened a ticket with them also, still waiting to hear back from them. I think they're using xterm.js

sholderbach:
Weird, the right side prompt gets lost but regular stdout gets wrapped consistently.

DhruvDh:
I don't think it's being resized, just a rendering error. Earlier in the video some parts of the scrollback buffer are rendered on size also.

Here is me printing the size the way crossterm computes it for unix in a loop - https://capture.dropbox.com/M0vQ5TUC9iVTAOu4

DhruvDh:
Is this unlikely to be merged?

sholderbach:
Sorry for not coming back to you. I think for the case of the terminal that starts without a size (i.e. most likely wrong implementation of the ANSI terminal spec) this fix sounds like the most reasonable. Resizes on the other hand can also trigger similar panics. In my testing `gnome-terminal` and `kitty` were protecting themselves against unreasonable sizes. With `alacritty` on the other hand I was able to trigger a panic on a vertical resize that collapsed the number of rows. I don't want to hide a bug behind other weird behavior but I think there are no perfect mitigations here so a solution similar to yours for the initial size could be the best for the resizes as well.

PR #405, opened 2022-04-20
Add vim's `C` binding

zim0369:
I have successfully made the changes you proposed.


zim0369:
What should I name the struct's value for `shift+s`?

sholderbach:
How about `RewriteCurrentLine`

PR #406, opened 2022-04-20
vim's 's' & 'S' bindings

zim0369:
Can you assure me that `cut_current_line()` is the correct function for this implementation?

I'm sorry again about duplicate commits. It's concerning that it might annoy you to see so many commits again and again. If that's a problem then please tell me. I'll redo my fork.

Also, was it correct behavior to make a different PR for implementation of every binding?

Can you also confirm that `x` binding is not copying the content to the clipboard?

I'll also try to work on `r` binding.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -48,6 +48,7 @@ impl Editor {
             EditCommand::ReplaceChars(n_chars, str) => self.replace_chars(*n_chars, str),
             EditCommand::Backspace => self.line_buffer.delete_left_grapheme(),
             EditCommand::Delete => self.line_buffer.delete_right_grapheme(),
+            EditCommand::CutChar => self.cut_char(),
             EditCommand::BackspaceWord => self.line_buffer.delete_word_left(),
             EditCommand::DeleteWord => self.line_buffer.delete_word_right(),
             EditCommand::Clear => self.line_buffer.clear(),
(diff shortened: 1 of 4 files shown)


sholderbach:
This won't leave an empty line to type on if you have multiple lines, so there are small extra steps necessary. (Should be possible with the existing edit commands I think.)

To try it out you can run it with `cargo run -- --vi`

sholderbach:
Having separate features as separate PR's is fine with me, can help with bisecting issues but we are not religious if there is a set of similar features all at once.

As soon as you don't need your commits on the `main` branch anymore you can `git reset --hard 2176424cec673255291a29cb72b1e14672ac9da9` to get back to the base of your changes and run `git pull` to get the `upstream` `main` 
In this command `217....` is the commit ID just before your changes (in this case the one being tagged `v0.4.0`) Attention `reset --hard` will nuke the tracked files in your directory as well.

For general logistical discussion you can talk to the team and folks on discord as well (we have a `#reedline` channel). Might be easier for some stuff than on issue/pr comments.

zim0369:
> This won't leave an empty line to type on if you have multiple lines, so there are small extra steps necessary. (Should be possible with the existing edit commands I think.)

The current behavior is similar to that of fish's. So do you want me to still change it to clear the entire line up to the prompt?

I think I'm fine with both the behaviors but I also think there should be a behavior to clear the entire thing(buffer do you call it?) for which `ctrl+c` seems fine though it doesn't clear the entire thing but starts with a new prompt but I think `ctrl+c` should be the one clearing the entire thing so that it saves the user some terminal space for cancelled commands.

zim0369:
I think I've made a mess. I made another PR for the `s` binding and I don't know how it got merged with this commit. Probably because of the same title but case insensitive on github's side. So then I changed the title.

Now I only made a commit on my side that fixes the issue `EditCommand::Delete` where on cutting a single character the content wasn't getting copied to the clipboard while `x` & `s` should copy the content to the clipboard. I tried to implement without affecting `EditCommand::Delete` as that might be in use elsewhere. My current implementation must be messy. I'm sorry for that.

sholderbach:
I rebased your changes in a different branch with the pull request #417. For future PRs please try to use separate "feature" branches so you can keep the `main` branch with the same history as `nushell/reedline` and your branches or commits just reflect the changes you want to contribute/get reviewed. 

If you want to clean up your fork without redoing the clone, there are the following steps on the command line:

Make sure you are on branch `main` (else `git switch main`)
Assumption: Your fork (zim0369/reedline) is the remote `origin` and `nushell/reedline` is the remote `upstream` (Confirm via `git remote -v`)

Get the current state of the upstream `main` into the git database:
```
git fetch upstream
```
Make sure you have stashed or contributed all changes as commits on a separate branch! (The next step will overwrite those, commits on `main` not part of a PR will also get hidden (recoverable via `git reflog` magic until garbage collection))

**Now the overwriting operation assuming upstream is the remote name for the nushell repo and origin is your fork**
```
git reset --hard upstream/main
```

and to get it to github
```
git push --force orgin main
```

PR #411, opened 2022-04-29
line buffer editor

### LINKED ISSUE #409: Make it possible to edit the current command in an editor (nushell)
This was originally a feature request for nushell:
https://github.com/nushell/nushell/issues/5329

Please make it possible to edit the current command in an editor (e.g. VIM).
I.e. add support for something like ZSH's `edit-command-line`, bound to a keyboard shortcut.

I love this feature from ZSH (and BASH), and without it I can't switch to nushell.
I have a workflow where I essentially store lots of small scripts in my ZSH history. For example, I have a script to build and test a set of binaries, and the script is maybe 20 lines. Because I have several projects with similar layouts, it's quick for me to create a new project, reverse search to find a command (script) similar to the one I want, quickly edit it in VIM, and then run it.

You might think the answer is to write these scripts to files, but it's actually much faster to keep them in my ZSH history: 1) I can quickly find just what I want using reverse search, 2) I can effortlessly keep multiple "versions" of the scripts without using git, 3) some commands are complicated enough they benefit from a real editor but they're not worth saving to a file.

Here's a GIF that shows a bit of what I mean:
https://dev.to/chhajedji/bash-edit-command-in-your-editor-and-execute-directly-30ef

elferherrera:
Introduces a new event to open a line editor to edit the line buffer

Note. There is an issue that I cannot figure out. After you open the line buffer (ctr + e), edit your line and save the result, if you press `ctr+c` there is a panic in reedline. I've been looking everywhere but I cannot find where it is coming from.  Any suggestions on where that error could come from?

closes #409

### DIFF
--- a/src/edit_mode/keybindings.rs
+++ b/src/edit_mode/keybindings.rs
@@ -170,4 +170,5 @@ pub fn add_common_keybindings(kb: &mut Keybindings) {
         KC::Char('n'),
         ReedlineEvent::UntilFound(vec![ReedlineEvent::MenuDown, ReedlineEvent::Down]),
     );
+    kb.add_binding(KM::CONTROL, KC::Char('e'), ReedlineEvent::OpenEditor);
 }
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -115,7 +115,9 @@ pub struct Reedline {
     menus: Vec<ReedlineMenu>,
 
     // Text editor used to open the line buffer for editing
-    text_editor: String,
+    // The first element in the tuple is the command used to open
+    // the line buffer and the second one is the file extension used for the temp file
+    buffer_editor: Option<(String, String)>,
 }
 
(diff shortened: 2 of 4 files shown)


kubouch:
Have you tried printing the backtrace? That should tell you where the error originates from.

elferherrera:
Yeap. Tried that but to backtrace is printed

kubouch:
Have you run it with `RUST_BACKTRACE=1`? It should print a backtrace.

elferherrera:
Tried and nothing comes out

kubouch:
That's super weird. How could it not print a backtrace? Are you sure you tried with a debug build? `RUST_BACKTRACE=full`?

elferherrera:
image

elferherrera:
I was afraid it was going to be that case. Well, I guess we can merge it and add it to nu

PR #412, opened 2022-04-30
Only partial complete suggestions when input is a prefix

Tropid:
This changes the partial completion mechanism so that it only replaces the entered text when the entered text is a prefix of the suggestion. This fixes problems with the fuzzy matching (see https://github.com/nushell/nushell/pull/5320#issuecomment-1110255908).

For example:

Suggestions: `build`, `all-build`
Before this PR: `build<tab>` -> ` ` (shared prefix is empty, therefore the entered input is replaced with an empty string)
After this PR: `build<tab>` -> `build` (there are suggestions which don't share a common prefix, therefore no partial completion happens)

To avoid a worse experience with file path completion I'd suggest to also consider merging https://github.com/nushell/nushell/pull/5387 in nushell.

This PR also adds some unit tests for the partial completion mechanism.

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -478,7 +478,11 @@ impl Menu for ColumnarMenu {
             let index = index.min(value.len());
             let matching = &value[0..index];
 
-            if !matching.is_empty() {
+            // make sure that the partial completion does not overwrite user entered input
+            let extends_input =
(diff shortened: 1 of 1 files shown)


fdncred:
since this goes with the other two nushell PRs that have already landed, let's land this. 
Thanks!!

PR #424, opened 2022-05-09
Start developer documentation

sholderbach:
## Goals:

- Provide pointers to further reading for folks not too familiar with the
problem set on the terminal
- Explain the architecture a bit to get people up to speed

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,0 +1,49 @@
+# Reedline internal developer documentation
+
+**Attention** The README files in the source folders are primarily intended to document the current implementation details and quirks as well as requirements we currently strive towards.
+It is not intended as the public documentation for library users!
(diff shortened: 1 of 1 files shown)


herlon214:
That's amazing, I've been wondering the same thing for Nushell. So the future folks won't need to struggle trying to get some domain knowledge

stormasm:
@sholderbach really nice to see this in there --- especially important as more projects start thinking about using reedline

sholderbach:
@stormasm I hope we get improvements to our public documentation as well for the people from other projects that want to use it. My intended recipient with this PR are more the (potential) contributors to reedline that might wonder about some quirks or are confused by some of our architecture choices. Also a good excercise to see the things that are weird and could be refactored and improved.

sholderbach:
I want to land the philosophical section first and come back to the description of the current code organization.

PR #425, opened 2022-05-09
Fix vi-mode word motions

sadmac7000:
The 'w' motion goes to the beginning of the next word to the right, not
the end. The 'e' motion, which is introduced here, goes to the end, but
lands on the last character, not after it.

### DIFF
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -759,6 +759,35 @@ mod test {
     }
 
     #[rstest]
+    #[case("", 0, 0)] // Basecase
+    #[case("word", 0, 3)] // Cursor on top of the last grapheme of the word
+    #[case("word and another one", 0, 3)]
+    #[case("word and another one", 3, 7)] // repeat calling will move
+    #[case("word and another one", 4, 7)] // Starting from whitespace works
+    #[case("word\nline two", 0, 3)] // Multiline...
+    #[case("word\nline two", 3, 8)] // ... contineus to next word end
+    #[case("weirdö characters", 0, 5)] // Multibyte unicode at the word end (latin UTF-8 should be two bytes long)
+    #[case("weirdö characters", 5, 17)] // continue with unicode (latin UTF-8 should be two bytes long)
+    #[case("weirdö", 0, 5)] // Multibyte unicode at the buffer end is fine as well
+    #[case("weirdö", 5, 5)] // Multibyte unicode at the buffer end is fine as well
+    #[case("word😇 with emoji", 0, 3)] // (Emojis are a separate word)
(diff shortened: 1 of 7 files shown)


src/core_editor/line_buffer.rs
```diff
@@ -179,6 +179,24 @@ impl LineBuffer {
             .unwrap_or_else(|| self.lines.len())
     }
 
+    /// Cursor position *at end of* the next word to the right
+    pub fn word_right_end_index(&self) -> usize {
+        self.lines[self.insertion_point..]
+            .split_word_bound_indices()
+            .find(|(i, word)| *i + word.len() > 1 && !is_whitespace_str(word))
+            .map(|(i, word)| self.insertion_point + i + word.len() - 1)
+            .unwrap_or_else(|| self.lines.len().saturating_sub(1))
```
sholderbach, line 188: Keep in mind that UTF-8 encoded strings can contain characters longer than 1 byte, and further string operations would panic if the character boundaries are violated.

Test case:

`word😇 with emoji`
go to insertion point `0`, hit `e`

We currently operate on the grapheme level i.e. multiple code characters that get displayed as one character on the screen (e.g. accents on some European scripts, joined emojis (most terminals actually don't fuse them), CRLF etc.)

```suggestion
        self.lines[..self.word_right_index()]
            .grapheme_indices(true)
            .next_back()
            .map_or(0, |(idx, _)| idx)
```

sholderbach (changes requested):
Thanks for tackling this! Also great that you fixed the vi parser tests already.
There is a small issue with the treatment of Unicode characters.

src/core_editor/line_buffer.rs
sholderbach: Sorry the particular suggestion will get stuck on the last character.

sadmac7000: new version should fix it

src/core_editor/line_buffer.rs
```diff
@@ -179,6 +179,30 @@ impl LineBuffer {
...
+            .split_word_bound_indices()
+            .filter_map(|(i, word)| {
+                word.grapheme_indices(true)
+                    .next_back()
+                    .map(|x| x.0)
+                    .filter(|k| *k != 0 && !is_whitespace_str(word))
+                    .map(|x| self.insertion_point + x + i)
+            })
+            .next()
+            .unwrap_or_else(|| self.lines.len().saturating_sub(1))
```
sholderbach, line 194: The default case will still panic for a `ö` in the last character position like this.

sholderbach:
Do you mind me pushing a little test set for that onto you branch?

sholderbach (approved):
Awesome! Thanks

PR #428, opened 2022-05-22
fix: list_menu not accounting for index + indicator

ahkrr:
The problem:

https://user-images.githubusercontent.com/44893716/169713547-304e506a-34d1-4a0b-913b-a21280d1ab56.mp4

The code didn't account for the index and the indicator in its calculation for the number of lines in the list_menu.
Now the unicode_width of the indicator and the number of digits of the index are accounted for.

### DIFF
--- a/src/menu/list_menu.rs
+++ b/src/menu/list_menu.rs
@@ -9,6 +9,7 @@ use {
     },
     nu_ansi_term::{ansi::RESET, Style},
     std::iter::Sum,
+    unicode_width::UnicodeWidthStr,
 };
 
 const SELECTION_CHAR: char = '!';
@@ -219,7 +220,13 @@ impl ListMenu {
                         None => (lines, None),
                         Some(total_lines) => {
                             let new_total_lines = total_lines
-                                + self.number_of_lines(&suggestion.value, painter.screen_width());
+                                + self.number_of_lines(
+                                    &suggestion.value,
+                                    //  to account for the index and the indicator e.g. 0: XXXX
+                                    painter.screen_width().saturating_sub(
+                                        self.indicator().width() as u16 + count_digits(lines),
+                                    ),
+                                );
 
                             if new_total_lines < available_lines {
                                 (lines + 1, Some(new_total_lines))
@@ -583,8 +590,18 @@ impl Menu for ListMenu {
     /// Calculates the real required lines for the menu considering how many lines
     /// wrap the terminal and if an entry is larger than the remaining lines
     fn menu_required_lines(&self, terminal_columns: u16) -> u16 {
-        self.get_values().iter().fold(0, |acc, suggestion| {
-            acc + self.number_of_lines(&suggestion.value, terminal_columns)
+        let mut entry_index = 0;
+        self.get_values().iter().fold(0, |total_lines, suggestion| {
+            //  to account for the the index and the indicator e.g. 0: XXXX
+            let ret = total_lines
+                + self.number_of_lines(
+                    &suggestion.value,
+                    terminal_columns.saturating_sub(
+                        self.indicator().width() as u16 + count_digits(entry_index),
+                    ),
+                );
+            entry_index += 1;
+            ret
         }) + 1
     }
 
@@ -662,6 +679,19 @@ fn number_of_lines(entry: &str, max_lines: usize, terminal_columns: u16) -> u16 
     };
 
     lines
+}
+
+fn count_digits(mut n: usize) -> u16 {
+    // count the digits in the number
+    if n == 0 {
+        return 1;
+    }
+    let mut count = 0;
+    while n > 0 {
+        n /= 10;
+        count += 1;
+    }
+    count
 }
 
 #[cfg(test)]


sholderbach:
Good catch! The fix looks a bit gnarly but seems to work for me.

ahkrr:
I'm not that attached to my code. If someone has a better solution to fixing the problem, then this PR can be closed and a different solution implemented.

sholderbach:
I think with the current setup this is the way to go for now. Refactoring the menu layouting code can come later

src/menu/list_menu.rs
```diff
@@ -219,7 +220,13 @@ impl ListMenu {
...
                         Some(total_lines) => {
                             let new_total_lines = total_lines
-                                + self.number_of_lines(&suggestion.value, painter.screen_width());
+                                + self.number_of_lines(
+                                    &suggestion.value,
+                                    //  to account for the index and the indicator e.g. 0: XXXX
+                                    painter.screen_width().saturating_sub(
+                                        self.indicator().width() as u16 + count_digits(lines),
+                                    ),
+                                );
```
blindFS, line 229: @ahkrr Sorry to bother, I'm trying to understand this in order to fix #842 

Got really confused for 2 reasons:
1. Why `self.indicator()` here, we only show indicator in the prompt line, right? Shouldn't we adjust `available_lines` for that matter?
2. Why subtracting the divisor? Shouldn't we add length to the dividend?

blindFS: @sholderbach Would you like to take another look at this?

ahkrr: If I revert this PR the accounting for how many lines the list_menu will take is once again off for lines that are just a bit longer than the terminal_width. This is the case because here counting the index or indicator matters.
In other cases the counting is only accidentally correct because the index or indicator doesn't result in line breaks.

To question 1.)
Edited: 
This either changed over time or was somthing i misunderstood. 
`self.indicator()` refers not to the multiline indicator but the marker. 

My reasoning for this is wrong. 
The `self.indicator()` is the multiline indicator.
I added both the `self.indicator()` and the `index` because lines can have the `index` (in the example below, 4: or 5:) or the multiline indicator (in the example the ::: ) and figuring out which line will have one but not the other would be more difficult. 
I always account for both and accept that some lines will have a bit of space free at the end.
This will result in some lines causing a premature linebreak. (the arrow in the example below)
Which is better than underestimating the number of lines and having the menu scroll outside the terminal window. 
``` nu
~? match            06.07.2025 11:22:47
4: let jo = 'azz' | into glob; 
:::match 'zzjozz' { 
:::        $jo => { 'string' }, 
:::        _ => { 'not a very good list # <- here space is wasted because the index is accounted for
' } 
:::}
5: let jo = 'azzjoz*' | into glob;
:::match 'zzjozz' {
:::        $jo => { 'string' },
:::        _ => { 'not a very good list 
' }
:::}
Page 2: records 4 - 7  total: 340  [FUL
L]
```

To question 2.)
Figuring out how much to add to each line is more difficult than reducing the available space that the lines are allowed to take. It is not impossible, but my approach worked when I tested it. 
It also resulted in fewer codechanges.

Testing the line counting again just now, I noticed that it is still sometimes off. 
I suspect that the last line of the list_menu (in the example, Page 2: records...) is assumed to always only be one line, which isn't always the case, but I could be wrong, and this needs investigating. 

The divide by zero when the width is zero is a different problem and also needs fixing.

My biggest mistake was not including test cases that exercise the edge cases that need to be handled.
Tests for these should definitely be added.

ahkrr: The 
``` rust
self.indicator().width() as u16 + 
```
can be removed from the calculation, which would also partially fix issue 842.
It would still end in a division by zero if the terminal_width was less than the display_width of the index.
For example, a terminal width of 2.

ahkrr: From my understanding, one would still need to account for the space taken by the index. 
That is why reverting this PR entirely causes problems.
Changing the code to
``` rust
painter.screen_width().saturating_sub(count_digits(lines))

// and 
terminal_columns.saturating_sub(count_digits(entry_index)),
```
should mostly work except when the terminal is smaller than the display_width of the index.

blindFS: @ahkrr thanks for the explanation, the `multiline indicator` is something I wasn't aware of.
But I still don't get the `saturating_sub` part,

image

You can see here the index is only shown on the first line, so how many lines it should take? I think it's `(number_of_a + length_of_'0: ')/term width`, if we ignore `multiline indicator` for now.

Your code worked because I believe (haven't tested yet) any large enough number would fix that issue.

ahkrr: I forgot to explain. It just so happens that the index is always at least 3 wide '0: ', which is also the width of the multiline indicator. So it just accidentally works out correct.
For indeces like '11: ' the index is wider than the indicator, so that is also not a problem.

blindFS: If we take `multiline indicator` into account, then things become more complicated:

image

You can see the required number of lines depends heavily on the item text (1st line is not full, indicator on 2nd line but not 3rd or 4th), and there's possible ansi characters in the indicator and unicode issues blablabla.

ahkrr: Each line will have either the index or indicator so it just accidentally works out that accounting for the index on each line will also account for the indicator.

ahkrr: If the multiline_indicator is userconfigurable it would also need to be taken care of in the calculation.

blindFS: Maybe we can try returning a loose upper bound, then mark it as a hack, and make the divisor at least 1 for the "divide by 0" panic.

ahkrr: I just found `multiline_marker: String,`, if that is indeed the multiline marker than changin my original PR from 
`self.indicator().width()` to  `self.multiline_marker.width()` may work

ahkrr: Yeah. Adding one to the `terminal_columns` if it is `0` would be a quick workaround.

ahkrr: Because `saturating_sub` is used we could always subtract an additional `1` and then add `1` back again. 
If the saturating_sub reached `0` it would result in `1`, and if it didn't result in `0`, we would be back at the original number.

blindFS: It can be tricky:

image

Where the text is `(\n\n\n\n\n\n\n\n\n\n\n)`

ahkrr: One approach could also be 
``` rust
terminal_columns.saturating_sub(
  cmp::max(self.multiline_marker().width(), count_digits(entry_index))
)
``` 
This way the lines with an index and the ones with a multiline_marker would allign when displayed.

ahkrr: When users have funky entries in their history, it shouldn't be a supprise if the menu_entry is also funky. 
Trying to account for all possible history entries isn't feasible, I think. 
As long as the menu doesn't break or panic, it should be fine.

blindFS: > One approach could also be
> 
> ```rust
> terminal_columns.saturating_sub(
>   cmp::max(self.multiline_marker().width(), count_digits(entry_index))
> )
> ```
> 
> This way the lines with an index and the ones with a multiline_marker would allign when displayed.

That doesn't fix the issue of my last screenshot, where the text length is actually pretty short, but the native newline characters make it take many lines.

Also not all lines are shown, it got truncated with the `...` notation.

ahkrr: The `...` notation is a nushell functionality I didn't implement, but I guess it is there so that one entry with many lines doesn't take up all the space. 
The multiple newline character issue I would argue like in my previous comment:
When users have funky entries in their history, it shouldn't be a supprise if the menu_entry is also funky.
Trying to account for all possible history entries isn't feasible, I think.
As long as the menu doesn't break or panic, it should be fine.

blindFS: Sorry, I got lost in my last 2 comments, actually it's already taken good care of, here:

https://github.com/nushell/reedline/blob/e4221b954ccc04940f3ed505ed4984d92fbe1f4c/src/menu/list_menu.rs#L594-L615

Seems to me the correct thing to do is prepend the prefix (index for the first line, multiline indicator for later ones) to the first argument in line-606/611. Then we don't subtract anything from terminal width.

ahkrr: I don't think the multiline_indicator is already written to suggestion.value. I seem to remember that I had each suggestion.value dbg!()_printed and it was only the entry_text with nothing added to it.

ahkrr:
``` rust
terminal_columns.saturating_sub(
  (cmp::max(self.multiline_marker().width(), count_digits(entry_index)) + 1)
) + 1
```
still seems like a good try to me.

blindFS: How about the prepending solution I just updated in my last comment?

blindFS: I think the number would be perfectly accurate then.

ahkrr: In `fn number_of_lines` you don't know which index it is you need to prepend to the entry and you would also need to pass self to the function to have access to the multiline_marker. 

You would also allocate a string for each line. 
For some reason the calculations are done 3 times for each entry when displaying the list_menu in the terminal.
 
But you are correct that your approach would give more accurate results. 

But instead of creating a new string for each line you could also do the terminal_columns calculation I suggested above seperately in the place where you would do the prepending to the line.

The number is still not perfectly accurate because of a different issue I mention in my first comment that is unreleated to what we are currently doing.

blindFS: Performance won't be a big problem because it stops early when the total number exceeds the quota available.

I tend to make the number as accurate as possible. And meanwhile it makes the code easier to understand.

Sorry I didn't get why is the number not perfectly accurate here. Can you elaborate it more?

ahkrr:
``` rust
    fn printable_entries(&self, painter: &Painter) -> usize {
        // The number 2 comes from the prompt line and the banner printed at the bottom
        // of the menu
        let available_lines = painter.screen_height().saturating_sub(2);
```
I think this is an issue.
The available_lines assumes that the prompt line and the banner at the bottom each only take up 1 line but this is incorrect. They can also take up more than one line if the terminal_width is small enough.

blindFS: Seems a lot of work to do.

Would you please provide some steps for me to reproduce your problem so that I can make sure my change won't result in a regression.

ahkrr: For the prompt_line one would also need to account for the self.indicator().width() here, I think.
I don't know what I can offer to help you reproduce my problems. I simply make the terminal window narrow and small enought so that the prompt_line, bottom line and the entries wrap and each take up more than 2 lines. Then I TAB to jump through the list and look if parts of the list_menu scroll outside off the screen. 

Adding tests for theses cases would be the proper thing to do. I already mentioned that this was a mistake of my original PR.

ahkrr: Once you have a PR ready I could compile your branch on my machine and play around with the results and look if I can still break something. 

I am also not a maintainer of nushell or reedline. I am just a fan that got notified by you.

blindFS: > Once you have a PR ready I could compile your branch on my machine and play around with the results and look if I can still break something. 
> 
> 
> 
> I am also not a maintainer of nushell or reedline. I am just a fan that got notified by you.

Thanks for your time. I couldn't reproduce the behaviour from your video with this PR reverted.

I thought a reproduce guideline would save you some effort. But never mind, I will let you know when I think a proper fix is ready.

ahkrr: One thing that I just remembered is that when you test things, make sure to test the edgecases. I think the problems that I was seeing were when the index (or multiline_marker) + the line where 0, 1, 2 ( a small number of characters) wider than the terminal. In the other cases the line calculations were technically right, but the way they were calculated was still wrong. 
I was still able to reproduce wrong behavior with this PR reverted.

blindFS: @ahkrr Thanks for the tip, it makes perfect sense after reading through the code. I'll try it.

PR #429, opened 2022-05-23
Typo

petrisch:
Reedline has now... instead of Nushell has now...


petrisch:
At least thats how I read it, right?

PR #432, opened 2022-05-26
Move some commonly used keybinds to common keybinds

Artturin:
* Move ctrl+d keybinding from emacs to common keybinds
* Move ctrl+w keybinding from emacs to common keybinds


sholderbach:
Thanks, `Ctrl-d` missing was annoying me whenever I tried vi-mode. I think there are some additional quirks that may need to be addressed but let's land that first (e.g. `CutWordLeft` will place it in the same cut buffer as `dw` e.g.)

PR #434, opened 2022-05-26
Move ctrl+a & ctrl+e to common navigation binds

Artturin:


sholderbach:
As far as I remember neither `zsh` with `bindkey -v` nor `vim` do that by default but it's not that much of an issue for me ;)

Artturin:
> As far as I remember neither `zsh` with `bindkey -v` nor `vim` do that by default but it's not that much of an issue for me ;)

It should, at least, be available in insert mode.
If in the future it should be something else in normal mode then it can be overridden

Artturin:
tested by adding
``
[patch."https://github.com/nushell/reedline"]
reedline = { path = "../reedline"}
```

to the nushell Cargo.toml

sholderbach:
Good catch with the `DeleteWord` vs `BackspaceWord`

PR #435, opened 2022-05-30
Fix clippy lints that will become warnings

sholderbach:
clippy from the nightly future

### DIFF
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -56,7 +56,7 @@ pub trait Completer: Send {
 }
 
 /// Suggestion returned by the Completer
-#[derive(Debug, Default, Clone, PartialEq)]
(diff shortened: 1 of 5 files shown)


sholderbach:
This [lint](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_to_owned)/change is nasty as  `ansi_term` and thus `nu_ansi_term` on their `AnsiGenericString` implement `Display` to apply the styling and `Deref` to just return the underlying `&str` and as [`ToString` is implemented for any `T: Display + ?Sized` ](https://doc.rust-lang.org/std/string/trait.ToString.html#impl-ToString-6) the to_string implementation relies on the `Display` defined for the type alias `AnsiString` and not on the `impl ToString for str` for the `&str` accessible through `Deref`. 

Applying the suggested fix will remove syntax highlighting!!!
image

PR #439, opened 2022-06-14
Expose the `History` query functionality

sholderbach:
Makes the history querying structs and enums pub as it seems reasonable
for now. Useful to create nushells `history` command.

Might make sense to change that API to something simpler cleaner later.

Relevant to nushell/nushell#5721


phiresky:
looks good to me

PR #440, opened 2022-06-14
Prepare 0.7.0 release

sholderbach:
Includes the new History API and their backends

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "reedline"
-version = "0.6.0"
+version = "0.7.0"
 authors = ["JT <jonathan.d.turner@gmail.com>"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "reedline"
-version = "0.6.0"
+version = "0.7.0"
 authors = ["JT <jonathan.d.turner@gmail.com>"]
```
fdncred, line 4: at some point we may want to change this to 'The Nushell Project Developers', or whatever nushell says.

sholderbach: Oops too late :D

fdncred: it can wait until next time - no biggy

PR #442, opened 2022-06-17
Add vi r (replace)

drbrain:
This is an incomplete implementation as it omits the behaviors for CTRL-V, CTRL-E, and CTRL-Y:

> Replace the character under the cursor with {char}. If {char} is a <CR> or <NL>, a line break replaces the character.  To replace with a real <CR>, use CTRL-V <CR>.  CTRL-V <NL> replaces with a <Nul>.
>
> If {char} is CTRL-E or CTRL-Y the character from the line below or above is used, just like with |i_CTRL-E| and |i_CTRL-Y|.  This also works with a count, thus `10r<C-E>` copies 10 characters from the line below.

The similarly-named ReplaceChars is used for history substitution and
is more complicated than we need (n_chars will always be 1, string will
always be a single character string), but could be used if the extra
replace_char() function is too much.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -53,6 +53,7 @@ impl Editor {
             EditCommand::InsertChar(c) => self.insert_char(*c),
             EditCommand::InsertString(str) => self.line_buffer.insert_str(str),
             EditCommand::InsertNewline => self.line_buffer.insert_newline(),
+            EditCommand::ReplaceChar(chr) => self.replace_char(*chr),
             EditCommand::ReplaceChars(n_chars, str) => self.replace_chars(*n_chars, str),
             EditCommand::Backspace => self.line_buffer.delete_left_grapheme(),
             EditCommand::Delete => self.line_buffer.delete_right_grapheme(),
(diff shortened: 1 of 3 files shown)


sholderbach:
Awesome, didn't know about the `Ctrl-<char>` special operations. 

One thing that might be improved is the cancellation of `<action>r<char>`. Currently `drl` is cancelled as invalid and executes `l`. `yrl` on the other hand executes `rl`. 

Nice that it is also tolerant to `<multiplier greater than available chars>r<char>`. I personally can live with the extra inserts. Don't know if we want to abort that operation in the future like vim (I know that for this change some changes to the vi emulation may be necessary instead of just concating operations)

drbrain:
I'll look at cancellation

For `<multiplier greater than available chars>r<char>` I think update this to match vim behavior of doing nothing (I wasn't aware of it). Can redline ring the terminal bell like vim does?

drbrain:
While investigating I passed the multiplier value into `parse_command()` but that quickly seemed too clunky without a larger refactoring of command parsing that seems too ambitious at present. Your explanation of the implementation makes me think there's a more-understandable way to make a check for overrun

sholderbach:
If you have no objections I would vote for landing the replace action and polishing the rough edges later.

PR #444, opened 2022-06-17
fix double bang

WindSoilder:
Fix nushell issue: https://github.com/nushell/nushell/issues/5788

I think `!!` should behave like `!-1`, and for `!-1`, the result index is `1`


fdncred:
nice job @WindSoilder. thanks for fixing this.

rgwood:
Thanks for the fix!

PR #445, opened 2022-06-17
introduce LastCommand variant for parse bang result

WindSoilder:
Sadly find the fix in https://github.com/nushell/reedline/pull/444 breaks `!!` in history search menu... Because `!!` will select the item with index `1`.

To avoid this, introduce `ParseAction::LastCommand` for `!!`, then `reedline` can handle `!!` and `!3` better.


fdncred:
thanks

PR #450, opened 2022-07-09
Support vi mode ; and , motions

drbrain:
The f, F, t, and T left-right motions can be repeated with the ; motion, or reversed with the , motion:

> ;
> Repeat latest f, t, F or T [count] times. See cpo-;
>
> ,
> Repeat latest f, t, F or T in opposite direction [count] times. See also cpo-;

I named these "to-till" motions as the first word of the description for the f and T motions are "To" and "Till" respectively.

Add storage for the last To or Till motion on the `Vi` struct, and pass it through to `parse_command()` so `;` or `,` can re-use it.

When a to-till motion is fully parsed the new to-till motion is stored in the `Vi` struct.

I'm not the happiest with finding a new to-till motion in the parsed command. I can't think of a better way to match it due to the `Vec`, but am open to suggestions.

It's probably reasonable to make `parse_command()` a method of `Vi` if my implementation is acceptable.

### DIFF
--- a/src/edit_mode/vi/command.rs
+++ b/src/edit_mode/vi/command.rs
@@ -152,17 +152,15 @@ where
         }
         Some(';') => {
             let _ = input.next();
-            match &vi.last_to_till {
-                Some(to_till) => Some(Command::ReplayToTill(to_till.clone())),
-                _ => None,
-            }
+            vi.last_to_till
+                .as_ref()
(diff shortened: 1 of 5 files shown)


drbrain:
I need to fix `,` behavior. For `fabcdefabcdef` with the cursor on the first `f` issuing `ff;,,` should return you to the first `f`. In my implementation `,` swaps between the second and third `f`s

PR #452, opened 2022-07-13
Support swtiching the case of a char with ~ in vi

drbrain:
> Switch case of the character under the cursor and move the cursor to the right.

vi behavior is to only swap the case of an ASCII character so this is
implementation matches that behavior


sholderbach:
Awesome thanks!

PR #459, opened 2022-08-03
Vi mode add support for d0, d^, c0, and c^

bnprks:
The main feedback question for this pull request is for the multi-line handling I've made it so `d0` and `d$` will only delete to the end of the **_line_**, but the previous behavior of `d$` deleted the end of the **_input_**, even for multi-line inputs.

I believe my change is more consistent with vi's behavior, but let me know if the prior `d$` behavior is preferable and I can edit accordingly.

### DIFF
--- a/src/edit_mode/vi/command.rs
+++ b/src/edit_mode/vi/command.rs
@@ -306,7 +306,7 @@ impl Command {
                 Motion::LeftBefore(c) => {
                     Some(vec![ReedlineOption::Edit(EditCommand::CutLeftBefore(*c))])
                 }
-                Motion::Start => Some(vec![ReedlineOption::Edit(EditCommand::CutFromStart)]),
+                Motion::Start => Some(vec![ReedlineOption::Edit(EditCommand::CutFromLineStart)]),
             },
             Self::Change => match motion {
                 Motion::End => Some(vec![
--- a/src/edit_mode/vi/command.rs
+++ b/src/edit_mode/vi/command.rs
@@ -359,7 +359,7 @@ impl Command {
                     ReedlineOption::Event(ReedlineEvent::Repaint),
                 ]),
                 Motion::Start => Some(vec![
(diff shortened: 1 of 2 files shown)


sholderbach:
Thanks, thats a good catch, yeah vi mode should respect the lines. 
`^` slightly differs from `0` in that in vi it moves to the first non-whitespace character instead of first column/character. But I am happy to take a follow up PR to deal with that.

src/edit_mode/vi/command.rs
```diff
@@ -306,7 +306,7 @@ impl Command {
                 Motion::LeftBefore(c) => {
                     Some(vec![ReedlineOption::Edit(EditCommand::CutLeftBefore(*c))])
                 }
-                Motion::Start => None,
+                Motion::Start => Some(vec![ReedlineOption::Edit(EditCommand::CutFromStart)]),
```
sholderbach, line 309:
```suggestion
                Motion::Start => Some(vec![ReedlineOption::Edit(EditCommand::CutFromLineStart)]),
```

src/edit_mode/vi/command.rs
```diff
@@ -358,7 +358,10 @@ impl Command {
                     ReedlineOption::Edit(EditCommand::CutLeftBefore(*c)),
                     ReedlineOption::Event(ReedlineEvent::Repaint),
                 ]),
-                Motion::Start => None,
+                Motion::Start => Some(vec![
+                    ReedlineOption::Edit(EditCommand::CutFromStart),
```
sholderbach, line 362:
```suggestion
                    ReedlineOption::Edit(EditCommand::CutFromLineStart),
```

sholderbach (changes requested):
Small change necessary to make sure that `d0` `c0` don't delete the previous lines as well.

sholderbach (approved):

PR #460, opened 2022-08-05
Improve undo functionality

### LINKED ISSUE #193: Ignore cursor movements in the undo stack

### LINKED ISSUE #194: Design decision: Coalesce manual delete or backspace as a single entry on the undo stack
The current undo stack is able to coalesce the set of characters typed to form a word to a single undoable entry for a word.

If we think this is reasonable, it would be great to also enable that for manual deletion of words by hitting either delete or backspace multiple times 

Currently the word boundary is defined as whitespace: 
https://github.com/nushell/reedline/blob/82bc3acb64e4d3a3a13b57d19dc014652e7daee2/src/core_editor/editor.rs#L210-L215
https://github.com/nushell/reedline/blob/82bc3acb64e4d3a3a13b57d19dc014652e7daee2/src/core_editor/line_buffer.rs#L280-L283

bnprks:
This provides two improvements to undo functionality:
1. Add menu completions and history navigation to the undo stack
2. Improve coalescing of text editing word-level undo blocks
      - fix #193 
      - fix #194

This is similar to work proposed in #372, but taking an independent approach.

My high-level strategy was:
- Prevent anyone outside of `Editor` from directly mutating the line_buffer.
- Add a parameter to all public `Editor` functions that mutate the line_buffer
  in order to tag how the undo behavior of the edit should work.

Code review questions:
- In the first commit, I fully delete `ReedlineEvent::ActionHandler` and `CircularCompletionHandler`.
  The code looked dead to me, but please correct me if it's still needed.
- Is passing `&mut Editor` rather than `&mut LineBuffer` through all the menu code
  alright? I feel like there's room for a refactor so that menu code returns its
  changes via `Option<LineBuffer>`, rather than messing with `Editor` directly.
  My way minimized the code changes, but I do feel it's a little messy.
- In the third commit, I introduced some churn by e.g. removing the wrapper methods `Editor::set_insertion_point` 
  and `Editor::clear_range`, opting to just call the equivalent methods on `LineBuffer`
  directly. I'm fine reverting since it's just stylistic changes I made in the
  process of adding functionality.


bnprks:
Three commits turned into 6 due to some fixups. I'd be happy to rebase to clean up the commit history, though I'm not exactly sure how to Github would treat the rebase without me starting a new pull request.

sholderbach:
Thanks for tackling those issues! Will probably have to wait till Sunday to look deeper into the diff.

About the history or intermediate results from a completion, would all changes to the displayed buffer be introduced into the UndoStack?

The `CircularCompletionHandler` was the original completion implementation and we left it in as it supports the bash like behavior of modifying inplace (ugh with undo) instead of showing the big list/menu of options. For nushell we currently don't ship it that way and thus haven't really tested the behavior extensively with the most recent code base.
So before it becomes tech debt we can probably remove it, although I find that the completion menu system is screaming for a simplification as a first class citizen (looking at the complicated keybinding configs necessary). But refactoring that would be out of the scope of this PR. There might be valid reasons to use a more minimalist completion in constrained environments, but I don't know if there is currently an actual userbase that uses the different configuration (maybe ask on discord if anybody went through the pain of reconfiguring for the worse experience).

> - Is passing `&mut Editor` rather than `&mut LineBuffer` through all the menu code
>   alright? I feel like there's room for a refactor so that menu code returns its
>   changes via `Option<LineBuffer>`, rather than messing with `Editor` directly.
>   My way minimized the code changes, but I do feel it's a little messy.

Haven't looked into the difference but keeping the diff constrained for the review seems reasonable to me here.

> - In the third commit, I introduced some churn by e.g. removing the wrapper methods `Editor::set_insertion_point` 
>   and `Editor::clear_range`, opting to just call the equivalent methods on `LineBuffer`
>   directly. I'm fine reverting since it's just stylistic changes I made in the
>   process of adding functionality.

Keeping it straightforward for now is probably good. 

In general we probably have to review the state of encapsulation across the public API at some point. 
The `Editor` is the stateful part with a history and the `LineBuffer` is generally OK to hand around and should simplify doing unicode and text object preserving operations, 
So from the perspective of your PR using the editor definitely makes sense. I don't know what we have or should expose to external implementers of menus/completions etc. I will make sure to think about it while looking at it.

(We tend to squash the commits of a PR together so you don't have to be religious about the commit history but commits that explain the steps are always appreciated :) )

src/edit_mode/vi/command.rs
```diff
@@ -306,7 +306,7 @@ impl Command {
                 Motion::LeftBefore(c) => {
                     Some(vec![ReedlineOption::Edit(EditCommand::CutLeftBefore(*c))])
                 }
-                Motion::Start => Some(vec![ReedlineOption::Edit(EditCommand::CutFromLineStart)]),
+                Motion::Start => Some(vec![ReedlineOption::Edit(EditCommand::CutFromStart)]),
```
sholderbach, line 309: That leads to the wrong behavior (see my addendum on #459). Might have still been on your working tree.

src/edit_mode/vi/command.rs
```diff
@@ -359,7 +359,7 @@ impl Command {
                     ReedlineOption::Event(ReedlineEvent::Repaint),
                 ]),
                 Motion::Start => Some(vec![
-                    ReedlineOption::Edit(EditCommand::CutFromLineStart),
+                    ReedlineOption::Edit(EditCommand::CutFromStart),
```
sholderbach, line 362: Same as above

src/edit_mode/vi/command.rs
bnprks: Oh yeah, I had some trouble initially merging your fixes into my working tree. I'll revert this

bnprks:
Thanks for taking a bit of a look already, no rush on the review since it's a lot of changes.
Just pushed a new commit to fix the regressions you noticed, and to strip out the unnecessary wrapper method changes.

> About the history or intermediate results from a completion, would all changes to the displayed buffer be introduced into the UndoStack?

That is the intention for the way it works now, yes. If the UndoBehavior allows merging edits, then it will pop the undo stack and push the new state. So every edit should go on to the `EditStack`, though the total size not grow if the edits can be merged.

I worry a bit about this from an efficiency perspective -- allocating a new `String` on basically every key press. Though the only increased cost from this pull request is for key presses during history/completion, which will now result in a new `String` allocation if they change the contents of the `LineBuffer`. If we move checking `UndoBehavior` to before we run any edits on the `LineBuffer`, then I think it is doable to edit in-place for the cases where we don't need to grow the size of the `EditStack`

> There might be valid reasons to use a more minimalist completion in constrained environments, but I don't know if there is currently an actual userbase that uses the different configuration (maybe ask on discord if anybody went through the pain of reconfiguring for the worse experience).

Do you have a sense of the right discord channel for this? I guess we can also wait to see if anyone else chimes in about my post in the reedline channel.

> In general we probably have to review the state of encapsulation across the public API at some point.
The Editor is the stateful part with a history and the LineBuffer is generally OK to hand around and should simplify doing unicode and text object preserving operations,

My stance from this pull request is that it's fine to give out immutable references to the Editor's LineBuffer, or for outside callers to clone + edit their own LineBuffer. But if direct mutation of the Editor's LineBuffer is possible, then the outside caller can easily introduce bugs if they forget to handle the EditStack appropriately.

bnprks:
> About the history or intermediate results from a completion, would all changes to the displayed buffer be introduced into the UndoStack?

I put together another commit that makes it so that only changes that have to grow the UndoStack will result in any LineBuffer clones being made. Otherwise the Editor's LineBuffer will be edited in-place: https://github.com/bnprks/reedline/pull/1. It's maybe a bit much to throw on to this pull request, but it's ready as a follow up.

bnprks:
Thanks for taking a look! I'm happy to make some more changes to make things really good. I'm pretty new to Rust and open source so it's a good chance for me to learn how to do things well.

> I see you introduced a special variant for history navigation, that seems to already address my concern that undoing after browsing through a ton of history entries shouldn't feel like browsing through all of that again. (Or am I misunderstanding that?)

Yes, that's about right. The purpose here is so that up and down arrows through the history will get collapsed into a single undo-able action, just like typing a whole word or deleting a whole word. The reason it gets a special variant is so that history navigation actions only get merged with other history navigation actions.

> Small thing that might be an issue, we generally do the operations like Backspace or Delete on grapheme level [...] while your special undos use chars [...] Problems should be rare but we might have to specifically test for that [...]

Good catch! The undo logic is just checking the chars for if they are newlines, other whitespace, or neither (aka text). So I think it shouldn't actually matter that we aren't looking at graphemes (unless some graphemes have a mixture of whitespace and non-whitespace characters in them). However, I realize from your comment there might be an issue for `\r\n` in Windows, as deletes into `\r\n` will see the `\r` char rather than the `\n` char and not set a proper undo point. I don't have access to a Windows machine to test the UI, but I've made a fix and added some automated tests.

> There are a few API inconsistencies I found but not sure if it is worth addressing before merging (e.g. some Editor methods now doing undo signaling while sister methods are not doing it.) Could be a source for future bugs but depends on how much we want to iterate on this PR.

I just made a commit that cleans up Editor so that all public methods have undo tracking either with a default behavior or as a parameter. Many of the private methods don't do undo tracking still.

> Blocker for merging right now is that Editor also needs to be made crate level pub via src/lib.rs to implement the Menu trait etc. (I think nushell has some custom implementations there). This will require some additional doc comments.

My last new commit makes Editor public. To keep the public API surface small, I moved a lot of the convenience functions to `pub(crate)` since any `Menu` implementation that had been working with a `&mut LineBuffer` should be translatable given just the `line_buffer` and `set_line_buffer`  functions.

src/core_editor/line_buffer.rs
```diff
@@ -902,13 +922,25 @@ mod test {
...
     #[rstest]
-    #[case("This is a te", 4)]
-    #[case("This is a test", 4)]
-    #[case("This      is a test", 4)]
-    fn word_count_works(#[case] input: &str, #[case] expected_count: usize) {
+    #[case("This is a test", 3)]
+    #[case("This is a test  ", 3)]
+    #[case("  This is a test", 4)]
+    #[case("This  \n    is a test", 4)]
+    fn word_start_works(#[case] input: &str, #[case] expected_count: usize) {
```
sholderbach, line 929: Just to reduce possible confusion: 
```suggestion
    fn word_starts_works(#[case] input: &str, #[case] expected_count: usize) {
```

src/core_editor/line_buffer.rs
```diff
@@ -902,13 +922,25 @@ mod test {
...
+        assert_eq!(expected_count, line_buffer.word_starts());
+        line_buffer.assert_valid();
+    }
+
+    #[rstest]
+    #[case("This is a test", 3)]
+    #[case("This is a test  ", 4)]
+    #[case("  This is a test", 3)]
+    #[case("This  \n    is a test", 4)]
+    fn word_end_works(#[case] input: &str, #[case] expected_count: usize) {
```
sholderbach, line 941:
```suggestion
    fn word_ends_works(#[case] input: &str, #[case] expected_count: usize) {
```

sholderbach (approved):
Awesome thanks! I feel like your code already reads pretty ideomatic, so happy to have you contributing! I think you addressed all my concerns. 

Would like to go ahead with the merge to use it in the next nushell release next week. After landing we can take care of changing the nushell side and benchmark your follow-on PR (looks to me like the rebase should be easy)

src/enums.rs
```diff
@@ -311,25 +309,76 @@ impl EditCommand {
...
-            EditCommand::Undo | EditCommand::Redo => UndoBehavior::Ignore,
+            EditCommand::Undo | EditCommand::Redo => EditType::UndoRedo,
         }
     }
 }
 
-/// Specifies how the (previously executed) operation should be treated in the Undo stack.
+/// Specifies the types of edit commands, used to simplify grouping edits
+/// to mark undo behavior
+#[derive(PartialEq)]
```
sholderbach, line 321: Clippy got a bit more pedantic with todays release: 

```suggestion
#[derive(PartialEq, Eq)]
```

bnprks:
To explain the pre-clippy changes: I just realized those `word_start_works` and `word_end_works` functions aren't needed anymore since I changed to the character-based undo algorithm. 

For what it's worth, it might be worth cherrypicking one of the new Editor functions from my staged pull since I think it will be the best way for Menus to edit the current linebuffer while avoiding clones: https://github.com/bnprks/reedline/blob/nocopy-EditStack/src/core_editor/editor.rs#L136-L143

PR #461, opened 2022-08-09
Ctrl+h edit binding moved from emacs to common

morzel85:
Moving the edit binding to `add_common_edit_bindings` because `Ctrl+h` works in Vim's **Insert** mode as backspace. This behavior is also supported by _vi_ mode in bash.


sholderbach:
Thanks!

PR #464, opened 2022-08-17
first attempt to get crossterm 25 working in reedline

fdncred:
This is the first attempt.
These seem to be working:
```
cargo run -- --list 
```
```
cargo run -- -k
```

This PR adds support for Crossterm 0.25.0. These are the new events supported. Some features are part of supporting the kitty protocol as documented [here](https://docs.rs/crossterm/0.25.0/crossterm/event/struct.PushKeyboardEnhancementFlags.html).

```rust
     /// Focus gained event
     FocusGained,

     /// Focus lost event
     FocusLost,

     /// Paste event for bracketed-paste
     Paste(String),
```
KeyBindings also now support a `KeyEventKind` and `KeyEventState`. This is what they look like.
```rust
pub enum KeyEventKind {
    Press,
    Repeat,
    Release,
}
pub struct KeyEventState: u8 {
    /// The key event origins from the keypad.
    const KEYPAD = 0b0000_0001;
    /// Caps Lock was enabled for this key event.
    ///
    /// **Note:** this is set for the initial press of Num Lock itself.
    const CAPS_LOCK = 0b0000_1000;
    /// Num Lock was enabled for this key event.
    ///
    /// **Note:** this is set for the initial press of Num Lock itself.
    const NUM_LOCK = 0b0000_1000;
    const NONE = 0b0000_0000;
}
```
There are also more `KeyCode`s that are supported now.
```rust
    ///
    /// **Note:** these keys can only be read if **both**
    /// [`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`] and
    /// [`KeyboardEnhancementFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES`] have been enabled with
    /// [`PushKeyboardEnhancementFlags`].


    /// Caps Lock key.
    CapsLock,
    /// Scroll Lock key.
    ScrollLock,
    /// Num Lock key.
    NumLock,
    /// Print Screen key.
    PrintScreen,
    /// Pause key.
    Pause,
    /// Menu key.
    Menu,
    /// The "Begin" key (often mapped to the 5 key when Num Lock is turned on).
    KeypadBegin,
    /// A media key.
    Media(MediaKeyCode),
    /// A modifier key.
    Modifier(ModifierKeyCode),
```
As well as new `Modifier`s
```rust
/// Represents key modifiers (shift, control, alt, etc.).
///
/// **Note:** `SUPER`, `HYPER`, and `META` can only be read if
/// [`KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES`] has been enabled with
/// [`PushKeyboardEnhancementFlags`].
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
pub struct KeyModifiers: u8 {
    const SHIFT = 0b0000_0001;
    const CONTROL = 0b0000_0010;
    const ALT = 0b0000_0100;
    const SUPER = 0b0000_1000;
    const HYPER = 0b0001_0000;
    const META = 0b0010_0000;
    const NONE = 0b0000_0000;
}
```
and `ModifierKeyCode`s
```rust
pub enum ModifierKeyCode {
    /// Left Shift key.
    LeftShift,
    /// Left Control key.
    LeftControl,
    /// Left Alt key.
    LeftAlt,
    /// Left Super key.
    LeftSuper,
    /// Left Hyper key.
    LeftHyper,
    /// Left Meta key.
    LeftMeta,
    /// Right Shift key.
    RightShift,
    /// Right Control key.
    RightControl,
    /// Right Alt key.
    RightAlt,
    /// Right Super key.
    RightSuper,
    /// Right Hyper key.
    RightHyper,
    /// Right Meta key.
    RightMeta,
    /// Iso Level3 Shift key.
    IsoLevel3Shift,
    /// Iso Level5 Shift key.
    IsoLevel5Shift,
}
```
And also `MediaKeyCode`s
```rust
pub enum MediaKeyCode {
    /// Play media key.
    Play,
    /// Pause media key.
    Pause,
    /// Play/Pause media key.
    PlayPause,
    /// Reverse media key.
    Reverse,
    /// Stop media key.
    Stop,
    /// Fast-forward media key.
    FastForward,
    /// Rewind media key.
    Rewind,
    /// Next-track media key.
    TrackNext,
    /// Previous-track media key.
    TrackPrevious,
    /// Record media key.
    Record,
    /// Lower-volume media key.
    LowerVolume,
    /// Raise-volume media key.
    RaiseVolume,
    /// Mute media key.
    MuteVolume,
}
```


fdncred:
There's a couple things that I need help with/understanding @sholderbach or others.

1. There needs to be a way to enable the fancy features I think. This is what I did with the `-k` "listen" functionality as per the documentation but I'm not sure if it should be somewhere else or maybe in the nushell itself?
```rust
    execute!(
         stdout,
         EnableBracketedPaste,
         EnableFocusChange,
         EnableMouseCapture,
         PushKeyboardEnhancementFlags(
             KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES
                 | KeyboardEnhancementFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES
                 | KeyboardEnhancementFlags::REPORT_EVENT_TYPES
         )
     )?;
```
2. I'm not sure what to do with the `Paste` event and need help figuring that out. It's kind of hooked up but probably does nothing. For instance, I added things like these in a few places but I'm not really sure what needs to be done with them. Any ideas?
```rust
Event::Paste(s) => ReedlineEvent::Paste(s),
```
and
```rust
ReedlineEvent::Paste(_) => Ok(EventStatus::Handled), // TODO: Handle Paste
```

sholderbach:
Ah, that's something we should get over the finish line for this release as well.

fdncred:
> Ah, that's something we should get over the finish line for this release as well.

Agreed. I'm just not sure what to do about those two points. I think the rest of the PR is pretty much ready. There wasn't a lot to it. There will probably be some things to change on the nushell side so that the `keybindings blah` commands work right.

fdncred:
`ls_colors` updated to crossterm 0.25 https://github.com/sharkdp/lscolors/pull/54

fdncred:
an issue in nushell about this https://github.com/nushell/nushell/issues/6622 - there are probably more than this one.

stevenxxiu:
I can offer some comments on the bracketed-paste issue.

[Bracketed-paste - Wikipedia](https://en.m.wikipedia.org/wiki/Bracketed-paste) mentions auto-indent. I don't think this is implemented anyway so shouldn't be an issue.

However what is an issue is where pasting many lines runs the first command:

```
echo foo
echo bar
```

In *Zsh* this doesn't happen. The user is free to edit both lines.

fdncred:
@stevenxxiu Feel free to dive in here. We definitely want to move forward with crossterm. I think we're a couple versions behind now. It would be good for someone to "fix" this PR so we can land it.

sholderbach:
Superseded by #560

PR #467, opened 2022-08-24
Basic external printer

GrxE:
Hi all,
I needed an external printer and made this one. For my use-case this seems enough. Any comments or ideas are welcome.


src/engine.rs
```diff
@@ -441,7 +456,16 @@ impl Reedline {
         loop {
             let mut paste_enter_state = false;
 
-            if event::poll(Duration::from_millis(1000))? {
+            #[cfg(feature = "external_printer")]
+            if let Some(ref external_printer) = self.external_printer {
+                if let Ok(line) = external_printer.receiver().try_recv() {
+                    self.painter.print_crlf()?;
```
sholderbach, line 462: I don't know if there might be a good usecase for this printer to not automatically act as `println` compared to `print` but yeah implementing the latter correctly is probably unevenly harder.

What I am unsure about. When invoking this we don't have a nicely defined cursor position (We use the regular ANSI terminal cursor as the user cursor.) If the external content get's printed after the cursor, nothing is stopping the editor from overwriting that output.

fdncred:
What exactly is an `external printer` used for?

src/engine.rs
```diff
@@ -441,7 +456,16 @@ impl Reedline {
...
+                    self.print_line(&line)?;
+                    self.repaint(prompt)?;
```
sholderbach, line 464: `repaint()` it self is now probalby pretty robust to what state the editor is in. But if the prompt origin position is fixed and the printed output scrolls the screen, stuff might shift around and things get overprinted.

src/external_printer.rs
```diff
@@ -0,0 +1,43 @@
...
+impl ExternalPrinter {
+    /// Creates an ExternalPrinter to store lines with a max_cap
+    pub fn new(max_cap: usize) -> Self {
+        let (sender, receiver) = bounded::<String>(max_cap);
+        Self { sender, receiver }
+    }
+    /// Gets a Sender to use the printer externally by sending lines to it
+    pub fn sender(&self) -> Sender<String> {
+        self.sender.clone()
+    }
```
sholderbach, line 24: Why have this and the explicit `print()` at the same time?

src/external_printer.rs
```diff
@@ -0,0 +1,43 @@
...
+        self.sender.send(line.to_string()).unwrap_or_default();
+    }
+
+    /// Convenience method to get a line if any, doesn´t block.
+    pub fn get_line(&self) -> Option<String> {
+        if let Ok(line) = self.receiver.try_recv() {
+            Some(line)
+        } else {
+            None
+        }
```
sholderbach, line 41:
```suggestion
        self.receiver.try_recv().ok()
        
```

I am not sure if that helper is helpful in the long run. Should we deal with the cancellation case of all senders being dropped?

src/external_printer.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+    /// Receiver to get messages if any
+    pub fn receiver(&self) -> &Receiver<String> {
+        &self.receiver
+    }
+
+    /// Convenience method if the whole Printer is cloned, blocks if max_cap is reached.
+    pub fn print(&self, line: &str) {
+        self.sender.send(line.to_string()).unwrap_or_default();
+    }
```
sholderbach, line 33: I think the general expectation is that output you want to print shouldn't get lost. 
Using a bounded channel makes certainly some sense to favor backpressure over unpredictable performance if you try to dump a ton of output.
But suppressing the `Err` case here could drop input if the receiver is done.

src/external_printer.rs
```diff
@@ -0,0 +1,43 @@
...
+use crossbeam::channel::{bounded, Receiver, Sender};
+
+/// An ExternalPrinter allows to print messages of text while editing a line.
+/// The message is printed as a new line, the line-edit will continue below the
+/// output.
+#[cfg(feature = "external_printer")]
+#[derive(Debug, Clone)]
+pub struct ExternalPrinter {
+    sender: Sender<String>,
+    receiver: Receiver<String>,
```
sholderbach, line 11: I forgot that the crossbeam channel by default is MPMC compared to the borked `std::mpsc`. Just cloning everything certainly makes the implementation simpler (while allowing some shenanigans like draining the channel from a different thread.)

sholderbach (commented):
I currently have not much time to build a little test app to see how that behaves in the real world. So take my comments with a grain of theoretical salt. Do you have something where you use that with multiple threads?

GrxE:
Hi, was busy making a small demo, here it is. 
https://github.com/GrxE/reedline_with_external_printer
I will come back to your commens soon. Thank you!

src/external_printer.rs
GrxE: I was thinking of print() as a simple to use convenience method. To have more control on the Sender-side I think it is useful to have a reference of the Sender<String> to make the sending more customizable (timeouts, blocking, etc.).

src/engine.rs
GrxE: This is what I was afraid of... I have to test it more by hand, but until now, all looks good, on my machine ;)

src/engine.rs
GrxE: Here is my small demo:
https://github.com/GrxE/reedline_with_external_printer

GrxE:
How can I reproduce the failure to fix it? When I run 'cargo build', 'cargo run' or 'cargo test', I get no failures.

fdncred:
@GrxE Looks like the CI just burped. It does that sometimes. I re-ran it and it works and is green now.

sholderbach:
Sorry for not coming back to you earlier. 

Thanks for the demo and changes!

I agree that we can probably land this as a MVP without reacting to all potential error cases and instead going for a `fail-safe` panic free route at first. 
So don't worry too much about my API nitpicks.

I was giving it a try with `gnome-terminal` and your demo (both with the Cargo.lock in that repo and an update to your state in this PR). I encountered a few problems with how the prompt will draw over the output line. For example when entering a line that was long enough to wrap the output got overwritten.
More philosophically for the minimal viable product do we want to keep the illusion that there is only the one prompt and the previously submitted ones in the scrollback? Right now to me it appeared that there will be always an unfinished prompt interjected between the outputs from the external printer.

GrxE:
Thanks for giving my code a try!
> I was giving it a try with `gnome-terminal` and your demo (both with the `Cargo.lock` in that repo and an update to your state in this PR). I encountered a few problems with how the prompt will draw over the output line. For example when entering a line that was long enough to wrap the output got overwritten. 

I am aware of this issue, but to fix it I need more insight in the codebase. I am watching the videos but I think this will take a while. If someone has an idea what is going on and how to possibly fix that, could speed things up.

>Right now to me it appeared that there will be always an unfinished prompt interjected between the outputs from the external printer.
This behaviour seems right for me, of course the previously entered partial line could be erased. I can change this (hopefully) if you like. I think this should be implemented as the crate/library owner(s) want it. Please let me know.

sholderbach:
> I am aware of this issue, but to fix it I need more insight in the codebase. I am watching the videos but I think this will take a while. If someone has an idea what is going on and how to possibly fix that, could speed things up.

I think we arrived at the current `Painter` implementation after the video series (basically Fernando, JT and I throwing the hot potato of a visual bug source between us until we were happy enough). This handles keeping track where we remember to start drawing the prompt and all following line and menu components and draw it out sequentially on repaint. We use the cursor storing protocol from ANSI to keep the editor cursor at the point where the user intends it to and restore to this position after finishing drawing (and hide and show appropriately).



> > Right now to me it appeared that there will be always an unfinished prompt interjected between the outputs from the external printer.
> This behaviour seems right for me, of course the previously entered partial line could be erased. I can change this (hopefully) if you like. I think this should be implemented as the crate/library owner(s) want it. Please let me know.

I agree that we should keep the prompts that were submitted by pressing enter, but I'm not sure if we want to keep

``` sh
$ h
text
$ he
text
$ hell
more text
$ hello|
```

If a resulting implementation would be brittle and unstable depending on the terminal implementation we should provide a way to disable it to avoid worse artifacts, but I am not sure if there is a need for the interjected prompt remainders (you would think they were submitted by looking at the scrollback)

GrxE:
> > > This behaviour seems right for me, of course the previously entered partial line could be erased. I can change this (hopefully) if you like. I think this should be implemented as the crate/library owner(s) want it. Please let me know.
> 
> I agree that we should keep the prompts that were submitted by pressing enter, but I'm not sure if we want to keep
> 

> 
> If a resulting implementation would be brittle and unstable depending on the terminal implementation we should provide a way to disable it to avoid worse artifacts, but I am not sure if there is a need for the interjected prompt remainders (you would think they were submitted by looking at the scrollback)

The fix for multiple-lines-buffer was more important - for me. I will try to implement it this way:

```shell
$ enter pressed after this line
text
text
more text
$ hello|
```

GrxE:
I see the point in polishing. It looks far better than before. The only way to get line-artifacts is by resizing the window - and I don´t think I am able to fix this ... So what do you think? Please comment!

sholderbach:
> I see the point in polishing. It looks far better than before. The only way to get line-artifacts is by resizing the window - and I don´t think I am able to fix this ... So what do you think? Please comment!

We generally don't handle resizing that well, so I don't think that is necessarily a show-stopper.

In general, I would prefer landing your PR sooner rather than later without having to stabilize the final API yet. On the nushell side, as far as I am aware, we don't have something immediately in the pipeline which would battle-test your printer. So I would be open to landing it to enable you or other people to experiment with it and get real-world feedback. 
Especially for how the API feels in practice or if we should make particular performance trade-offs, we probably need real use cases. 

The only potential "tech debt" we would take on is having to make sure we also take care of the external printing source when working on improvements to the `Painter` in the future.

sholderbach:
All my limited testing looked excellent after your fixes!

fdncred:
after looking through this a bit, I'm now wondering how I can use "external_printer" to make my prompt async. 🤔

PR #468, opened 2022-08-25
Add shared emacs-mode bindings to vi-mode

bnprks:
GNU Readline has several emacs bindings that are active in vi mode.
Previously some of these were only active in insert mode when GNU
readline has them always active, and some were never activated.
The only deviation from GNU readline behavior here is that I've added
Alt-f and Alt-Right for symmetry -- for some reason these two don't
work in my computer's copy of GNU readline.
Newly active in all modes:
- Alt-Left / Alt-b (word left)
- Alt-Right / Alt-f (word rignt)
- Ctrl-y (paste)
- Ctrl-u (cut from start)
- Ctrl-t (swap graphemes)

Newly active in normal mode:
- Ctrl-h (move left/same as backspace)
- Ctrl-w (backspace word)
- Ctrl-k (cut to end) [not active in insert mode]

### DIFF
--- a/src/edit_mode/vi/vi_keybindings.rs
+++ b/src/edit_mode/vi/vi_keybindings.rs
@@ -20,9 +20,11 @@ pub fn default_vi_normal_keybindings() -> Keybindings {
 
     add_common_control_bindings(&mut kb);
     add_common_navigation_bindings(&mut kb);
-    // Replicate vi's default behavior for Backspace and delete
+    add_common_edit_bindings(&mut kb);
+
+    // Replicate vi's default behavior for Backspace
     kb.add_binding(KM::NONE, KC::Backspace, edit_bind(EC::MoveLeft));
-    kb.add_binding(KM::NONE, KC::Delete, edit_bind(EC::Delete));
+    kb.add_binding(KM::CONTROL, KC::Char('h'), edit_bind(EC::MoveLeft));
 
     kb
 }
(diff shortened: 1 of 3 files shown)


src/edit_mode/vi/vi_keybindings.rs
```diff
@@ -20,20 +20,27 @@ pub fn default_vi_normal_keybindings() -> Keybindings {
 
     add_common_control_bindings(&mut kb);
     add_common_navigation_bindings(&mut kb);
-    // Replicate vi's default behavior for Backspace and delete
+    add_common_edit_bindings(&mut kb);
```
sholderbach, line 23: I am not sure that is really consistent with vi normal mode.

sholderbach (commented):
I don't know if we want to one-for-one replicate the `GNU readline` bindings. 
I personally tried to follow the `vim` defaults as the guidelines for our vi mode behavior.
`zle` for by default doesn't seem to carry over too much of the emacs bindings to vi insert or normal mode. https://linux.die.net/man/1/zshzle

If there are no good usability reasons to also carry the emacs bindings I would hesitate a bit to make sure that the modal changes are easily distinguishable (and we have unassigned keys left for special behavior.) If someone has established a mishmash of vi and emacs as their workflow nobody should stop them from configuring emacs bindings on top of the vi modes (with the caveat for the things the vi parser will intercept).

`Alt-Arrowkeys` seems fine for me and might help with Mac keyboards (but I am not hundred percent sure about what was tricky there)

bnprks:
Okay, that sounds reasonable. I think we should just close this and let users add their own config if they want to emulate GNU readline's mismash of emacs + vim bindings in its vi mode. 

I think it's safer to not try getting into the Alt-arrow bindings, as at least on my mac running iTerm2 or native terminal Alt-left/Alt-right is read by crossterm as Alt-b/Alt-f. So Alt-arrow isn't possible without polluting the vi mode with some other emacs-style movement commands.

PR #469, opened 2022-08-27
Fix panic when using sqlite as history backend

nibon7:
This PR fixes panic when using sqlite as history backend.

Fixes nushell/nushell#6251

We can reproduce the panic with the following steps:
1. `echo %` <kbd>enter</kbd>
2. `echo %%` :boom: 

[asciicast](https://asciinema.org/a/yY3B6vWGhYptvj7j3PUkAIKTU)

### DIFF
--- a/src/hinter/default.rs
+++ b/src/hinter/default.rs
@@ -24,7 +24,11 @@ impl Hinter for DefaultHinter {
                 .expect("todo: error handling")
                 .get(0)
                 .map_or_else(String::new, |entry| {
-                    entry.command_line[line.len()..].to_string()
+                    entry
+                        .command_line
+                        .get(line.len()..)
+                        .unwrap_or_default()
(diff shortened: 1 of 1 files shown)


fdncred:
I still can't reproduce this problem but your change is benign enough so let's allow it and see if it helps other users.

nibon7:
> I still can't reproduce this problem but your change is benign enough so let's allow it and see if it helps other users.

I can reproduce nushell/nushell#6251 when enable sqlite backend :smile_cat:

fdncred:
I'm on macOS and cannot. I exclusively use the sqlite history now everywhere. maybe it was a linux issue or somehow platform specific?

nibon7:
> maybe it was a linux issue or somehow platform specific?

Can't reproduce on Windows, seems linux only.

PR #470, opened 2022-09-05
Add Reedline::has_last_command_context to allow checking if update_last_command_context will fail

unrelentingtech:
This is necessary to avoid the problem with running `executehostcommand` keybindings in nushell right after launching it (without having run actual user-input commands yet).

Since the error types are not exposed outside of the crate, we have to use a check instead of matching on the error.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -377,6 +377,14 @@ impl Reedline {
         self.history.sync()
     }
 
+    /// Check if any commands have been run.
+    ///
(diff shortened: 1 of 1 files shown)


sholderbach:
Thanks for figuring that out and providing a fix! I will land it to get it into this release, but maybe we can harden the API in the future (checking for three things on the nushell side feels a bit brittle)

PR #472, opened 2022-09-05
Allow external manipulation of the command line buffer, fixes #423

### LINKED ISSUE #423: ExecuteHostCommand & current line buffer usage
Hello,

is it possible to use the immediate contents of the line buffer inside of `ReedlineEvent::ExecuteHostCommand` when keybinding is used?

Not sure if using API is appropriate for this use-case:

User:
- types some command into CLI, e.g.: `hello --address`
- presses keybinding e.g. SHIFT + '?'
- keybinding:
  - does NOT edit the pendign line itself
  - executes host-command that includes the line being written, e.g. `help hello --address`
  - keeps the pending line without any change for firther editing (if applicable)

Is above possible either via keybinding declaration & it's command parameter, or other approach?

Explanation:
-  I am using `reedline` together with `clap` crate in REPL mode 
   (clap's `try_parse_from()` insterad of from building from env on binary execution)
- Above behavior might greatly complement `clap`'s ability to provide extensive & automatic help sub-commands that typically come "free" with its `Command`/`Subcommand` types

unrelentingtech:
This allows nushell to implement a command similar to fish's `commandline`:

https://fishshell.com/docs/current/cmds/commandline.html

that would allow commands run via `executehostcommand` bindings to do interesting things to the contents of the input buffer.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -415,6 +415,11 @@ impl Reedline {
         result
     }
 
+    /// Returns the current contents of the input buffer.
+    pub fn current_buffer_contents(&self) -> &str {
+        self.editor.get_buffer()
+    }
+
     /// Writes `msg` to the terminal with a following carriage return and newline
     fn print_line(&mut self, msg: &str) -> Result<()> {
         self.painter.paint_line(msg)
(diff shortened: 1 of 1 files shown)


fdncred:
@sholderbach can we land this? I'd like to get it into nushell to play with now that 68 is out.

sholderbach:
I am excited and curious about nushell/nushell#6492.
My gut reaction was, to be a bit cagy about exposing the full editing behavior with `run_edit_commands`. In theory, you shouldn't be able to do worse crimes than what the user can already do with the keybindings, but right now we control the order of executing and printing. So my strict take would be to make the three operations used by the `commandline` commands separate methods. 
But in principle the nushell side needs the testing so I would also be open to landing this first and then concern our selves with the proper API

sholderbach:
I am gonna land this to unblock the downstream PR for testing. Might polish the API later

PR #474, opened 2022-09-08
Don't bundle sqlite by default

jcgruenhage:
This came up in https://github.com/void-linux/void-packages/pull/39143#issuecomment-1239582933 while talking about bundling or not bundling deps. IMO, libraries like reedline shouldn't force using a bundled version, and if that decision is to be made at all, it should happen in the application, not somewhere deep in the dependency tree. If an application wants sqlite bundled, the application itself can add a dep on rusqlite and add the `bundled` feature, and cargo will use the bundled variant for reedline as well, via feature unification. Disabling the bundling in a similar fashion isn't possible though.

Ideally, I'd like to have an override for this available in libsqlite3-sys, but that hasn't happened yet (see https://github.com/rusqlite/rusqlite/issues/769), so the quick solution for me is patching out the bundling in the deps instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,7 +395,6 @@ version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f0455f2c1bc9a7caa792907026e469c1d91761fb0ea37cbb16427c77280cf35"
 dependencies = [
- "cc",
  "pkg-config",
  "vcpkg",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,6 +395,7 @@ version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f0455f2c1bc9a7caa792907026e469c1d91761fb0ea37cbb16427c77280cf35"
 dependencies = [
+ "cc",
  "pkg-config",
  "vcpkg",
 ]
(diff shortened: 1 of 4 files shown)


jcgruenhage:
I'd say pointing to the documentation in the library and defaulting to `bundled` on the application side would probably be good enough already to cover most of everything. This would mean distributions can patch it in a single place (== the Cargo.toml in the application itself), while those users not installing it from their distribution will not notice a difference.

But yeah, as you can see in your own test cases, with latest Ubuntu it's still not possible to not use the bundled variant, where as the sqlite version shipped by Void Linux is new enough to work as expected for us.

sholderbach:
What about having an additional feature `sqlite-dynlink = ["rusqlite"]` for you guys that want to package it as slim as possible and `sqlite = ["rusqlite/bundled"]` as a safe default. I am not sure what the minimum supported rust version for that is (but I think we are already quite recent in our code base).

jcgruenhage:
As mentioned in the description of the PR: The "proper" way would be to have an env-var override in `libsqlite3-sys` directly, instead of each library in the chain having to make decisions like this.

About a dylib feature in here for  us to enable, that would work with this specific case, but as soon as there's something pulled in from crates.io that depends on reedline, we'd have to patch that lib again because adding a dylib feature that's incompatible with the default features means we can't enable that in the Cargo.toml of the application, because the features are additive and we can't remove the other feature.

If you're not interested in having this merged, I can fully understand it, feel free to just close this PR. It would make my life slightly easier if this got merged, but at the same time it's more of a workaround and less a proper fix anyway.

sholderbach:
Sorry for not coming back on that earlier. 
For now I think having the two features is a good compromise, to enable testing and make it easier for people starting out. As we are most likely used directly used by the executable binary projects, nagging them to take care when packaging should be feasible.

PR #476, opened 2022-09-12
Add duplicate column in sqlite history

Kangaxx-0:
Fix part of https://github.com/nushell/nushell/issues/6502, specifically, address  @sholderbach's comment  

> The history menu we provide out of the box could be more helpful to have functionality to filter for success and dedup afterwards.

image

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1,10 +1,10 @@
 #[cfg(feature = "bashisms")]
+use crate::menu_functions::{parse_selection_char, ParseAction};
 use crate::{
-    history::SearchFilter,
-    menu_functions::{parse_selection_char, ParseAction},
+    result::{ReedlineError, ReedlineErrorVariants},
+    CommandLineSearch, SearchFilter,
 };
 
-use crate::result::{ReedlineError, ReedlineErrorVariants};
 use {
     crate::{
         completion::{Completer, DefaultCompleter},
@@ -504,7 +504,10 @@ impl Reedline {
                             last_ecs.extend(ec);
                         }
                         (ref mut a @ Some(_), other_event) => {
-                            reedline_events.push(ReedlineEvent::Edit(a.take().unwrap()));
+                            reedline_events.push(ReedlineEvent::Edit(
+                                a.take()
+                                    .expect("Cannot take out item from last edit commands"),
+                            ));
 
                             reedline_events.push(other_event);
                         }
@@ -829,6 +832,20 @@ impl Reedline {
                         let buf = self.editor.get_buffer();
                         if !buf.is_empty() {
                             let mut entry = HistoryItem::from_command_line(buf);
+
+                            // check if the command has been seen in the history
+                            if let Ok(dup) = self.history.search(SearchQuery {
+                                filter: SearchFilter::from_text_search(
+                                    CommandLineSearch::Substring(entry.command_line.clone()),
+                                ),
+                                ..SearchQuery::everything(SearchDirection::Backward)
+                            }) {
+                                if dup.is_empty() {
+                                    entry.duplicate = 0;
+                                } else {
+                                    entry.duplicate = 1;
+                                }
+                            }
                             // todo: in theory there's a race condition here because another shell might get the next session id at the same time
                             entry.session_id =
                                 Some(*self.history_session_id.get_or_insert_with(|| {
--- a/src/history/base.rs
+++ b/src/history/base.rs
@@ -189,6 +189,7 @@ mod test {
             cwd: Some(cwd.to_string()),
             duration: Some(Duration::from_millis(1000)),
             exit_status: Some(exit_status),
+            duplicate: 0,
             more_info: None,
         }
     }
--- a/src/history/file_backed.rs
+++ b/src/history/file_backed.rs
@@ -310,6 +310,7 @@ impl FileBackedHistory {
             cwd: None,
             duration: None,
             exit_status: None,
+            duplicate: 0,
             more_info: None,
         }
     }
--- a/src/history/item.rs
+++ b/src/history/item.rs
@@ -79,6 +79,8 @@ pub struct HistoryItem<ExtraInfo: HistoryItemExtraInfo = IgnoreAllExtraInfo> {
     pub duration: Option<Duration>,
     /// the exit status of the command
     pub exit_status: Option<i64>,
+    /// the bit flags of the duplicated command
+    pub duplicate: u8,
     /// arbitrary additional information that might be interesting
     pub more_info: Option<ExtraInfo>,
 }
@@ -95,6 +97,7 @@ impl HistoryItem {
             cwd: None,
             duration: None,
             exit_status: None,
+            duplicate: 0,
             more_info: None,
         }
     }
--- a/src/history/sqlite_backed.rs
+++ b/src/history/sqlite_backed.rs
@@ -38,6 +38,7 @@ fn deserialize_history_item(row: &rusqlite::Row) -> rusqlite::Result<HistoryItem
             .get::<&str, Option<i64>>("duration_ms")?
             .map(|e| Duration::from_millis(e as u64)),
         exit_status: row.get("exit_status")?,
+        duplicate: row.get("duplicate")?,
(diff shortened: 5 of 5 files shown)


fdncred:
if we know it's a duplicate, maybe we just shouldn't add it at all.

Kangaxx-0:
> if we know it's a duplicate, maybe we just shouldn't add it at all. if we do add this, i'd rather put it in `more_info` so we don't mess up the data storage by having a different schema for different versions of nushell.

To your first comment, the issue creator asked the same thing 

> What if instead of adding new column, we just add a mechanism to skip putting a new command that already exists?

I personally don't like this, the reason is - People intends to switch the configuration from `plaintext` to sql, probably because they prefer a database which gives them more information, flexibility, and also supports some fancy functionalities, if we skip duplicate commands, is it against the intention of using database?

To not mess up the schema, reedline should support a kind of migration for the forward compatibility, because as nu/reedline evolves, I believe the schema will require some changes sooner or later.

Kangaxx-0:
To move the duplication information to more_info is not a blocker for me, I just want to discuss and know how you guys want to manage the history schema going forward if you don't mind :)

fdncred:
I'm fine with other people jumping in with opinions about this and the schema. For me, I don't want to change the schema. The `more_info` was meant as a way to augment the schema without breaking the schema. However, I confess that I'm not sure exactly how to use it properly since I don't think it's a normal column but maybe a json column or some such thing.

sholderbach:
I am not sure tacking the change on the *write* addresses the concern as consistently as ensuring the deduplication on *read*: If you desire to filter the rich history according to a particular property you might inadvertently end up with a situation that duplicates reappear even though they are not consecutive in the global file. Global uniqueness will violate the perceived order if something like `ls` will only appear once early in you nushell history.

The "no duplicate" property is relevant for reading with filters either when doing regular up down history browsing, using the menu and lastly maybe when collecting data via the `history` command in nushell  (here a flag could switch between full history or just deduplicated sequence)

When @phiresky added the feature, some concern was given to allow users to collect statistics or to allow particular filtering based on for example the working directory. Collapsing the entries would loose information.

Kangaxx-0:
I am going to change my changes to discard the command when a duplication is detected.  But to be honest, if you tell me Nu shell uses a database, but it would not save the data when the command is detected as a duplicated one, it feels weird :)

fdncred:
How so other history dbs handle this like zsh-histdb or atuine?

Kangaxx-0:
@fdncred  @sholderbach A quick local change to prevent duplication being saved, how do you feel

Fix6502

fdncred:
Can you explain what's going on? I'm not following the gif. I mean, it doesn't appear that dupes are being added but how are you doing that and are you updating the duplicate command with the latest run time?

Kangaxx-0:
> Can you explain what's going on? I'm not following the gif. I mean, it doesn't appear that dupes are being added but how are you doing that and are you updating the duplicate command with the latest run time?

With the new change - As long as the entry command is detected as a dup, reedline will not add the entry to history.

fdncred:
What is the new change where it doesn't have the dupes column?

Kangaxx-0:
> What is the new change where it doesn't have the dupes column?

I am sorry, probably did not make this clear. I took your guys' suggestion and cut off a new branch(No new column, dupes will not be saved), the new GIF does nothing with this pr, I just want to send it over and ask the feedback :)

If this is what we want, I will create another PR and close off this one

fdncred:
ah, no worries.

1. I like it!
2. I'd like to see it update the timestamp column when you run a duplicate entry so that it has the latest time of when you last ran it. Maybe I'm wrong. What do you think?
3. I can see people wanting to be able to enable/disable this functionality so I'm thinking that it may need to have config option in nushell.

Kangaxx-0:
> 2. I'd like to see it update the timestamp column when you run a duplicate entry so that it has the latest time of when you last ran it. Maybe I'm wrong. What do you think?

Since we discard and skip the subsequent database operations, I would not to touch the pervioud record

> 3. I can see people wanting to be able to enable/disable this functionality so I'm thinking that it may need to have config option in nushell.

Thanks, this is a good catch! Let me publish my another pr, then I will address some trade-offs there :)

phiresky:
Hey, I'm the author of (most of the) SQLite history code. I'm looking at this because I was mentioned above. I didn't read all the discussion to see if this was mentioned before, but here's my opinion:

Tl;dr: adding a `duplicate` column is a hack that would cause issues in the future.

The sqlite history intentionally stores "duplicates" because they aren't actually duplicates (they will always differ in at least the timestamp, but probably also the directory, exit status, duration, ...).

It's not a good idea to store a `duplicate` flag in this database table, regardless of whether it's in more_info or a new column.

The meaning of such a column is ill-defined. Does it mean `duplicate` for the directory+commandline? does it mean duplicate for just the command line? A command such as `cargo run` has a different meaning depending on directory. If you filter by directory and only want unique commands then filtering by your `duplicate` column will break. Same as for any other filter combination explicitly encouraged with the extended database. So basically you'd need many columns like `duplicate_command_line_global`,`duplicate_command_line_directory`, `duplicate_command_line_exit_status`.

When querying with this column it's also inherently ambiguous what results you want. Do you want the command `ls` with the most recent timestamp run in the most recent directory? Or the oldest one? Or the last one with a successful exit status?

As a second point, it's also bad database design: This column would be a [denormalized](https://en.wikipedia.org/wiki/Denormalization) piece of information that's computable purely on information stored elsewhere in the database. It can become inconsistent with the actual contents. Consider e.g. future functionality of merging of two histories, or changing the command_line contents with some normalization (e.g. formatting), or making duplicates case-insensitives, ...

---


A better solution is to add a flag to the search function that excludes duplicates during querying using a GROUP BY command_line or similar. grouping by an indexed column should be O(log(n)) in the number of fetched history entries. [The same](https://github.com/larkery/zsh-histdb/blob/30797f0c50c31c8d8de32386970c5d480e5ab35d/sqlite-history.zsh#L439) is done in zsh-histdb to filter out uniques in the history view.

So my question would be what the goal is with this PR. Looking at #6502, the issue seems to be that using the `history` command dumps the whole history without any filtering, then any filters are only applied afterwards, leading to bad performance (idk why it should be O(n^2) though, it should still be O(log(n)) where n is the full history size?).

 So the "clean" solution there would be to allow passing down query functionality to the actual database. Maybe by adding a new command with arguments like `history-query --unique=command-line --limit=1000 --exit-status=0`. Or by adding a good customizable history menu into reedline itself (how does pagination even work otherwise?). Or maybe by using `open x.sqlite3 | query 'select * from history group by command_line'` command?

Kangaxx-0:
Thanks for the comments
> The meaning of such a column is ill-defined. Does it mean `duplicate` for the directory+commandline? does it mean duplicate for just the command line? A command such as `cargo run` has a different meaning depending on directory. If you filter by directory and only want unique commands then filtering by your `duplicate` column will break. Same as for any other filter combination explicitly encouraged with the extended database. So basically you'd need many columns like `duplicate_command_line_global`,`duplicate_command_line_directory`, `duplicate_command_line_exit_status`.

I totally agree with this, it is a not good naming which can cause confusion.

> As a second point, it's also bad database design: This column would be a [denormalized](https://en.wikipedia.org/wiki/Denormalization) piece of information that's computable purely on information stored elsewhere in the database. It can become inconsistent with the actual contents. Consider e.g. future functionality of merging of two histories, or changing the command_line contents with some normalization (e.g. formatting), or making duplicates case-insensitives, ...

I am not a guy from SQLite (did not find a bit column in official document), but adding bit column as true/false flag is not against the design principle, I do not know this is a normalization vs denormalization problem

Lastly, core team and the original issue creator member would like to discard dupes command before saving, I create another PR - https://github.com/nushell/reedline/pull/479

fdncred:
Thanks @phiresky for your input. Much appreciated as always. I'm kinda thinking we should close this in favor of maybe tweaking the nushell history command. Thoughts?

fdncred:
Closing for now. Thanks so much for the ideas and work. Let me know what you think in #479

PR #478, opened 2022-09-12
Revert resize handling removal

### LINKED ISSUE #477: nu doesn't handle resizing inteligently

sholderbach:
Attempt to test if bringing the basic resize behavior back will improve the situation in some terminals that implicitly resize on startup 

Tries to fix #477

- Revert "remove anims and resize repaint (#451)"
- Remove animations again

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -114,9 +114,6 @@ pub struct Reedline {
     hinter: Option<Box<dyn Hinter>>,
     hide_hints: bool,
 
-    // Is Some(n) read_line() should repaint prompt every `n` milliseconds
-    animate: bool,
-
     // Use ansi coloring or not
     use_ansi_coloring: bool,
 
@@ -171,7 +168,6 @@ impl Reedline {
             hinter,
             hide_hints: false,
             validator,
-            animate: false,
             use_ansi_coloring: true,
             menus: Vec::new(),
             buffer_editor: None,
@@ -251,14 +247,6 @@ impl Reedline {
     #[must_use]
     pub fn with_ansi_colors(mut self, use_ansi_coloring: bool) -> Self {
         self.use_ansi_coloring = use_ansi_coloring;
-        self
-    }
-
-    /// A builder which enables or disables animations/automatic repainting of prompt.
-    /// If `repaint` is true, every second the prompt will be repainted and the clock updates
-    #[must_use]
-    pub fn with_animation(mut self, repaint: bool) -> Self {
-        self.animate = repaint;
         self
     }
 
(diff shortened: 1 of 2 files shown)


fdncred:
i just check this out and built it and have the animated clock. I'm so confused.

sholderbach:
Sorry my bad, I got the logic mentally backwards.

dbuch:
I've tested this, and i solves the right prompt problem. I vote get it in, and progress from here.
There are some issues still that are annotated by todo/fix in code:

https://github.com/nushell/reedline/blob/main/src/painting/painter.rs#L383
https://github.com/nushell/reedline/blob/main/src/painting/painter.rs#L387

These probably needs attention :) I did try look into it. Logging/tracing would be useful I think @sholderbach, @fdncred would that be desired i could cook somthing up?

fdncred:
@dbuch oh, please do look at this. I'd love to land this but it's really trading one problem for another. however, if I knew you planned to work on the remaining issues I personally would feel much more comfortable landing this one with the hopes that you could cook something up to make it better. note, I have no fantasy that you can or would fix everything but we generally approve iterative approaches to solving "hard" issues like this. So, yes, please cook something up. :)

dbuch:
@fdncred Okay, I'll try. I still think this piece should be merged, do you know what the trade are?
This PR is after all - afaik - fairly okay now that amimation isn't restored again.

Next step, for me atleast, would be to implement some logging preferably to some file.

fdncred:
@dbuch I think if you run this PR and resize left and right and diagonally left and right, you'll see artifacts. Now if you do the same thing _before_ this PR, there are probably less artifacts with resizing. However the tradeoff with _before_ this PR is that sometimes you start nu and you get a prompt at the top of the screen, hit enter and all of a sudden the prompt is in the center of the screen.

dbuch:
@fdncred you are right! I use kitty, and because (i think) i use HDPI and kitty resizes late i get that prompt in the middle. But only at the beginning.
But the cause of this is likely because of invalid row calculation when shrinking down. I'll investigate abit more.

sholderbach:
A fresh set of eyes looking at this problem is absolutely welcome! @dbuch 

In my experimentation the resizing artifacts were especially pronounced when there was content close to the bottom of the screen + wrapping of text. Also the choice of terminal emulator had some influence on the severity of the visual artifacts. While `kitty` for example seems to hold of until sending the resize event, `alacritty` seemed to eagerly send events that were not properly handled by reedline.

Having optional tracing to study what is happening would be great, especially maybe with timing information to tweak how our event loop reacts to those things (https://github.com/nushell/reedline/blob/7d721a10e802f768889b0a6e4dd44f69f163ad9c/src/engine.rs#L40-L50, https://github.com/nushell/reedline/blob/7d721a10e802f768889b0a6e4dd44f69f163ad9c/src/engine.rs#L484-L499).
My memory is failing me what problem I was trying to observe but in one case I couldn't reproduce a certain behavior while launching reedline from `lldb`.

sholderbach:
PR has no direct connection to current problems or codebase. Closing despite some interesting discussion

PR #479, opened 2022-09-13
discard dup command

Kangaxx-0:
Reference link - https://github.com/nushell/reedline/pull/476, reedline will discard the command if the command name has been seen in the history table before

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1,10 +1,9 @@
 #[cfg(feature = "bashisms")]
+use crate::menu_functions::{parse_selection_char, ParseAction};
 use crate::{
-    history::SearchFilter,
-    menu_functions::{parse_selection_char, ParseAction},
+    result::{ReedlineError, ReedlineErrorVariants},
+    CommandLineSearch, SearchFilter,
 };
-
-use crate::result::{ReedlineError, ReedlineErrorVariants};
 use {
     crate::{
         completion::{Completer, DefaultCompleter},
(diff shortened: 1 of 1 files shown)


fdncred:
iirc it seems like it needed to be something like `with_animation()` so when it's initialized it can be with or defaulted without. `with_animation` was probably removed but it's something similar i think in the builder.

fdncred:
after reading phiresky's [comments](https://github.com/nushell/reedline/pull/476#issuecomment-1245480086), i tend to agree with him. i wonder if it would be better to augment the `history` command somehow.

Kangaxx-0:
Per phiresky's comment, having the ability to send the query to DB is a definitely better idea for sure, but the question is -  what do we want to support from nushell command param/flag,  doesn't `query db` gives user more flexibility :)

Please feel free to close my prs if it is not needed

fdncred:
I think I'm going to close these two. Thanks for your ideas and trying to close issues! Very much appreciated!!!

@Kangaxx-0 would you be interested in trying your hand at doing this via history params like `history --uniq --limit 1000 --exit_status 0` and the history command just queries the db?

Kangaxx-0:
> I think I'm going to close these two. Thanks for your ideas and trying to close issues! Very much appreciated!!!
> 
> @Kangaxx-0 would you be interested in trying your hand at doing this via history params like `history --uniq --limit 1000 --exit_status 0` and the history command just queries the db?

Sure, this is an interesting one for me too

Issue #438, opened 2022-06-05
Make chrono optional


sigmaSd:
Chrono is a relatively big dependency, at least making it optional would allow user to choose to remove it 

cargo bloat: chrono alone is  adding 1.1% to text section
```
File  .text    Size    Crate Name
0.1%   1.2% 19.4KiB      std addr2line::ResDwarf<R>::parse
0.1%   1.1% 18.4KiB      std std::backtrace_rs::symbolize::gimli::resolve::{{closure}}
0.1%   0.7% 11.7KiB reedline reedline::engine::Reedline::handle_editor_event
0.1%   0.6% 10.7KiB reedline reedline::main
0.0%   0.6%  9.5KiB      std addr2line::ResUnit<R>::parse_lines
0.0%   0.6%  9.4KiB   chrono <chrono::format::strftime::StrftimeItems as core::iter::traits::iterator::Iterator>::next
0.0%   0.5%  8.5KiB      std miniz_oxide::inflate::core::decompress
0.0%   0.5%  7.6KiB   chrono chrono::format::format_inner
```

here is an example implementation https://github.com/nushell/reedline/compare/main...sigmaSd:opt_chrono?expand=1

comparison with this pr
```
# with chrono
cargo tree | wc -l
> 88
```
```
# without chrono
 cargo tree --no-default-features | wc -l
> 77
```

sholderbach:
Do you happen to know an alternative light weight crate for that task? In this case it is just needed for the basic timestamp in the example Prompt and now in the History with metadata (#401)

phiresky:
that can probably also be done with `std::time::SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)`

sholderbach:
For the time-stamping in the db that should be fine (if timezone or strict ordering concerns are not an issue). I don't recall the std library directly providing the formatting machinery for it to be useful in the example prompt (but yeah making a dependency for that a disable-able default feature could be an option for the dependency conscious)

PR #481, opened 2022-09-16
change history session_id to a systematically generated number

fdncred:
This PR changes the `history_session_id` from a regular number like 1,2,3,4,5 to a systematically generated number based on the date of reedline's first commit and is expressed in nanoseconds. 

I submit this change in order to be able to generate a `history_session_id` without the need of looking anything up in the history database. This number should be pretty unique although not precisely unique. Each time you start up reedline you get a new time and therefore you get a new `history_session_id`.

I also changed the trait so that `next_session_id()` no longer exists.

I created a public function in the Reedline impl to `get_history_session_id()`. Hopefully that is enough to return it so that we can use it in nushell in order to filter on each session id which will be unique to each terminal window running reedline/nushell.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -195,9 +195,9 @@ impl Reedline {
         let nanoseconds_since_first_commit =
             duration_since_first_commit.num_nanoseconds().unwrap_or(0);
 
-        eprintln!(
-            "nanoseconds_since_first_commit: {}",
-            nanoseconds_since_first_commit
+        std::env::set_var(
+            "REEDLINE_SESSION_ID",
+            nanoseconds_since_first_commit.to_string(),
         );
 
         Some(HistorySessionId::new(nanoseconds_since_first_commit))
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -156,7 +156,7 @@ impl Reedline {
         let hinter = None;
         let validator = None;
         let edit_mode = Box::new(Emacs::default());
-        let hist_session_id = Self::get_history_session_id();
+        let hist_session_id = Self::create_history_session_id();
 
         Reedline {
             editor: Editor::default(),
@@ -185,7 +185,7 @@ impl Reedline {
     }
 
     /// Get a new history session id based on the current time and the first commit datetime of reedline
-    fn get_history_session_id() -> Option<HistorySessionId> {
+    fn create_history_session_id() -> Option<HistorySessionId> {
         //Sun Feb 28 22:42:07 2021 +1300
         let first_commit_date_time_naive = NaiveDate::from_ymd(2021, 2, 28).and_hms(22, 42, 7);
         let first_commit_date_time_utc =
@@ -195,12 +195,12 @@ impl Reedline {
         let nanoseconds_since_first_commit =
             duration_since_first_commit.num_nanoseconds().unwrap_or(0);
 
-        std::env::set_var(
-            "REEDLINE_SESSION_ID",
-            nanoseconds_since_first_commit.to_string(),
-        );
-
         Some(HistorySessionId::new(nanoseconds_since_first_commit))
+    }
+
+    /// Return the previously generated history session id
+    pub fn get_history_session_id(&self) -> Option<HistorySessionId> {
+        self.history_session_id
     }
 
     /// A builder to include a [`Hinter`] in your instance of the Reedline engine
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -3,8 +3,6 @@ use crate::{
     history::SearchFilter,
     menu_functions::{parse_selection_char, ParseAction},
 };
-
-use crate::result::{ReedlineError, ReedlineErrorVariants};
 #[cfg(feature = "external_printer")]
 use {
     crate::external_printer::ExternalPrinter,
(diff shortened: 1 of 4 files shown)


src/engine.rs
```diff
@@ -182,6 +184,25 @@ impl Reedline {
         }
     }
 
+    /// Get a new history session id based on the current time and the first commit datetime of reedline
+    fn create_history_session_id() -> Option<HistorySessionId> {
+        //Sun Feb 28 22:42:07 2021 +1300
+        let first_commit_date_time_naive = NaiveDate::from_ymd(2021, 2, 28).and_hms(22, 42, 7);
+        let first_commit_date_time_utc =
+            DateTime::<Utc>::from_utc(first_commit_date_time_naive, Utc);
```
dandavison, line 192: What's the purpose of using the first commit of reedline as the baseline rather than using the standard unix timestamp (i.e. time since Jan 1 1970)?

fdncred: Hey Dan, The purpose was that I wasn't sure exactly if nanoseconds since UNIX_EPOCH would always (in the near future) fit into a i64. So, while I was thinking about that, I just said to myself, I really don't care that it's some universal time thing like unix epoch, I just care that the number is pretty unique and won't overrun an i64 any time soon. So, I chose the first commit date and time as the constant to compare against.

fdncred: I wanted to use a [ulid](https://github.com/ulid/spec) but it's alphanumeric and I didn't want to change the datatype in the sqlite db.

dandavison: OK, it looks like a u64 fits 585 years of nanoseconds, and an i64 292 years, so I'd say go for the standard UNIX timestamp.

```
In [12]: ns_per_year = 365 * 24 * 60 * 60 * 1e9

In [13]: (2 ** 64 - 1) / ns_per_year
Out[13]: 584.942417355072

In [14]: (2 ** 63 - 1) / ns_per_year
Out[14]: 292.471208677536
```

fdncred: I wonder why I can't do this math in nutshell. 🤔

fdncred: Anyway, I changed it to unix_epoch.

dandavison: Great, yes I think it makes sense to keep it simple and unsurprising -- users are sure to look at these things from time to time.

sholderbach:
I don't have much expertise to bring in regards to the practical database integrity concerns, but avoiding the lookup sounds good. At the moment a nanosecond timescale might still be unlikely to cause an issue with e.g. tmux launching a bunch of sessions at the same time.

If we are changing the values in the database anyways, we might consider the complaint brought by #438 and rely on the std library unix time (don't know about whether we have to accomodate timezone concerns here) and defer chrono types to the places that calculate and display.

fdncred:
If we can get by without chrono, then I think we should. However, that should probably be another PR.

> I don't have much expertise  ...

I just wanted to give you the opportunity to object since I changed the trait.  I haven't changed nushell at all yet. I just wanted to make sure (1) sql history storage works (2) we can access `get_history_session_id()` from nushell so we can create a command with it, or reference it somehow. I haven't tried item 2 yet.

phiresky:
I think it's a good idea to use a generated unique session ID instead of a sequential one, especially since it's probably required to allow creating a session id without modifying our database schema a lot otherwise (adding a separate sessions table).

I think the reason why I didn't do so in the first place is that such an ID would usually be a text string and thus use much more storage than small integers. But that consideration is probably not too important, considering almost everyone seems to have a history with < 1M entries.

Regarding this PR, I think it might be smart to use a stringy id that has a better guarantee to actually be globally unique. There's a lot of research and discussion ([e.g. blog post](https://supabase.com/blog/choosing-a-postgres-primary-key)) in this area because just being unique is not the only consideration. UUID, [nanoid](https://github.com/ai/nanoid#comparison-with-uuid) and [ULID](https://github.com/ulid/spec) and [XID](https://github.com/rs/xid) come to mind.

Pros of using e.g. nanoid over your PR code:

- No [Year 2262 problem](https://en.wikipedia.org/wiki/Time_formatting_and_storage_bugs#Year_2262)
- "Battle-tested" - using timestamps as a unique id is (in general) frowned upon I'd say, most people use UUIDs or similar.
- Possibility of collisions might be higher than expected e.g. if something starts two shells at the same time and you're on an OS that doesn't actually use nano-second accuracy for SystemTime (e.g. Windows is only 100ns-steps, macOS only 1000ns-steps, see [the docs](https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html#platform-specific-behavior)).
- Using a timestamp might lead to people assuming the id is always a timestamp even though it's not intended to be (basically ossification, making potential later changes harder)

Pros of using epoch nanoseconds as a 64bit signed int:

- No schema change needed (column stays integer)
- More storage efficient than text ids

sholderbach:
Thanks for the valuable feedback @phiresky. The point you raise about the clock precision certainly puts collisions in the realm of reasonable possibilities, especially given that side-channel mitigations might push people/OSes to degrade that further. ULID or xid with the ability to lexically sort them sounds kind of interesting for our use, but yeah we would need a way to translate to a new schema.

fdncred:
let's move forward with this and if we need a ulid later we can figure out how to make a migration strategy.

PR #485, opened 2022-09-19
enable easier history session id conversion

fdncred:
This PR adds a small change to enable easier conversion from HistorySessionId to i64.


sholderbach:
Sounds reasonable

PR #488, opened 2022-09-29
Remove flicker on external print

tailhook:
Just tested external print and noticed this issue. It's not super important, but nice to have.

Previously each external print was causing a visible flicker on at least two terminal emulators I've tested (alacritty, xfce4-terminal).

I'm not sure this is the best solution. I've looked into just removing `.flush()` in the `print_line`. The latter is used in just two places here and in printing history. It would be fine to also optimize out flush on every line when printing history, however that would likely break abstraction and make painter somewhat more stateful (i.e. needs exposing `.painter.stdout.flush()` in a some way).

Current approach also makes painter a bit stateful and works just because `self.paint()` is called right after `print_external_message` (https://github.com/nushell/reedline/blob/f994cc9/src/engine.rs#L492-L497), but at least scope of that dependency is a local enough. Feel free to suggest a better way.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -454,6 +454,9 @@ impl Painter {
     }
 
     /// Prints an external message
+    ///
+    /// This function doesn't flush the buffer. So buffer should be flushed
+    /// afterwards perhaps by repainting the prompt via `repaint_buffer()`.
     pub fn print_external_message(
         &mut self,
         messages: Vec<String>,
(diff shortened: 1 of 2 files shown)


src/painting/painter.rs
```diff
@@ -454,6 +454,9 @@ impl Painter {
     }
 
     /// Prints an external message
+    ///
+    /// This function doesn't flush the buffer. So buffer should be flushed
+    /// afterwards perhaps by repainting the prompt via `repaint_buffer()`.
     pub fn print_external_message(
```
sholderbach, line 460: This should only be called in the context you describe so let's hide it
```suggestion
    pub(crate) fn print_external_message(
```

sholderbach (approved):
Good call! Awesome that you are giving it a test drive!

PR #490, opened 2022-10-07
Add `Submit` and `SubmitOrNewline` editor events

tailhook:
* `Submit` is used to return data unconditionally. This is useful for example if you want compiler to show you where the unclosed paren that doesn't pass validation actually is.
* `SubmitOrNewline` is similar to `Enter` but inserts a new line if the cursor is not at the end of the buffer. It's useful to insert newlines in the middle of the text when editing.

This also removes unconditional `KeyCode::Enter` to `ReedlineEvent::Enter` event mapping and sets it in the keybindings normally (so you can rebind to `Submit` or `SubmitOrNewline`)

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -22,6 +22,9 @@ pub fn default_emacs_keybindings() -> Keybindings {
     add_common_navigation_bindings(&mut kb);
     add_common_edit_bindings(&mut kb);
 
+    // This could be in common, but in Vi it also changes the mode
+    kb.add_binding(KM::NONE, KC::Enter, ReedlineEvent::Enter);
+
     // *** CTRL ***
     // Moves
     kb.add_binding(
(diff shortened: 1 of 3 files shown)


sholderbach:
That sounds like a good idea to decouple some of that hard coded behavior on enter. I will try to take a look if there is stuff that might break. Having the option to explicitly bind it to a separate binding sounds definitely nice, weighing the options in my head of whether we should have all three or a config between  the `Enter` and `SubmitOrNewline` behavior

tailhook:
Yes, `Enter` and `SubmitOrNewline` usually may be configurable. In `Rustyline` it's a bool [field in enum variant](https://docs.rs/rustyline/latest/rustyline/enum.Cmd.html#variant.AcceptOrInsertLine). On the other hand, I don't know if it makes sense to make this variable somewhere in the `EditMode`, as that would require a special method to configure that (so one more thing to learn), while changing binding sounds obvious and clear to me.

`Submit` on the other hand has to be a separate, as we, for example, bind `Submit` on `Alt+Return` and `SubmitOrNewline` on `Return`.

sholderbach:
Seems to work fine in my testing! Thanks for the elegant solution!

sholderbach:
I will land it as is to allow users to play with it in the next (nushell) release. If we want to make a later breaking change we to the bindings we may bunch it up with other likely changes.

PR #492, opened 2022-10-15
render right prompt on the last line of the left prompt

nibon7:
This PR makes nushell+starship behave like other shells (e.g. zsh).
Related: starship/starship#3982 
Closes nushell/nushell#4909

* nushell native prompt
Screenshot from 2022-10-16 00-06-28
* nushell + starship(disable `line_break` and `add_newline`)
Screenshot from 2022-10-16 00-07-09
* nushell + starship(enable `line_break` and disable `add_newline`)
Screenshot from 2022-10-16 00-07-51
* nushell + starship(enable `line_break` and `add_newline`)
Screenshot from 2022-10-16 00-08-34

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -183,16 +183,19 @@ impl Painter {
             .screen_width()
             .saturating_sub(prompt_length_right as u16);
         let input_width = lines.estimate_first_input_line_width();
-        let screen_width = self.screen_width();
-        let required_lines = lines.required_lines(screen_width, None);
+
+        let mut row = self.prompt_start_row;
+        if lines.right_prompt_on_last_line {
+            let screen_width = self.screen_width();
+            let required_lines = lines.required_lines(screen_width, None);
+
+            row += required_lines.saturating_sub(1);
+        }
 
         if input_width <= start_position {
             self.stdout
                 .queue(SavePosition)?
-                .queue(cursor::MoveTo(
-                    start_position,
-                    self.prompt_start_row + required_lines.saturating_sub(1),
-                ))?
+                .queue(cursor::MoveTo(start_position, row))?
                 .queue(Print(&coerce_crlf(&lines.prompt_str_right)))?
                 .queue(RestorePosition)?;
         }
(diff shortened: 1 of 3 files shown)


fdncred:
to me, it's a bit weird to have the left prompt and the right prompt on different lines. My personal preference would be like the middle one in #4909. Is this option available with your change?

```
starship on  master [$] is 📦 v1.4.2 via 🦀 v1.59.0                             10/16/2022 12:08:15 AM
❯ 
```

nibon7:
> Is this option available with your change?
Nope. This PR always render the right prompt on the last row. Shall we provide an option to render it on the first or the last row?

nibon7:
I have a base implementation to control whether we render the right prompt on the first or last line by setting an environment variable. @fdncred 

Screenshot from 2022-10-16 21-11-23

fdncred:
it's a beautiful things @nibon7! thanks for considering my request!

sholderbach:
I don't have strong opinions on this one. Will the right prompt still be hidden when the first line is full? Do we need to make this an option?

fdncred:
I do think this should be some configurable option in nushell vs an env var maybe?

nibon7:
> Will the right prompt still be hidden when the first line is full?

In fact, we just calculate the width of the last row, so if there are multiple rows and the last row is not full, it won't be hidden.

https://github.com/nushell/reedline/blob/5061d2288c7ab8373700253ad56127fb8b0b5f73/src/painting/prompt_lines.rs#L95-L118

e.g.  `(term size).columns is 79`, the following code still shows the right prompt.
```nu
let-env PROMPT_COMMAND = { "LLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLONG_LINE_EATEN\n<PROMPT_INDICATOR>" }
let-env PROMPT_COMMAND_RIGHT = { "RIGHT_PROMPT" }
```

nibon7:
> I do think this should be some configurable option in nushell vs an env var maybe?

Agreed, I did this just for a quick test.

fdncred:
@sholderbach I think we're ready to land this if you are, as well as https://github.com/nushell/nushell/pull/6781 when ready.

sholderbach:
Let's get this in, I haven't fully validated the last line behavior but that can probably best evaluated with the nushell side with the different prompt configurations.

PR #493, opened 2022-10-15
examples updated

perlindgren:
This PR covers all examples except "custom keybindings", however in the "completions" example a custom keybinding is exemplified.

Let me know if you think it is worth merging in current state or if some further changes are needed.

/Per


sholderbach:
Yeah having the separate examples to run is probably a great way to experiment with the individual features in isolation. Thanks. I will land that first, that is a good starting point to polish the API and fix the documentation

PR #498, opened 2022-10-18
Tab inline completion

dandavison:
This PR makes nushell tab completion behave more like zsh/bash. Assuming `quick_completions = true`:

- Changes Tab so that it completes instead of selecting the next menu item (arrow keys can be used for that)
- Completion is "as far as possible". So, the entire word if there's a unique completion, or else up to the shared prefix of possible completions.
- Gets rid of the quick completion behavior whereby a word was completed automatically as soon as the input came to have a unique completion.

See https://github.com/nushell/nushell/pull/6802 to use this with nushell.


fdncred:
I'm guessing we need to land this one so we don't have to point to @dandavison's repo?

sholderbach:
Sounds good!

PR #500, opened 2022-10-22
Custom validator and prompt code examples

perlindgren:
I made a small example how to implement a custom validator and a custom prompt.

I'm not sure if this is the right way to do it, but the example works as I anticipated.
One option is to merge as is, or rename it to `validator_prompt` (is it kind of covers both), or alternatively split them in two (as they are kind of independent).

Let me know what you think is best way forward.

### DIFF
--- a/examples/custom_prompt.rs
+++ b/examples/custom_prompt.rs
@@ -1,0 +1,75 @@
+// Create a reedline object with a custom prompt.
+// cargo run --example custom_prompt
+//
+// Pressing keys will increase the right prompt value
+
+use reedline::{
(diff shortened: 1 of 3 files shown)


sholderbach:
Would you mind fixing the lints from `cargo clippy --examples`?

perlindgren:
Will do...

perlindgren:
I think this should do.

sholderbach:
Sorry for letting you hang out to dry on this one. Decided to split them up into two as there was no direct interaction. Thanks for the examples, this should really help newcomers and implementers while experimenting!

PR #501, opened 2022-10-22
show right prompt on indicator line in last line mode

nibon7:
When enable last line mode, the right prompt may be rendered incorrectly if we have multiline input or hint.
This PR makes right prompt show on the indicator line when last line mode enabled.

* before
Screenshot from 2022-10-23 06-51-07
Screenshot from 2022-10-23 06-50-57

* after
Screenshot from 2022-10-23 06-54-45
Screenshot from 2022-10-23 06-54-59

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -196,8 +196,7 @@ impl Painter {
 
         let mut row = self.prompt_start_row;
         if lines.right_prompt_on_last_line {
-            let required_lines = lines.required_lines(screen_width, None);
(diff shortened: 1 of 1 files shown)


sholderbach:
Yeah that might be a bit less visually jarring if the right prompt doesn't flicker around. What will happen if the hinted line is long enough to reach the inserted right prompt?

nibon7:
> What will happen if the hinted line is long enough to reach the inserted right prompt?

the right prompt will be hidden

sholderbach:
Great!

PR #503, opened 2022-10-30
Update workflow to new actions

sholderbach:
Still staying with `actions-rs` instead of `actions-rust-lang` for setup. (nushell uses them for setup with integrated caching)
Remove build stage as this should also be covered by test.
clippy --all to see if this checks the examples as well.


sholderbach:
Closing as #505 does the same thing more compactly.

PR #504, opened 2022-10-30
fix: actually use sqlite-dynlib feature

jcgruenhage:
Follow-up on https://github.com/nushell/reedline/pull/474

7d721a1 introduced the sqlite-dynlib feature, but using it in places
where sqlite is currently used was not possible, because the cfg gates
weren't adjusted yet. This commit replaces each `feature = "sqlite"`
with `any(feature = "sqlite",feature = "sqlite-dynlib")`, meaning that
it doesn't matter which sqlite feature you use, reedline will expose the
same API.

### DIFF
--- a/src/history/base.rs
+++ b/src/history/base.rs
@@ -173,9 +173,9 @@ pub trait History: Send {
 
 #[cfg(test)]
 mod test {
-    #[cfg(feature = "sqlite")]
+    #[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
     const IS_FILE_BASED: bool = false;
-    #[cfg(not(feature = "sqlite"))]
+    #[cfg(not(any(feature = "sqlite", feature = "sqlite-dynlib")))]
     const IS_FILE_BASED: bool = true;
 
     use crate::HistorySessionId;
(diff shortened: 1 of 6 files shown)


sholderbach:
Right I totally forgot the impact in the code. Sad that the dynamic linking is a bit hard to test with the default actions runner.

Thanks for the fix!

jcgruenhage:
No worries, I also didn't think about it back then, just stumbled over it now, when trying to build our patched nushell package:

```
error[E0432]: unresolved import `reedline::SqliteBackedHistory`
  --> crates/nu-cli/src/repl.rs:21:51
   |
21 | use reedline::{DefaultHinter, EditCommand, Emacs, SqliteBackedHistory, Vi};
   |                                                   ^^^^^^^^^^^^^^^^^^^
   |                                                   |
   |                                                   no `SqliteBackedHistory` in the root
   |                                                   help: a similar name exists in the module: `FileBackedHistory`

error[E0432]: unresolved import `reedline::SqliteBackedHistory`
 --> crates/nu-command/src/misc/history.rs:9:5
  |
9 |     SqliteBackedHistory,
  |     ^^^^^^^^^^^^^^^^^^^
  |     |
  |     no `SqliteBackedHistory` in the root
  |     help: a similar name exists in the module: `FileBackedHistory`
```

It looks to me that this works as expected now, although I haven't quite finished up the packaging of 0.70.0 for void yet

sholderbach:
Thank you for putting in the hard work to deal with the packaging! Great to have nushell available in more distributions.

PR #505, opened 2022-10-30
Update workflow to new actions (alternative)

sholderbach:
Alternative to #503

Using `actions-rust-lang` instead of `actions-rs` for setup. (nushell uses them for setup with integrated caching)

Remove build stage as this should also be covered by test.
clippy --all to see if this checks the examples as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,48 +31,35 @@ jobs:
     runs-on: ${{ matrix.platform }}
 
     steps:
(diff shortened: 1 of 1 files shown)


sholderbach:
Difference in runtime from an empty cache to #503 is 3 mins vs 4 mins. But this one being 1 min faster might just be an artifact of the machine this job was running on. Choosing this one as the workflow file is simpler and more compact.

Issue #516, opened 2022-11-23
Make insertion to the history optional (e.g. with leading space)


sholderbach:
Currently all submitted entries are directly appended to the history. (With the sqlite history we support amending the history entry afterwards based on an ID that is not part of the text based history)
This might not always be desired (e.g. dangerous operation, something containing a secret key/password).
Some shells support ignoring all lines that start with a space.

Goal: allow us to make history insertion conditional/optional

Requirements: 
- Needs to also work with the simple text based history.
- hardcoding space at the beginning might be bad in some context (other programming languages)

## References

bash/zsh support it

rgwood:
I can take this on. This is easy if we hardcode the functionality, I just need to figure out the best way to make it configurable.

One way would be to use Bash's HISTCONTROL and/or HISTIGNORE environment variables: 
- https://github.com/nushell/nushell/issues/6108
- https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-HISTCONTROL

I don't _love_ that approach, it gets into unpleasant "exactly how Bash-compatible does it need to be?" questions.

Another option: maybe we add a `history_ignore_prefix: Option<char>` to the Reedline struct? `None` by default, but users can configure it to be `Some(' ')` if they want.

PR #518, opened 2022-11-24
Add history exclusion configuration

rgwood:
A rough draft PR to solve #516, by letting users configure which command lines get excluded from history. For example, users can configure reedline so that command lines starting with a space are not saved.

@sholderbach Let me know what you think, happy to revise this.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -60,6 +60,7 @@ fn main() -> Result<()> {
 
     let mut line_editor = Reedline::create()
         .with_history(history)
+        .with_history_exclusion_prefix(Some(" ".to_string()))
         .with_completer(completer)
         .with_quick_completions(true)
         .with_partial_completions(true)
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1523,7 +1523,7 @@ impl Reedline {
         // Additional repaint to show the content without hints etc.
         self.repaint(prompt)?;
 
-        if !buffer.is_empty()  {
+        if !buffer.is_empty() {
             // Support the Bash HISTCONTROL thing of skipping history saving if command starts with a certain string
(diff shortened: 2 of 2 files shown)


sholderbach:
Doing the string match internally is certainly the minimally invasive change. How will this interact with the data amending as done for the sqlite history?

rgwood:
@sholderbach I don't think there's any special handling for SQLite history needed, because all the logic is happening in engine.rs. Perhaps I've missed something?

image

sholderbach:
Starting the session without a normal command being run crashes, as soon as I try to add an entry that is skipped (with `--features sqlite`):
image

I think it tries to modify the last entry (and passes if it can mess with the last succesfully stored one)

rgwood:
Oops! Will take a look and try to add some tests, thanks.

rgwood:
Hm. Y'know, I'm not sure whether that's a bug. A case could be made that the demo is intentionally not handling that edge case well:

https://github.com/nushell/reedline/blob/076ce6f7fca5f3e1f63bbad9b9b940ac1463bd2a/examples/demo.rs#L130

sholderbach:
I think this rather indicates even with ignoring error handling that https://docs.rs/reedline/latest/reedline/struct.Reedline.html#method.update_last_command_context would try to modify the last remembered entry.

rgwood:
Right. I don't see the problem but I'm probably missing something.

rgwood:
current status: I'm not really sure what the best way to implement this is. Might need to start storing the last command in memory instead of always looking it up in the DB. I've run out of steam to work on this for now but might come back to it another time; if anyone else wants to pick it up feel free.

sholderbach:
Closed by #566

Issue #41, opened 2021-04-10
[ENH] Prompt segment API


sholderbach:
Define a `PromptSegment` trait that specifies how we interact with a prompt segment which can display information about the environment (e.g. working directory, time, last return code, git branch...). A Vec of trait objects could be used to allow custom configs.

In the draw call we could print prompt segments in different possible locations: Separate line above/below, same line, left/right

### Open questions:

Who is responsible for shrinking presented information if the window width is to low to print all segments?

PR #519, opened 2022-11-30
Make DefaultPrompt configurable

jmoore34:
I was working on a small REPL application, and I found that the default prompt was a bit overkill (and more suited for a shell).

So, I made the `DefaultPrompt` struct more configurable, so you can do things like this:

```rust
    let prompt = DefaultPrompt {
        left_prompt: DefaultPromptSegment::Basic("rand-api".to_owned()),
        right_prompt: DefaultPromptSegment::Empty
    };
    // or
    let prompt = DefaultPrompt::new(
        DefaultPromptSegment::Basic("rand-api".to_owned()),
        DefaultPromptSegment::Empty
    );
```

with a result like this:
WindowsTerminal_qLrrmRTYx1

I modified `DefaultPrompt::default()` to produce the same behavior as before. So, users who were using `DefaultPrompt::default()` (as per the example in the readme) should in theory see no breaking changes.

I.e., the following

```rust
let prompt = DefaultPrompt::default()
```

will still produce the following:

WindowsTerminal_UEBcbXOWfr

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -102,7 +102,7 @@ fn main() -> Result<()> {
     // Adding vi as text editor
     line_editor = line_editor.with_buffer_editor("vi".into(), "nu".into());
 
-    let prompt = DefaultPrompt::new();
+    let prompt = DefaultPrompt::default();
 
     loop {
         let sig = line_editor.read_line(&prompt);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -253,8 +253,8 @@ pub use history::{
 
 mod prompt;
 pub use prompt::{
-    DefaultPrompt, Prompt, PromptEditMode, PromptHistorySearch, PromptHistorySearchStatus,
-    PromptViMode, DefaultPromptSegment
+    DefaultPrompt, DefaultPromptSegment, Prompt, PromptEditMode, PromptHistorySearch,
+    PromptHistorySearchStatus, PromptViMode,
 };
 
 mod edit_mode;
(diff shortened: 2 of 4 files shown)


jmoore34:
If requested I can change the
```rust
Basic(String)
```
variant of `DefaultPromptSegment` to 
```rust
Basic(&'a str)
```
(I mainly used String to avoid putting `'a` everywhere)

kubouch:
I think it could be useful for people who utilize our crates to build their own CLIs. Just curious, would setting the `PROMPT_COMMAND` environment variables work for your use case?

jmoore34:
Er, I'm not sure what that would look like

kubouch:
Since your application must be using the Nushell runtime, you should be able to set environment variables, that's how you change the prompt in Nushell.

But anyway, I think this would help people building their own REPLs with Nushell as it makes the API cleaner and easier to add more types of segments. Thanks!

kubouch:
(Ignore my comment, I was thinking I was in the Nushell repo 😆)

sholderbach:
Thanks for giving that a shot! I love that you work towards making `reedline` also usable for users beyond nushell. Kind of dig the Idea of the prompt segments as well (opened #41 before we had sketched out how nushell should behave). Doesn't necessarily make sense for nushell but is a great quality of life improvement for other users that want to get things going quickly without having to reimplement everything. The only thing I am wondering about is if we should make this whole `DefaultPrompt` optional behind a feature gate or have more than one `impl Prompt` (`DefaultPrompt` feels kind of weird as a name).

jmoore34:
Personally I think the code is small enough that a feature gate would be overkill. (Not to mention I imagine plenty of applications would rather use a default prompt instead of fully implementing Prompt themselves.)

I think DefaultPrompt is a suitable name, but some alternatives might be SimplePrompt or BasicPrompt.

PR #531, opened 2023-01-13
Allow configuration of multiline prompt color

dgkf:
The color of the multiline prompt is currently hard-coded to be `LightBlue`. This ended up being a pretty straightforward change to allow it to be configured using the `Prompt` trait similar to the other prompt colors.

One oddity in here is that the multiline prompt uses a `nu_ansi_term::Color`, whereas the other colors use a `crossterm::style::Color`. I'm not sure which one is preferred at this point, but I'd be willing to work toward some consistency if there's a current direction toward one or the other.

### DIFF
--- a/src/painting/styled_text.rs
+++ b/src/painting/styled_text.rs
@@ -1,5 +1,10 @@
+use std::borrow::Borrow;
+
+use nu_ansi_term::Style;
+
+use crate::Prompt;
+
 use super::utils::strip_ansi;
-use nu_ansi_term::{Color, Style};
 
 /// A representation of a buffer with styling, used for doing syntax highlighting
 pub struct StyledText {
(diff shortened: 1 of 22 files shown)


src/painting/styled_text.rs
```diff
@@ -34,18 +37,22 @@ impl StyledText {
...
+        // multiline_prompt: &str,
         use_ansi_coloring: bool,
     ) -> (String, String) {
         let mut current_idx = 0;
         let mut left_string = String::new();
         let mut right_string = String::new();
-        let prompt_style = Style::new().fg(Color::LightBlue);
+
+        let multiline_prompt = prompt.render_prompt_multiline_indicator();
+        let prompt_style = Style::new().fg(prompt.get_prompt_multiline_color());
```
dgkf, line 49: This is the heart of the change. Instead of fixing this to `Color::LightBlue`, it is instead configured in `Prompt::get_prompt_multiline_color()`.

src/prompt/base.rs
```diff
@@ -105,6 +106,10 @@ pub trait Prompt: Send {
     fn get_prompt_color(&self) -> Color {
         DEFAULT_PROMPT_COLOR
     }
+    /// Get the default right prompt color
+    fn get_prompt_multiline_color(&self) -> nu_ansi_term::Color {
+        DEFAULT_PROMPT_MULTILINE_COLOR
+    }
```
dgkf, line 112: Added a new function to the `Prompt` to allow for it fetch a color. For now it uses a `nu_ansi_term::Color` unlike the others.

sholderbach (approved):
Thanks that is very reasonable!

PR #534, opened 2023-01-31
Fix columnar menu completion corrupting the editor insertion point

csko:
See: https://github.com/nushell/nushell/issues/7885 for more context.

Repro:
1. type `` open FILENAME` ``
2. press TAB (`` ` `` gets removed at the end)
3. press `` ` ``

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -667,8 +667,8 @@ impl Menu for ColumnarMenu {
             line_buffer.replace_range(start..end, &value);
 
             let mut offset = line_buffer.insertion_point();
-            offset += value.len();
-            offset -= end.saturating_sub(start);
+            offset = offset.saturating_add(value.len());
+            offset = offset.saturating_sub(end.saturating_sub(start));
             line_buffer.set_insertion_point(offset);
             editor.set_line_buffer(line_buffer, UndoBehavior::CreateUndoPoint);
         }


csko:
Should we bump the version?

src/menu/columnar_menu.rs
```diff
@@ -667,7 +667,8 @@ impl Menu for ColumnarMenu {
             line_buffer.replace_range(start..end, &value);
 
             let mut offset = line_buffer.insertion_point();
-            offset += value.len().saturating_sub(end.saturating_sub(start));
+            offset += value.len();
+            offset -= end.saturating_sub(start);
```
sholderbach, line 671: Can we prove that this never underflows? (Not sure if we have enough tests with gnarly insertion points in compleitions in the `reedline` repo for that)

src/menu/columnar_menu.rs
```diff
@@ -810,4 +811,25 @@ mod tests {
             append_whitespace: false,
         }
     }
+
+    #[test]
+    fn test_menu_replace_backtick() {
+        // https://github.com/nushell/nushell/issues/7885
```
sholderbach, line 817: Thanks for the practical test!

sholderbach (commented):
Thanks for tackling this problem! 

The menu system at the moment is still overly complex in my view. Thanks for spotting that bug!

src/menu/columnar_menu.rs
csko: good point, I've changed these to saturating_sub/add

sholderbach (approved):
Let's give this a go

PR #536, opened 2023-02-08
Add History::clear() function

### LINKED ISSUE #535: Add an API to clear history entirely
Nushell currently does some hacky file deletion stuff to clear history (via `history --clear`), and it doesn't work well for SQLite history.

Let's add a function to the `History` trait that deletes all history; the SQLite implementation can run some `DELETE` or `TRUNCATE` SQL.

I'll pick this up when I have time.

rgwood:
Adds a `History::clear()` function that deletes all history items. Closes #535.

Will eventually close https://github.com/nushell/nushell/issues/7991

### DIFF
--- a/src/history/base.rs
+++ b/src/history/base.rs
@@ -334,4 +334,42 @@ mod test {
 
         Ok(())
     }
+
+    // test that clear() works as expected across multiple instances of History
+    #[test]
+    fn clear_history_with_backing_file() -> Result<()> {
+        #[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
+        fn open_history() -> Box<dyn History> {
+            Box::new(
+                crate::SqliteBackedHistory::with_file("target/test-history.db".into()).unwrap(),
+            )
+        }
+
+        #[cfg(not(any(feature = "sqlite", feature = "sqlite-dynlib")))]
+        fn open_history() -> Box<dyn History> {
+            Box::new(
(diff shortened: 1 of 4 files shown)


src/history/base.rs
```diff
@@ -322,4 +324,14 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn clear_history() -> Result<()> {
+        let mut history = create_filled_example_history()?;
+        assert_ne!(history.count_all()?, 0);
+        history.clear().unwrap();
+        assert_eq!(history.count_all()?, 0);
```
sholderbach, line 333: Maybe we should drop it and try to reopen it, to exclude a potential error case like we were seeing with nushell's sqlite "clear"

sholderbach (commented):
Awesome! Thanks

src/history/base.rs
rgwood: Hmm, that's a good idea but it's a little tricky because `create_filled_example_history()` doesn't actually create files. It creates an in-memory database for SQLite and never sets the filename for the text history.

It would be a good idea to have tests that exercise filesystem functionality though 🤔

sholderbach: I thought we had such tests but maybe that was before the sqlidation or are only on the individual implementations.

rgwood: @sholderbach I've added a test that uses a file-backed history.

rgwood: omg the reedline CI is so fast compared to Nushell 😭

sholderbach:
Awesome! Thanks.

Issue #539, opened 2023-02-15
Find common string returns wrong index for the root that matches


alesito85:
**Platform** macOS
**Terminal software** nushell

I've noticed this issue in nushell and have reported it there https://github.com/nushell/nushell/issues/8078
The problem is that depending on the available words, the function could potentially return a wrong index at which the root ends.

## Steps to reproduce
1. Word list that the issue is seen in ["qml", "qmake", "qmltc", "qmlls", "qmldom", "qmake6", "qmltime", "qmllint", "qmlscene", "qmlformat", "qmleasing", "qmlpreview", "qmlprofiler", "qmlplugindump", "qmltestrunner"]
2. Run this input through find_common_string
3. Observe that the function returns index 3 when it should return 2

kronberger-droid:
fixed in #969

PR #540, opened 2023-02-15
Attempt at fixing find_common_string

alesito85:
This PR attempts to fix issue #539

### Description

I suspect it may have to do with the way map works, so I've done the check differently. But I'm new with Rust so there's probably a better option.

I've also updated the test case so that it shows the issue that it previously didn't.

### To-Do

After seeing reviews for this PR I may also check the test function for non_ansi characters.

### DIFF
--- a/src/menu/menu_functions.rs
+++ b/src/menu/menu_functions.rs
@@ -145,31 +145,54 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
 pub fn find_common_string(values: &[Suggestion]) -> (Option<&Suggestion>, Option<usize>) {
     let first = values.iter().next();
 
+    let mut new_index: Option<usize> = match first {
+        Some(first) => Some(first.value.len()),
+        None => Some(0),
+    };
+
+    if values.len() <= 1 {
+        return (first, new_index);
+    }
+
     let index = first.and_then(|first| {
-        values.iter().skip(1).fold(None, |index, suggestion| {
-            let x = first
-                .value
-                .char_indices()
-                .zip(suggestion.value.char_indices())
-                .find(|((_, mut lhs), (_, mut rhs))| {
-                    lhs.make_ascii_lowercase();
-                    rhs.make_ascii_lowercase();
-
-                    lhs != rhs
-                });
-
-            match x {
-                Some((lhs, rhs)) => {
-                    let (lhsi, _) = lhs;
-                    let (rhsi, _) = rhs;
-                    match index {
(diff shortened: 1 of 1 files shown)


alesito85:
@rgwood please take a look when you can.

sholderbach:
Thanks for expanding the test. Haven't looked into the implementation details to make a Post mortem what went wrong there.

alesito85:
I'm eager to do more work on this if you find it necessary. Honestly I'm interested myself about why exactly the original code didn't work properly but as I was reading it I couldn't really figure it out. Maybe I just need some timeout.

Thanks for the review.

src/menu/menu_functions.rs
```diff
@@ -463,7 +505,101 @@ mod tests {
...
+                append_whitespace: false,
+            })
+            .collect();
+        let res = find_common_string(&input);
+
+        match res {
+            (Some(first), Some(index)) => assert!(
+                first.value == input.first().unwrap().value && index == first.value.len()),
+            _ => assert!(false, "There should be a result with both options as some with one element to check against"),
+        }
```
sholderbach, line 531: This whole pattern is repeated, can we factor it out so the tests are more readable and we are more confident what the differences between the tests are (and avoid introducing a bug in the test fixture)

src/menu/menu_functions.rs
```diff
@@ -145,12 +145,24 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
 pub fn find_common_string(values: &[Suggestion]) -> (Option<&Suggestion>, Option<usize>) {
     let first = values.iter().next();
 
+    let mut new_index: Option<usize> = match first {
+        Some(first) => Some(first.value.len()),
+        None => Some(0),
+    };
```
sholderbach, line 151: That's a `.map(...).unwrap_or_default()`

src/menu/menu_functions.rs
```diff
@@ -145,12 +145,24 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
...
+
+    if values.len() <= 1 {
+        return (first, new_index);
+    }
```
sholderbach, line 155: Maybe worth adding a comment what the different cases are. We have this early return and the chain from hell below.

sholderbach (commented):
I still struggle to follow the logic but that is also not your fault. I will skip on it for the upcoming release but would like to get the fix as soon as possible.

kronberger-droid:
fixed in #969

PR #544, opened 2023-02-26
allow reedline ctrl+o to take editor arguments

merelymyself:
relevant: https://github.com/nushell/nushell/pull/8105

allows ctrl+o keybinding to take command arguments into account


sholderbach (approved):
Thanks! Looks good to me

sholderbach:
OK based on https://github.com/nushell/nushell/issues/8714 this will break on Windows with the omnipresence of spaces in filepaths as we still need to separate flags from the command.

PR #548, opened 2023-03-12
fix: singlebyte char assumption in `parse_selection_char`

micron-mushroom:
`parse_selection_char` was working under the assumption that all characters are 1 byte wide, which caused nushell to crash when character sequences like "は!!" were entered (nushell/nushell#8412).

### DIFF
--- a/src/menu/menu_functions.rs
+++ b/src/menu/menu_functions.rs
@@ -124,13 +124,10 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
                         action,
                     }
                 }
-                _ => {
-                    index += 1;
-                    continue;
-                }
+                _ => {}
             }
         }
-        index += 1;
+        index += char.len_utf8();
     }
 
     ParseResult {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -283,7 +283,7 @@ checksum = "e11dcc7e4d79a8c89b9ab4c6f5c30b1fc4a83c420792da3542fd31179ed5f517"
 dependencies = [
  "cfg-if",
  "rustix",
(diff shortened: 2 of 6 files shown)


micron-mushroom:
And I was working under the assumption that the "CI gods" would have no problem with what I submitted. Until I realized I hadn't updated my toolchain in months.

sholderbach:
Can we add a test that reproduces the problem in nushell/nushell#8412 so we don't break this again. This whole code path is pretty complex, so I am not fully confident on the spot, if the removal of the addition in the catch all case is fine.

src/menu/menu_functions.rs
```diff
@@ -124,13 +124,10 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
                         action,
                     }
                 }
-                _ => {
-                    index += 1;
-                    continue;
-                }
```
sholderbach, line 130: Is this functionally really equivalent?

micron-mushroom: I believe so since both statements are within the `while` loop. They would be different if the `continue;` wasn't there in the first place, but it is. Just to check, I compiled nushell with the edited version to see if there were any problems. I didn't run into any, so I reasoned they were the same.

sholderbach: You are right! Thanks, yeah after your change it is also a bit more readable.

sholderbach:
I would love to see some more tests in that area but I think for now let's land the straightforward fix.
Thanks for picking up that issue!

ryanwhitehouse:
Added some tests here https://github.com/nushell/reedline/pull/553/files :)

PR #552, opened 2023-03-15
Make `Event` public to allow implements custom `EditMode`

### LINKED ISSUE #547: Cannot implement a custom `EditMode`
Implementing the `EditMode`  trait requires to use the `Event` type, which comes from `crossterm` but is not re-exported from `reedline`, which prevents implementation:

```rust
pub struct Editor;

impl EditMode for Editor {
    fn parse_event(&mut self, event: crossterm::event::Event) -> reedline::ReedlineEvent {
        todo!()
    }

    fn edit_mode(&self) -> reedline::PromptEditMode {
        todo!()
    }
}
```

```rust
error[E0053]: method `parse_event` has an incompatible type for trait
  --> repl/src/editor.rs:10:38
   |
10 |     fn parse_event(&mut self, event: crossterm::event::Event) -> reedline::ReedlineEvent {
   |                                      ^^^^^^^^^^^^^^^^^^^^^^^
   |                                      |
   |                                      expected enum `crossterm::event::Event`, found enum `Event`
   |                                      help: change the parameter type to match the trait: `crossterm::event::Event`
   |
   = note: expected signature `fn(&mut editor::Editor, crossterm::event::Event) -> ReedlineEvent`
              found signature `fn(&mut editor::Editor, Event) -> ReedlineEvent`
```

ClementNerma:
Fixes #547

### DIFF
--- a/src/edit_mode/base.rs
+++ b/src/edit_mode/base.rs
@@ -1,5 +1,5 @@
 use crate::{enums::ReedlineEvent, PromptEditMode};
-use crossterm::event::Event;
+pub use crossterm::event::Event;
 
 /// Define the style of parsing for the edit events
 /// Available default options:
--- a/src/edit_mode/mod.rs
+++ b/src/edit_mode/mod.rs
@@ -4,7 +4,7 @@ mod emacs;
 mod keybindings;
 mod vi;
 
-pub use base::EditMode;
+pub use base::{EditMode, Event};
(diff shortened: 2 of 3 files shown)


sholderbach:
We will probably break this when updating to the newer `crossterm` version supporting finer grained control over the KeyEvents. But let's land it for now to make sure we export a useable API also in the future.

PR #553, opened 2023-03-17
Allow multi byte characters as the marker without panicking

ryanwhitehouse:
This is an attempt to fix https://github.com/nushell/nushell/issues/8412.
I believe that the issue is caused because of the reference to reedline.  

I believe that if we implement this fix, bump the version and use the new version inside nushell, then the bug should be resolved.  

This is my first PR in Rust so happy to change anything you see as an issue!

### DIFF
--- a/src/menu/menu_functions.rs
+++ b/src/menu/menu_functions.rs
@@ -262,12 +262,12 @@ mod tests {
     #[cfg(feature = "bashisms")]
     #[test]
     fn handles_multi_byte_char_as_marker_and_number() {
-        let buffer = "Testは4!";
+        let buffer = "searchは6";
         let parse_result = parse_selection_char(buffer, 'は');
 
-        assert_eq!(parse_result.remainder, "Test");
-        assert_eq!(parse_result.index, Some(4));
-        assert_eq!(parse_result.marker, Some("は4"));
+        assert_eq!(parse_result.remainder, "search");
+        assert_eq!(parse_result.index, Some(6));
+        assert_eq!(parse_result.marker, Some("は6"));
     }
 
     #[cfg(feature = "bashisms")]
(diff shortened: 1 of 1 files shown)


src/menu/menu_functions.rs
```diff
@@ -88,7 +88,7 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
                 }
                 Some(&x) if x.is_ascii_digit() || x == '-' => {
                     let mut count: usize = 0;
-                    let mut size: usize = 1;
+                    let mut size: usize = marker.len_utf8();
```
sholderbach, line 91: Correct me if I am wrong but doesn't size refer to the digit or `-` sign which is guaranteed to be single byte ascii in this case.

src/menu/menu_functions.rs
```diff
@@ -73,7 +73,7 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
                     return ParseResult {
                         remainder: &buffer[0..index],
                         index: Some(0),
-                        marker: Some(&buffer[index..index + 2]),
+                        marker: Some(&buffer[index..index + 2 * marker.len_utf8()]),
```
sholderbach, line 76: Well spotted! (Probably rare that a multibyte char is a marker but based on the API we should do that)

sholderbach (commented):
Thanks for adding the tests!

src/menu/menu_functions.rs
ryanwhitehouse: Yeah, marker.len_utf8() would always be 1 here.   I will change it back :)

src/menu/menu_functions.rs
ryanwhitehouse: Yeah, definitely an edge case that isn't common.

src/menu/menu_functions.rs
ryanwhitehouse: This is needed to make this test succeed, since the marker is a multi byte char.  
```
fn handles_multi_byte_char_as_marker_and_number() {
        let buffer = "Testは4!";
        let parse_result = parse_selection_char(buffer, 'は');

        assert_eq!(parse_result.remainder, "Test");
        assert_eq!(parse_result.index, Some(4));
        assert_eq!(parse_result.marker, Some("は4"));
    }
```

ryanwhitehouse: The size is only used to pull out the marker value.  
```
marker: Some(&buffer[index..index + size]),
```
Usually it would just be the one character (marker.len_utf8()) but it will expand to pull in the number or - as well.

sholderbach (approved):
Thanks!

PR #555, opened 2023-03-18
Add const to some new functions.

jaudiger:
Constify the applicable new functions.

There are a lot more functions that could be constified. But I prefer to do it: step by step where it makes sense.

PS: I already did some tests in local, and adding const to some functions unlock the constification to more functions.


jaudiger:
Any thought on it? If this is not the right time, I can just close it. I would prefer to keep the number of active opened PR as low as possible

PR #560, opened 2023-03-25
Update crossterm to version 0.26.1

WindSoilder:
**WARN**: do not merge this yet, it's a breaking change to nushell.

After investigate, the main changes from crossterm `0.24.0` to `0.26.1` includes the following:
1. remove `crossterm::cursor::CursorShape` and `crossterm::cursor::SetCursorShape`, using `crossterm::cursor::SetCursorStyle` instead.
2. supports [kitty_keyboard_protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), so [crossterm::event::Event](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyEvent.html) adds `kind` and `state` fields.  In the case, I don't think we need to take care of `state` field, but need to be careful to `kind` field, when we release a key, crossterm will fire a [Release event](https://docs.rs/crossterm/latest/crossterm/event/enum.KeyEventKind.html#variant.Release).

Followed by @sholderbach 's suggestion, I create a `ReedlineRawEvent`, which is a wrapper for `crossterm::event::Event`, `ReedlineRawEvent` will make sure that it won't be created by `Release event`. 

Also add an example `event_listener_kitty_proto.rs` to play with kitty protocol.

## Additional note about wezterm:
`wezterm` have an issue about `Release event`, it will pass `Press event` to crossterm.  So we have to suggest user to turn off [enable_kitty_keyboard option](https://wezfurlong.org/wezterm/config/lua/config/enable_kitty_keyboard.html), luckly it's disabled by default.  Here is the relative issue: https://github.com/wez/wezterm/issues/3220

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,9 +187,9 @@ dependencies = [
 
 [[package]]
 name = "crossterm"
-version = "0.24.0"
+version = "0.26.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab9f7409c70a38a56216480fba371ee460207dd8926ccf5b4160591759559170"
+checksum = "a84cda67535339806297f1b331d6dd6320470d2a0fe65381e79ee9e156dd3d13"
 dependencies = [
  "bitflags",
  "crossterm_winapi",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ doctest = true
 [dependencies]
 chrono = "0.4.19"
 clipboard = { version = "0.5.0", optional = true }
-crossterm = { version = "0.24.0", features = ["serde"] }
+crossterm = { version = "0.26.1", features = ["serde"] }
 itertools = "0.10.3"
 nu-ansi-term = "0.47.0"
 serde = { version = "1.0", features = ["derive"] }
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -12,7 +12,7 @@ use {
     },
 };
 
-use crossterm::cursor::CursorShape;
+use crossterm::cursor::SetCursorStyle;
 use reedline::CursorConfig;
 #[cfg(not(any(feature = "sqlite", feature = "sqlite-dynlib")))]
 use reedline::FileBackedHistory;
@@ -61,8 +61,8 @@ fn main() -> Result<()> {
     let completer = Box::new(DefaultCompleter::new_with_wordlen(commands.clone(), 2));
 
     let cursor_config = CursorConfig {
-        vi_insert: Some(CursorShape::Line),
-        vi_normal: Some(CursorShape::Block),
+        vi_insert: Some(SetCursorStyle::BlinkingBar),
+        vi_normal: Some(SetCursorStyle::SteadyBlock),
         emacs: None,
     };
 
--- a/examples/event_listener.rs
+++ b/examples/event_listener.rs
@@ -38,11 +38,17 @@ fn print_events_helper() -> Result<()> {
             // It's guaranteed that read() wont block if `poll` returns `Ok(true)`
             let event = crossterm::event::read()?;
 
-            if let Event::Key(KeyEvent { code, modifiers }) = event {
+            if let Event::Key(KeyEvent {
+                code,
+                modifiers,
+                kind,
+                state,
+            }) = event
+            {
                 match code {
                     KeyCode::Char(c) => {
                         println!(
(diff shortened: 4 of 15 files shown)


fdncred:
Nice. I'd be happy to close https://github.com/nushell/reedline/pull/464 if this implements the stuff it was trying to.

WindSoilder:
Oh, sorry I don't know #464 exists, I should check it first :-(

This pr is just a migration without adding any new features.

After taking a closer look at #464, I found it didn't handle `KeyEventKind::Release`, I guess it may generate 2 `d` if we just type `d` on windows sytem or some other terminals.

fdncred:
> Oh, sorry

No worries at all. I'm super pumped to see you work on this. Great work!

> After taking a closer look at https://github.com/nushell/reedline/pull/464, I found it didn't handle KeyEventKind::Release, I guess it may generate 2 d if we just type d on windows system or some other terminals.

#464 was very incomplete. I'd go with your implementation but just add the things that were left out that are in 464 maybe?

WindSoilder:
Sure, I can try to have a look :-)  Maybe the most tricky part it how to handle `Paste` Event

fdncred:
We could also do it in 2 steps.
1. land something that gets us on the latest and works in nushell
2. land other enhancements like paste

WindSoilder:
I'd prefer to do it in 2 steps, to keep our changeset smaller :-)

WindSoilder:
Blocked by https://github.com/sharkdp/lscolors/pull/58

sholderbach:
Thanks for tackling this big challenge @WindSoilder!

That whole wezterm bug is really driving home what a mess the whole terminal "protocol" is on the edges. Props to the crossterm team for still making it work!

Sorry that I don't have too much time this week to review or red-team this PR.

Things I would check manually as they are not really directly tested on our side:

- entering non-ASCII characters both in uppercase and lowercase (we had to special-case that in the past for characters that use `AltGr` on European keyboards which also maps in some cases to Ctrl-Alt #139 ).
- paste behavior for single and multiline input.
- If modifier keys still work with special keys like `Alt-arrow` etc.

fdncred:
From my testing, I haven't seen any problems. I've only tested on windows with Windows Terminal and with WezTerm with `enable_kitty_keyboard = true`.

sholderbach:
Gave it some manual testing as well and seems to keep all expected functionality.

I would like to see this landed with some priority so we can finally get rid of the duplicated old crossterm version in our dependencies.

Great work @WindSoilder !

fdncred:
I think there's 2 options to landing this now.
1. wait until sharkdp makes a new LS_COLORS release
2. include the lscolors crate pointing to the branch that has this fix, which landed a couple weeks ago. Only problem is, if there isn't a new release by the time we release in ~3weeks we'll have to roll back i think.

WindSoilder:
I'd prefer 1st option, and created a relative issue to ask for a new version..

https://github.com/sharkdp/lscolors/issues/63

WindSoilder:
> I think there's 2 options to landing this now.

Since we don't get reply yet, I'd propose 3rd option: copy-paste lscolors' crossterm relative code in the pr: https://github.com/nushell/nushell/pull/8623/commits/e423dca5d5b36032bf3c5242e51ffe61abbe1597

Then we can play with latest crossterm version.  And we can adjust our dependency once lscolor is published.

sholderbach:
Option 4: (I think would also need a version change) use nu_ansi_term instead of crossterm to interface with lscolors. It is lighterweight (should finish earlier in the compile graph) and has been seeing less changes in recent months.

fdncred:
I just noticed that lscolors has a nu-ansi-term feature. We should use that.

WindSoilder:
> Option 4: (I think would also need a version change)

@sholderbach Yeah, it's still need lscolor version change too....just create a pr to update it...
https://github.com/sharkdp/lscolors/pull/64

WindSoilder:
Hooray!  lscolors released a new version with latest crossterm and nu_ansi_term version, I've adjusted nushell pr, so it's ready for now

fdncred:
I vote to land this. Even if it doesn't enable all the features of crossterm 26.1, it gets us on that version where we can add those features later that were in my original PR for crossterm 25.

sholderbach (approved):
Thanks so much! Let's go

fdncred:
Leaving this here in case we start seeing problems. Some people are having problems with crossterm + Windows. See the linked issues here. https://github.com/mikaelmello/inquire/issues/132

PR #561, opened 2023-03-25
Forward buffer content and offset to EditMode

ClementNerma:
This is a breaking change.

It would allow to improve keybindings precision, such as autocompleting a quote (") only if the character on the right of the cursor isn't already one.

### DIFF
--- a/src/edit_mode/base.rs
+++ b/src/edit_mode/base.rs
@@ -7,7 +7,7 @@ pub use crossterm::event::Event;
 /// - Vi
 pub trait EditMode: Send {
     /// Translate the given user input event into what the `LineEditor` understands
-    fn parse_event(&mut self, event: Event) -> ReedlineEvent;
+    fn parse_event(&mut self, event: Event, line: &str, cursor: usize) -> ReedlineEvent;
 
     /// What to display in the prompt indicator
     fn edit_mode(&self) -> PromptEditMode;
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -108,7 +108,7 @@ impl Default for Emacs {
 }
 
 impl EditMode for Emacs {
-    fn parse_event(&mut self, event: Event) -> ReedlineEvent {
+    fn parse_event(&mut self, event: Event, _line: &str, _offset: usize) -> ReedlineEvent {
         match event {
             Event::Key(KeyEvent { code, modifiers }) => match (modifiers, code) {
                 (modifier, KeyCode::Char(c)) => {
@@ -179,7 +179,7 @@ mod test {
             modifiers: KeyModifiers::CONTROL,
             code: KeyCode::Char('l'),
         });
-        let result = emacs.parse_event(ctrl_l);
+        let result = emacs.parse_event(ctrl_l, "", 0);
 
         assert_eq!(result, ReedlineEvent::ClearScreen);
     }
@@ -198,7 +198,7 @@ mod test {
             modifiers: KeyModifiers::CONTROL,
             code: KeyCode::Char('l'),
         });
-        let result = emacs.parse_event(ctrl_l);
+        let result = emacs.parse_event(ctrl_l, "", 0);
 
         assert_eq!(result, ReedlineEvent::HistoryHintComplete);
     }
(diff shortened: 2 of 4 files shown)


sholderbach:
I consider that to be the wrong place to implement such logic. The responsibility of `EditMode` is to convert the raw events coming in like a keypress into atomic operations defined elsewhere (the raison d`être for having `EditMode` primarily being our vi emulation and allowing rebinding of keys).

Adding that in my view violates separations of concerns and makes things much more state dependent than they already are. Furthermore it adds some stuff to the each `parse_event` call that is rarely if never needed there and bloats the code path in the long term.

Nushell implements the smarts around the `Completer` primarily through hacky bits in the `Menu` system. I don't like the state around that but that would need concerted work to separate some core completion functionalities and operations from the graphical UI elements in the menu.

Not sure what you are hacking so can not point you directly to a place where you could implement your logic instead.

ClementNerma:
I understand your concern.

Here is a usage example for this PR: in the shell I'm currently writing, I want the right arrow key to make the cursor go right (seems logical) but when it's at the very end of the line, where it can't do anything, I want it to autocomplete the suggestion provided by the history completer.

That's the default behaviour of ZSH's autocomplete plugin, and I believe Fish does the same.

Which is something that cannot be done with this crate (unless I misunderstood something).

I understand your concern with separating the keybindings and this kind of auto-completion, but in this case how could the behaviour I describe be achieved using `reedline`?

sholderbach:
We do this through having a special operation for the history hint completion and let the exectuion of this operation report if it succeeds and fall through to the base move.

See the default keybindings that define that https://github.com/nushell/reedline/blob/9d6e22dd06d79b3f5ece716ef560148714c5489c/src/edit_mode/keybindings.rs#L128-L136

ClementNerma:
I see, thanks for the example!

Another use case I have is auto-closing symbols. For instance, writing `(` should automatically append a matching `)`.
Except if the character on the cursor's right is `(`, in which case we skip it (= we just go right).

How could this be implemented?

sholderbach:
Mhh this sounds like something that requires careful changes to the reedline internals.

Requirements: 
- reedline itself shouldn't carry strong assumption about the lexing/grammar. (so a language using should be reasonably be able to define special cases) (current violator in the code base is the `bashism` feature)
- configurability without too much pain. (users might not like the auto closing behavior)

> Another use case I have is auto-closing symbols. 
> For instance, writing `(` should automatically append a matching `)`. 

In theory we could have a callback after every keypress that could modify the buffer. This would give you the freedom to implement stuff like that but doesn't compose well and would certainly incur a performance hit.

> Except if the character on the cursor's right is `(`, in which case we skip it (= we just go right).

Have an additional `ReedlineEvent`/`EditCommand` in place of the basic char insertion for the opening braces/parens. 
This would probably conflict with the current design of the edit modes that generally hand all printable characters without modifiers through to be inserted into the buffer or search.

Either the config point would be the keybinding setup as a lot of stuff is handled currently or this would be an additional option (not that `Reedline`'s builder isn't already bloated)



One thing I would caveat: 

Is it truly desirable to have auto closing parens in a line editor experience? At least in nushell, we use the unclosed paren as a way to let the user press enter and get a multiline input. If the paren would autoclose, pressing enter would let the `Validator` we implemented conclude everything is hunky-dory and the input would be submitted.

ClementNerma:
So it isn't possible in the current state of `reedline`, got it.

I'm currently building my own shell, so all decisions I'm making aren't definitive, I'm still designing everything :)

It's true that using a good validator may be better than parenthesis, but my current design is that I use the parsed AST of the input to perform the syntax highlighting. Invalid syntax in the input (e.g. missing parenthesis or closing quote) = no highlighting at all. Which is not ideal but far easier than implementing a new lexer.

sholderbach:
Will close this PR as I don't think locally reimplementing `EditMode` will get you started off quickly without duplicating a bunch of work or colliding with some assumptions in the rest of reedline.

ClementNerma:
I understand, I'll keep maintaining my fork (which simply contains the PR I've made) as I can't do anything else for now, but I agree it's best to keep this pr closed for now :)

PR #562, opened 2023-03-27
Allow history searching via session id

fdncred:
This PR attempts to allow reedline to have the ability to search the entire history, *OR* search only based on the current history session id. To demonstration this functionality, I've added a few commands to the demo example.

1. `history` - This shows the entire history regardless of session
2. `history session` - This shows the history only with the current history session id
3. `history sessionid` - This shows the current history session id or None
4. `toggle history_session` - This shows the ability to start off without a history session, create one, and then start to show only the history items with the current history session id. It also shows the opposite. You can toggle the history session searching off.

I've only tested this with the sqlite_backed history. More testing is required.
I haven't added any new tests. It may be a good idea to add some but I'm not quite sure how.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -66,7 +66,17 @@ fn main() -> Result<()> {
         emacs: None,
     };
 
+    // Setting history_per_session to true will allow the history to be isolated to the current session
+    // Setting history_per_session to false will allow the history to be shared across all sessions
+    let history_per_session = false;
+    let mut history_session_id = if history_per_session {
+        Reedline::create_history_session_id()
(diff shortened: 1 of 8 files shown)


sholderbach:
Thanks for moving our built-in history facilities forward. 

All that feature-gated code we already have in different places is really unfortunate. The same applies to the testing that either tests one implementation or the other depending on the provided feature flag. We have some tests that should really behave the same accross histories but other things shouldn't hold us back from providing the best experience based on the particular version. Maybe we need to move some of those tests to a more dependency-injection style.

PR #566, opened 2023-04-06
History filter

### LINKED ISSUE #516: Make insertion to the history optional (e.g. with leading space)

samlich:
Supersedes #518 @rgwood fixes #516 @sholderbach

This allows excluding commands from the history by prefixing with a space (or other prefix), by wrapping the history with a filter.

Fish does this pretty well, so I try to follow their functionality. It stores one item in memory, so you can still recall it, before the next command is run.

I don't know how functions like `History::count` are used, so it doesn't include that item in the count. And it doesn't return it for searches other than over a range.


src/history/base.rs
```diff
@@ -89,7 +89,7 @@ pub struct SearchQuery {
     /// if given, only get results after/before this id (depending on direction)
     pub end_id: Option<HistoryItemId>,
     /// How many results to get
-    pub limit: Option<i64>,
+    pub limit: Option<usize>,
```
sholderbach, line 92: Is there a particular rationale for expanding the value range here? I think `i64` was introduced to be compatible with the sqlite data types but I think we are not too specific here.

src/history/base.rs
```diff
@@ -178,6 +178,45 @@ pub trait History: Send {
...
+
+    fn delete(&mut self, h: HistoryItemId) -> Result<()> {
+        (**self).delete(h)
+    }
+
+    fn sync(&mut self) -> std::io::Result<()> {
+        (**self).sync()
+    }
+}
+
```
sholderbach, line 219: I don't see the reason for reimplementing `History` for `Box<impl History>` as the trait is wholy defined for working with references to the `History` which is trivially obtainable from a boxed value.

src/history/file_backed.rs
```diff
@@ -77,7 +77,12 @@ impl History for FileBackedHistory {
     fn load(&self, id: HistoryItemId) -> Result<super::HistoryItem> {
         Ok(FileBackedHistory::construct_entry(
             Some(id),
-            self.entries[id.0 as usize].clone(),
+            self.entries
+                .get(id.0 as usize)
+                .ok_or(ReedlineError(ReedlineErrorVariants::OtherHistoryError(
+                    "Item does not exist",
+                )))?
+                .clone(),
```
sholderbach, line 85: Did you run into problems here that caused a panic? Should we fix that if necessary?

src/history/filter.rs
```diff
@@ -0,0 +1,167 @@
...
+        self.exclusion_prefix = exclusion_prefix;
+    }
+
+    /// Change history being wrapped. Existing entries are not modified according to the filter.
+    pub fn set_wrapped(&mut self, wrapped: T) {
+        self.wrapped = wrapped;
+    }
+}
+
+impl<T: History> History for HistoryFilter<T> {
```
sholderbach, line 39: Quite frankly, I think wrapping a `History` in something that is itself a `History` feels overly complicated.
Maybe it shows that the API or how we use it inside the `Reedline` engine is not quite right. Can we fix this in a compositional way that doesn't look like "OOP inheritance we have at home"?

src/history/filter.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// Change or remove prefix used to filter items. Existing entries are not modified.
+    pub fn set_exclusion_prefix(&mut self, exclusion_prefix: Option<String>) {
+        self.exclusion_prefix = exclusion_prefix;
+    }
+
+    /// Change history being wrapped. Existing entries are not modified according to the filter.
+    pub fn set_wrapped(&mut self, wrapped: T) {
+        self.wrapped = wrapped;
+    }
```
sholderbach, line 36: Do we actually need to be able to modify all this?

src/history/filter.rs
```diff
@@ -0,0 +1,167 @@
+use crate::{
+    result::{ReedlineError, ReedlineErrorVariants},
+    History, HistoryItem, HistoryItemId, Result, SearchDirection, SearchFilter, SearchQuery,
+};
+
+/// A history that wraps another history, and does not forward values beginning with `exclusion_prefix` (if present).
+/// If an item is filtered, it is stored in memory and able to be retrieved, until the next item is inserted.
+#[derive(Debug)]
+pub struct HistoryFilter<T> {
```
sholderbach, line 9: Based on the usage pattern in `src/engine.rs` I would make this `pub(crate)` only. To me, through the builder pattern it looks like an internal implementation detail.

sholderbach (changes requested):
Thanks for giving this a shot!

Great to see that on first glance in your implementation it doesn't panic when trying to ignore the first entry of a session with the sqlite history (see my problem with https://github.com/nushell/reedline/pull/518#issuecomment-1326853062). (maybe worth adding a test case/fool proof API to guard against that in the future)

Two functional problems at the moment:

1. It needs to work when building with the `bashism` feature which adds the history substition in the buffer like `sudo !!` used in nushell (see the CI failure)
2. adding the ` ` only suppresses the entry from being stored in the file to be recalled in the next session. If i enter a line e.g. ` ls` I can still call it up by pressing `up-arrow`. In bash this would already be suppressed. (we may discuss this behavior if other shells/line editors deviate from that)

From a maintainer perspective I would like to see a few changes to make this implementation a little less complex to keep it simple to understand/extend/fix in the future.

In particular to me wrapping the existing `History` in something that itself just `impl History` feels like patching some stuff and repeating a bunch of other work. The risk of making assumptions about `History` behavior that are not true for other `History`s becomes quite high in my mind.

src/history/file_backed.rs
samlich: When testing, I had a bug that sent the i64::MAX-1 used for the filtered-item id to the wrapped history, which caused a panic. But I don't think any actual code ever uses an ID that doesn't exist.

src/history/base.rs
samlich: I changed it cause the file-backed and filter used usize. I'll change it back.

samlich:
It should compile with bashisms, as I reverted the i64->usize change. I don't know whether it is the correct behavior of substituting the excluded command, or if it shouldn't do that. I'm guessing it shouldn't, if bash doesn't save the item at all, and this should replicate bash.

The up-arrow behavior is copied from fish, I'm guessing just so you can still access the command if you accidentally put a space before it.

I did the wrapper design for the behavior from fish, where the filtered command still acts like a history item, until the next command is entered. It could probably just be implemented on `Reedline`, and still even have that function (if it's desired), since I didn't even make it searchable anyway.

If you don't want that feature, then I think it would just mean adding an extra flag that says whether a command was filtered or not, and then not doing anything in `update_last_command_context`, when set.

With that feature, the excluded command would be stored in `Reedline`, and then an additional flag that says whether the history cursor is on that item. I think that would only add changes to `next_history`, `previous_history`, and `update_buffer_from_history`.

samlich:
I've made the changes described in the previous comment, split into two commits.

The "keep 1 filtered history item" commit adds the behavior from fish, where one filtered item is retained. It does add a bit of complexity.

Let me know if this looks good, and whether or not you want that second commit, and I'll squash the changes and add a test(s).

sholderbach (commented):
Thanks for iterating here! This looks a lot more digestible as a diff and hopefully easier to maintain.

For the exact behavior around going back to the last entry I would welcome the feedback from other contributors, what their preferences are.

One thing we need to test is how holding back the last item interacts with nushells `history` command as this reads the history directly from the file instead of sharing it in memory (https://github.com/nushell/nushell/blob/bdaa32666aca357eadab2449d53e358662309bc8/crates/nu-cli/src/commands/history.rs#L57-L89)

sholderbach (approved):
Everyone seems to be fine with the fish-like up arrow behavior for now.

So let's give this a testing in nushell as soon as possible!

Thanks for resolving this @samlich

PR #567, opened 2023-04-06
enable deleting history items with keybinding

samlich:
Inspired by https://github.com/fish-shell/fish-shell/issues/9454

When a history item is selected, delete it with shift-del (apparently a standard binding, according to the issue linked above).

Will not delete (it just refuses, without showing feedback), when using a file-backed history where the history item being targeted has already been written to the file. It's not trivial to implement for this case, due to the concurrent modification of the text file.

I don't understand the

```
self.editor.move_to_start(UndoBehavior::HistoryNavigation);
self.editor.move_to_line_end(UndoBehavior::HistoryNavigation);
```
statements, so those changes in src/engine.rs may not be correct. I see in some places it just does `move_to_line_end`.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1117,7 +1117,7 @@ impl Reedline {
                 {
                     Some(Ok(())) => {
                         self.update_buffer_from_history();
-                        // TODO: are these needed/correct?
+                        // Move to end of first line, see `Self::previous_history()`.
                         self.editor.move_to_start(UndoBehavior::HistoryNavigation);
                         self.editor
                             .move_to_line_end(UndoBehavior::HistoryNavigation);
@@ -1188,6 +1188,7 @@ impl Reedline {
             .back(self.history.as_ref())
             .expect("todo: error handling");
         self.update_buffer_from_history();
+        // Move to end of *first* line, so that pressing up again goes directly to previous item.
         self.editor.move_to_start(UndoBehavior::HistoryNavigation);
         self.editor
             .move_to_line_end(UndoBehavior::HistoryNavigation);
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -770,8 +770,7 @@ impl Reedline {
                     .delete_item_at_cursor(self.history.as_mut())
                 {
                     Some(Ok(())) => Ok(EventStatus::Handled),
-                    None => Ok(EventStatus::Inapplicable),
-                    Some(Err(_)) => Ok(EventStatus::Inapplicable),
+                    Some(Err(_)) | None => Ok(EventStatus::Inapplicable),
                 }
             }
             ReedlineEvent::PreviousHistory | ReedlineEvent::Up | ReedlineEvent::SearchHistory => {
@@ -1108,24 +1107,22 @@ impl Reedline {
                 Ok(EventStatus::Handled)
             }
             ReedlineEvent::DeleteHistoryItem => {
-                if self.input_mode == InputMode::HistoryTraversal {
-                    match self
-                        .history_cursor
-                        .delete_item_at_cursor(self.history.as_mut())
-                    {
-                        Some(Ok(())) => {
-                            self.update_buffer_from_history();
-                            // are these needed/correct?
-                            self.editor.move_to_start(UndoBehavior::HistoryNavigation);
-                            self.editor
-                                .move_to_line_end(UndoBehavior::HistoryNavigation);
-                            Ok(EventStatus::Handled)
-                        }
-                        None => Ok(EventStatus::Inapplicable),
-                        Some(Err(_)) => Ok(EventStatus::Inapplicable),
+                if self.input_mode != InputMode::HistoryTraversal {
+                    return Ok(EventStatus::Inapplicable);
+                }
+                match self
+                    .history_cursor
+                    .delete_item_at_cursor(self.history.as_mut())
+                {
+                    Some(Ok(())) => {
+                        self.update_buffer_from_history();
+                        // TODO: are these needed/correct?
+                        self.editor.move_to_start(UndoBehavior::HistoryNavigation);
+                        self.editor
+                            .move_to_line_end(UndoBehavior::HistoryNavigation);
+                        Ok(EventStatus::Handled)
                     }
-                } else {
-                    Ok(EventStatus::Inapplicable)
+                    Some(Err(_)) | None => Ok(EventStatus::Inapplicable),
                 }
             }
             ReedlineEvent::Multiple(events) => {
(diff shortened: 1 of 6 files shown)


src/engine.rs
```diff
@@ -670,6 +670,16 @@ impl Reedline {
                 // A handled Event causes a repaint
                 Ok(EventStatus::Handled)
             }
+            ReedlineEvent::DeleteHistoryItem => {
+                match self
+                    .history_cursor
+                    .delete_item_at_cursor(self.history.as_mut())
```
sholderbach, line 676: Do we need to update what is visible in the traditional `Ctrl-r` search? (this is not the menu but the `reedline::ReedlineEvent::SearchHistory` triggered search.)

src/engine.rs
```diff
@@ -670,6 +670,16 @@ impl Reedline {
...
+                {
+                    Some(Ok(())) => Ok(EventStatus::Handled),
+                    None => Ok(EventStatus::Inapplicable),
+                    Some(Err(_)) => Ok(EventStatus::Inapplicable),
```
sholderbach, line 680: This can be simplified to a single pair of `match` arms or an `if let`.

src/engine.rs
```diff
@@ -1003,6 +1013,27 @@ impl Reedline {
...
+                    match self
+                        .history_cursor
+                        .delete_item_at_cursor(self.history.as_mut())
+                    {
+                        Some(Ok(())) => {
+                            self.update_buffer_from_history();
+                            // are these needed/correct?
+                            self.editor.move_to_start(UndoBehavior::HistoryNavigation);
+                            self.editor
+                                .move_to_line_end(UndoBehavior::HistoryNavigation);
```
sholderbach, line 1027: I am not quite sure why they both appear for `previous_history` but only one for `next_history`. I can not immediately recall what problem with undo and redo might have prompted it.

Worth thinking about and playing with. This operation is irreversible on the history. Will the undo/redo move leave the buffer in a broken state?

src/engine.rs
```diff
@@ -1003,6 +1013,27 @@ impl Reedline {
...
+                            Ok(EventStatus::Handled)
+                        }
+                        None => Ok(EventStatus::Inapplicable),
+                        Some(Err(_)) => Ok(EventStatus::Inapplicable),
+                    }
+                } else {
+                    Ok(EventStatus::Inapplicable)
```
sholderbach, line 1034: Again some nesting levels that can be simplified.

src/history/file_backed.rs
```diff
@@ -191,13 +191,26 @@ impl History for FileBackedHistory {
...
+    fn delete(&mut self, h: super::HistoryItemId) -> Result<()> {
+        let id = usize::try_from(h.0).map_err(|_| {
+            ReedlineError(ReedlineErrorVariants::OtherHistoryError(
+                "Invalid ID (not usize)",
+            ))
+        })?;
+        if self.len_on_disk <= id {
+            self.entries.remove(id);
+            Ok(())
+        } else {
```
sholderbach, line 203: I think this really needs to be covered by tests. The `len_on_disk` starts to get interesting when we need to trim the buffer and other sessions are potentially having a say during `.sync()`

src/history/cursor.rs
```diff
@@ -88,6 +88,34 @@ impl HistoryCursor {
...
+                Ok(()) => {
+                    // Cursor must be moved *after* deletion, as deleting may
+                    // alter which entry a `HistoryItemId` points to.
+                    if self.back(history).is_err()
+                        || self.current.is_none()
+                        || self.current.as_ref().unwrap().id == Some(id)
+                    {
+                        if self.forward(history).is_err() {
+                            self.current = None;
+                        }
```
sholderbach, line 105: see the clippy fail.

sholderbach (commented):
Sounds like an interesting addition.

I can see the concern from https://github.com/fish-shell/fish-shell/pull/9515 that it possibly should have a consistent behavior when not in history traversal mode.

Before landing this we should give it some adverserial testing trying to break it manually.

I assume for the sqlite history you were able to get it to correctly delete? It would be great if we could get it to consistently purge the text file based history as well, I think if we provide the feature the users will have the asssumption that it will delete their secrets consistently.

src/engine.rs
samlich: If you mean `self.update-buffer_from_history()`, it doesn't seem like it, as `history_search_paint()` is used instead of `buffer_paint()`, and that function doesn't support `SubstringSearch`.

src/engine.rs
samlich: I didn't look closely enough, it's `move_to_line_end` vs `move_to_end` from [this change](https://github.com/nushell/reedline/pull/306/files#diff-ff278ec599adfd033d35a083ebe8948db8bf177c9817dee8ddc4f3a9e3f5728eR831) to deal with multiline entries. I've added a comment.

samlich:
I still need to do the tests.

I made it so it clears the buffer, if not going through history. If it fails to delete a history item, it doesn't clear it, in order to signify that.

For the file history, the concern is that if one instance deletes a line, and then another instance tries to delete a (different) line afterwards, its index is going to have changed. I guess the way to solve this would be to reload it and then search for a matching string. I'm thinking add a `delete: Option<usize>` field, which is used by `sync()`. The delete function would then call `sync`, if the entry is stored on the file. Does that seem reasonable?

samlich:
I made it so for the file-backed history, it will delete all occurrences of the referenced entry. This solves the concurrency issue, but changes the behavior of `delete()`. I will sometimes rename a script, but it will keep recommending the old name, and since I've run the script many times, there will be many entries, so I like this behavior better.

Let me know if that seems good, and I'll make the same change to the sqlite implementation, and then add the tests.

sholderbach:
> I made it so for the file-backed history, it will delete all occurrences of the referenced entry. This solves the concurrency issue, but changes the behavior of `delete()`. I will sometimes rename a script, but it will keep recommending the old name, and since I've run the script many times, there will be many entries, so I like this behavior better.

Yeah I can see the value of your proposed behavior both for purging the history from something secret as well as to suppress a history completion/hint that is no longer relevant.

Will ping the rest of the team if they want to advocate for a particular behavior.

> Let me know if that seems good, and I'll make the same change to the sqlite implementation, and then add the tests.

Awesome! Thanks for sticking with us and sorry for my radio silence on the PRs

sholderbach:
If I remember @fdncred 's feedback correctly the main concern here was the behavior of what happens when you try to delete a history entry when you are not actually on the history but just hit the keybinding. Maybe that is an additional source of confusion and we should just ignore that?

cc @amtoine @rgwood for UX comments

amtoine:
i'm not sure what should be tested with the UX here and how :confused:

stormasm:
https://github.com/nushell/nushell/issues/11620

Since this is still open I am referencing the above issue as well....

kronberger-droid:
Closing as stale, same id problem as #736 and two review threads never got tests.
`SqliteBackedHistory::delete` exists on main now, so a sqlite-only keybinding would be a small new PR if anyone wants it.

PR #569, opened 2023-04-11
re-export crossterm::style::Color for custom prompt implementations

xiuxiu62:
Some default Prompt methods require crossterm::Color for manual implementation (`get_prompt_color` for example).  This re-exports Color for use by custom Prompt implementers.


fdncred:
Thanks

Issue #558, opened 2023-03-23
Pasting multi-line content from some terminal emulator is broken


sigoden:
**Platform**: macOS
**Terminal software**: iTerm2 + nvim embeded terminal.

## Steps to reproduce
Using iterm2, open neovim 0.8.3 (installed from homebrew)
```
nvim
```
In neovim, type `:terminal` to open an embedded terminal.

copy the following multi-line content (using Command+c):
```
hello world

hello world

hello world
```

in the embedded terminal opened by neovim, run
```
cargo run --example demo
```

type `{` to enter multiline mode, type `cmd+v` to paste the contents, will see the result:

<img width="849" alt="Screenshot 2023-03-23 at 16 12 50" src="https://user-images.githubusercontent.com/4012553/227142200-e96bba31-8475-4301-8934-d93949c5e9b3.png">

The multi-line content is concated into a single line.

PR #571, opened 2023-04-17
support bracketed paste

WindSoilder:
Closes: #543
Fixes: #558

Once it's ok, we can make nushell support bracketed paste easily :-)

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -1,7 +1,8 @@
 use {
     crossterm::{
-        event::{KeyCode, KeyModifiers},
-        Result,
+        cursor::SetCursorStyle,
+        event::{DisableBracketedPaste, EnableBracketedPaste, KeyCode, KeyModifiers},
+        execute, Result,
     },
     nu_ansi_term::{Color, Style},
     reedline::{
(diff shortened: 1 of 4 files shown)


rgwood (approved):
LGTM, thanks for the fix!

@sholderbach I think this is pretty low-risk... any objection to merging?

sholderbach:
Code looks good to me!

But the behavior for pasting multiline text will be different depending on whether the bracketed paste mode is enabled.
With the regular text pasting we will validate the input after each newline and submit if complete.

If memory serves correctly, this was partially done to reflect the admittedly dangerous behavior of other line editors/shells that accept and run snippets you copy from stackoverflow directly. 
But this might have also been kept this way as the nushell repl execution side behaved weirdly when given multiple lines.

WindSoilder:
> But the behavior for pasting multiline text will be different depending on whether the bracketed paste mode is enabled.

Yeah, it's true, actually I'm not really sure If we want to make it to be default.
As a relatively low level lib, I think it's good to make nushell itself to make the decision.  Which is in nushell pr: https://github.com/nushell/nushell/pull/8907
What do you think?  Or is it better to make it built-in in reedline?

PR #573, opened 2023-04-21
feat: add `current_insertion_point()` to return the current insertion point of the input buffer

stevenxxiu:
For [fix: set the initial repl cursor pos, fixes #8943 by stevenxxiu · Pull Request #8955 · nushell/nushell](https://github.com/nushell/nushell/pull/8955).


fdncred:
Thanks!!

PR #574, opened 2023-04-22
Restore the cursor shape when reedline exits

nibon7:
Restore the cursor shape when reedline exits. 
Tested on Linux (Alacritty 0.13.0) and Windows terminal.


fdncred:
Nice @nibon7, people have been asking for this. One question, does `SetCursorStyle::DefaultUserShape` set it to what the cursor shape used to be before launching reedline/nushell, or is it setting it to some default?

fdncred:
This makes it sound like restoring the cursor to whatever it was originally isn't possible across all platforms https://github.com/crossterm-rs/crossterm/issues/646

nibon7:
> This makes it sound like restoring the cursor to whatever it was originally isn't possible across all platforms [crossterm-rs/crossterm#646](https://github.com/crossterm-rs/crossterm/issues/646)

Maybe. I tested on windows and linux, and it seems ok.

* Windows Terminal  (block)

https://user-images.githubusercontent.com/15247421/233819649-647ab520-290f-43a2-a0a6-4aac5b7a31f8.mp4

* Windows Terminal (underscore)

https://user-images.githubusercontent.com/15247421/233819655-1b1fa45c-cb41-4df6-81c8-8c93427ca92f.mp4


* Linux Alacritty (block)

https://user-images.githubusercontent.com/15247421/233819212-43921953-170c-4535-b5fd-b1683a38dbc8.mp4

* Linux Alacritty (underscore)

https://user-images.githubusercontent.com/15247421/233819221-c43bd5b5-0ff9-46c9-a3c3-291657553fa2.mp4

fdncred:
I'm fine with moving ahead with it, even thought it doesn't seem like it would work. Your examples seem to prove otherwise.

src/engine.rs
```diff
@@ -141,6 +142,15 @@ struct BufferEditor {
 
 impl Drop for Reedline {
     fn drop(&mut self) {
+        if self.cursor_shapes.is_some() {
+            let _ignore = terminal::enable_raw_mode();
+            let mut stdout = std::io::stdout();
+            let _ignore = stdout.queue(SetCursorStyle::DefaultUserShape);
+            let _ignore = stdout.queue(Show);
+            let _ignore = stdout.flush();
+            let _ignore = terminal::disable_raw_mode();
```
sholderbach, line 151: We don't need to duplicate it

```suggestion
```

sholderbach:
Thanks!

maxomatic458:
is there a way to disable this behavior?
because im unable to use the block cursor in windows terminal now

sholderbach:
> is there a way to disable this behavior? 

Not yet/(not sure if nushell will unset the cursor_shape option when none is given in the config)

> because im unable to use the block cursor in windows terminal now

Based on the settings on Windows terminal or inside nushell?

maxomatic458:
i set my cursor to block in windows terminal.
Is there a option in nushells config for the cursor?

sholderbach:
> i set my cursor to block in windows terminal. Is there a option in nushells config for the cursor?

In the config we have the option here to allow you to disambiguate between the vi modes

https://github.com/nushell/nushell/blob/d45e9671d41fa7625de9d337704c8903de35ebbe/crates/nu-utils/src/sample_config/default_config.nu#L286-L290

mb21:
For reference, this seems to work on Apple Terminal on macOS 13.5.1 (block):


https://github.com/nushell/reedline/assets/157497/1a5bdf90-a01f-40a6-a60b-ef5756720e1b

Issue #576, opened 2023-04-27
Bracketed Paste does not preserve newline characters in some terminal


sigoden:
**Platform** macOS
**Terminal software** iTerm2

Bracketed Paste does not preserve newline characters in some terminal.

## Steps to reproduce
1. run demo
```
cargo run --example demo
```

2. copy multiple line text
```
struct BufferEditor {
    editor: String,
    extension: String,
}
```
3. paste texts to demo repl
```
}   extension: String,ufferEditor {   
```

This may related to  crossterm-rs/crossterm#780 
In some terminal, they write \r instead of \n when pasting.

```
Event: Paste("struct BufferEditor {\r    editor: String,\r    extension: String,\r}")
```

sholderbach:
Thanks for the report!

This seems to be related to the trouble we were observing in nushell/nushell#8907

sigoden:
We can fix it with following patch.
```diff
diff --git a/src/edit_mode/emacs.rs b/src/edit_mode/emacs.rs
index ec35384..54067aa 100644
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -156,7 +156,7 @@ impl EditMode for Emacs {
             Event::Resize(width, height) => ReedlineEvent::Resize(width, height),
             Event::FocusGained => ReedlineEvent::None,
             Event::FocusLost => ReedlineEvent::None,
-            Event::Paste(body) => ReedlineEvent::Edit(vec![EditCommand::InsertString(body)]),
+            Event::Paste(body) => ReedlineEvent::Edit(vec![EditCommand::InsertString(body.replace('\r', "\n"))]),
         }
     }
 
diff --git a/src/edit_mode/vi/mod.rs b/src/edit_mode/vi/mod.rs
index 2a20155..ab5dc1b 100644
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -153,7 +153,7 @@ impl EditMode for Vi {
             Event::Resize(width, height) => ReedlineEvent::Resize(width, height),
             Event::FocusGained => ReedlineEvent::None,
             Event::FocusLost => ReedlineEvent::None,
-            Event::Paste(body) => ReedlineEvent::Edit(vec![EditCommand::InsertString(body)]),
+            Event::Paste(body) => ReedlineEvent::Edit(vec![EditCommand::InsertString(body.replace('\r', "\n"))]),
         }
     }
 
```

[...]

PR #577, opened 2023-04-28
fix bracketed paste

WindSoilder:
Fixes: #576

Actually the issue comes from iterm2 setting when I want to paste multiline script:
img

But yeah, when we enable bracketed paste, we can avoid pasted script to be executed automatically, I think it's safe to transfer from `\r` to `\n`. 

Currently I'm not sure how windows handle bracketed feature, since it doesn't work with crossterm lib, I'm going to convert from "\r\n" to "\n" to make sure that everything is working properly

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -156,7 +156,9 @@ impl EditMode for Emacs {
             Event::Resize(width, height) => ReedlineEvent::Resize(width, height),
             Event::FocusGained => ReedlineEvent::None,
             Event::FocusLost => ReedlineEvent::None,
-            Event::Paste(body) => ReedlineEvent::Edit(vec![EditCommand::InsertString(body)]),
+            Event::Paste(body) => ReedlineEvent::Edit(vec![EditCommand::InsertString(
+                body.replace("\r\n", "\n").replace('\r', "\n"),
+            )]),
         }
     }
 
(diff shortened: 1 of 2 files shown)


fdncred:
Where is that transforms dialog in iterm2?

WindSoilder:
It's auto popped up when I want to paste multiline content..
I guess it's enabled by `Warnings` config
图片

fdncred:
ok, i see it now. thanks

sholderbach:
Are we good here with the result of the PR so we can enable bracketed paste on the nushell side if desired?

fdncred:
@sholderbach it works on my mac just fine. I'm not on windows today so I can't test it there. Not sure about linux.

WindSoilder:
> we can enable bracketed paste on the nushell side if desired?

After this pr, I think we can enable bracketed paste on mac/Linux, but still not on windows, it relys on crossterm issue

fdncred:
> it depends on crossterm issue

can't we check the platform and have a windows fix too? until crossterm decides to fix it?

WindSoilder:
Sorry I don't have a work around for this on windows, see the comment: https://github.com/nushell/reedline/issues/576#issuecomment-1525731687

It indicats that crossterm::Paste event doesn't work on windows.

fdncred (approved):
at least on windows, it looks like it's pasting multi-line content fine. @sholderbach i'm fine with landing it.

PR #578, opened 2023-04-29
fix: prompt position on resize

Banyc:
Fix: https://github.com/nushell/nushell/issues/9015

Previously, when the terminal window shrinks its height, the prompt eats the previous line.
Now, the prompt will not overwrite the previous line.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -413,8 +413,9 @@ impl Painter {
             // Note: you can't just subtract the offset from the origin,
             // as we could be shrinking so fast that the offset we read back from
             // crossterm is past where it would have been.
-            let prompt_height = prev_terminal_size.1 - prev_prompt_row;
-            self.prompt_start_row = height - prompt_height;
+            let prompt_height = prev_terminal_size.1.saturating_sub(prev_prompt_row);
+            let prompt_height = prompt_height.max(1);
+            self.prompt_start_row = height.saturating_sub(prompt_height);
         } else if prev_terminal_size.1 < height {
             // Terminal is growing down, so move the prompt down the same amount to make space
             // for history that's on the screen
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -407,13 +407,14 @@ impl Painter {
         self.terminal_size = (width, height);
         // TODO properly adjusting prompt_origin on resizing while lines > 1
 
-        if prev_prompt_row >= (height - 1) {
+        if prev_prompt_row >= height {
             // Terminal is shrinking up
             // FIXME: use actual prompt size at some point
             // Note: you can't just subtract the offset from the origin,
             // as we could be shrinking so fast that the offset we read back from
             // crossterm is past where it would have been.
-            self.prompt_start_row = height - 2;
+            let prompt_height = prev_terminal_size.1 - prev_prompt_row;
+            self.prompt_start_row = height - prompt_height;
         } else if prev_terminal_size.1 < height {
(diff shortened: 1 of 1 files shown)


Banyc:
I haven't checked the u16 overflow and the UX_TESTING yet because it's time to go to bed here so I will do the checks and tests tomorrow.

sholderbach:
Thanks so much for giving this a shot, there have been many futile attempt, would be great if you can resolve that!

For UX testing, interesting edge cases are:

Resize related
- vertical resizing (as primarily your issue in https://github.com/nushell/nushell/issues/9015)
- horizontal resizing

What happens below the prompt
- buffer contains multiple lines due to wrapping text
- triggering the completion menu

Possible artifacts due to assumption reedline made from its development with a relatively consistent prompt.
- (different prompts (there might be some assumptions left from our initial set of prompts))

Cursor position behavior is some what dependent on the scrollback of the terminal emulator
- initially opened terminal with out scrolling
- at the bottom of the screen where the scroll buffer will expand.

Banyc:
After a bunch of random tests, I found out there is a trade off between **not overwriting any previous output** and **producing fewer incomplete duplicate prompts** on window resizing. I prefer the former since:

- we can just ignore the previous incomplete prompt fragments;
- the previous output is very precious.

So I decided to just set the `prompt_start_row` at the last line every time the terminal window resizes for now. Then we take a look at how the other shell deal with this problem and come up with an incremental solution.

Banyc:
Basically every time the window changes in size, it's expected to see incomplete duplicate prompt fragments. The main focus of the checklist expects no outputs being overwritten by the prompt.

Thanks for the checklist so that I found a bug on horizontal resizing just now (and fixed)!

# Checklist

Resize related

- vertical resizing (as primarily my issue in https://github.com/nushell/nushell/issues/9015): fixed
- horizontal resizing: no output being eaten

What happens below the prompt

- buffer contains multiple lines due to wrapping text: ok with example/demo
- triggering the completion menu: ok with example/completions

Possible artifacts due to assumption reedline made from its development with a relatively consistent prompt.

- (different prompts (there might be some assumptions left from our initial set of prompts))
  - example/demo: ok
  - example/custom_prompt: ok but no completions

Cursor position behavior is some what dependent on the scrollback of the terminal emulator

- initially opened terminal without scrolling: the cursor is at the right position
- at the bottom of the screen where the scroll buffer will expand: no sure how to operate on this test

sholderbach:
I will have to put aside a bit of time for playing around but your suggestions sound good and the items on the list are promising!


---
> I found out there is a trade off between **not overwriting any previous output** and **producing fewer incomplete duplicate prompts** on window resizing. I prefer the former since:
> 
>     * we can just ignore the previous incomplete prompt fragments;
> 
>     * the previous output is very precious.

Fully agree on your preferences here! Keeping the potentially valuable output to me would be higher priority than dealing fully gracefully with the rarer resizing. (I know that resizing can also be quite frequent if you change a tiling layout in desktop or terminal multiplexer...)

---
Regarding your question on the test suggestions:


> at the bottom of the screen where the scroll buffer will expand: no sure how to operate on this test

I was thinking about pressing enter often enough that the prompt is at the last line of the screen to fit the buffer or multiline buffer there. The screen shouldn't start flowing over with content because text wrapping is not correclty accounted for in the prompt positioning.

jntrnr:
I played around with the demo a bit, and it looked pretty good. I think we could land and dogfood in nightly nushell to see how it works in practice.

Thanks for the PR!

Issue #546, opened 2023-03-10
Characters are not handled correctly in keybindings


ClementNerma:
I'm trying to integrate custom keybindings with the editor, here is what I got:

```rust
let mut keybindings = default_emacs_keybindings();

keybindings.add_binding(
    KeyModifiers::NONE,
    KeyCode::Char('c'),
    ReedlineEvent::Edit(vec![EditCommand::Clear]),
);

// ...
rl.with_edit_mode(Emacs::new(keybindings)) // ...
```

This doesn't work. If I type the letter "c" nothing happens. I also tried with various keys, tried in various keyboard layouts including QWERTY (which is not my native layout), nothing works.

If I use another key code such as `KeyCode::Tab`  it works perfectly. But with custom characters specifically it doesn't seem to work.

Any idea why?

ClementNerma:
So I did a bit of testing, and here are some of the results:

`Ctrl + Backspace` produces a `Ctrl + W`
`Shift + Enter` produces an `Enter` event (no modifier)

sholderbach:
The binding you describe in the initial description should be possible after #580 

For the other problems you describe `cargo run --example event_listener` can be helpful to see if we are actually able to receive or disambiguate the event

> `Ctrl + Backspace` produces a `Ctrl + W` 

I think this combination can be an artifact of both how the terminal handles it and it is encoded in the basic ASCII representation underlying the ANSI terminal protocol. e.g. on Windows Terminal in WSL2 I can observe `Ctrl-H` instead of `Ctrl-Backspace` with `cargo run --example event_listener`/nu's `keybinding listen` as this corresponds to the character that aligns in the [Control code block of ASCII](https://en.wikipedia.org/wiki/ASCII#Control_code_chart)

> `Shift + Enter` produces an `Enter` event (no modifier)

This is also very terminal emulator implementation dependent.

ClementNerma:
Hi, I'm stuck again on this problem after coming back to a project of mine that uses reedline.

How do you work around this limitation? `cargo run --example listener` gives me the correct keys, but `ReedlineRawEvent::into()` doesn't.

sholderbach:
Can you open a new issue with the particular difference you observe between the listener and the internals?

ClementNerma:
Followed at #663

PR #580, opened 2023-05-05
Allow binding of CTRL+ALT+<c>, SHIFT+<c> and CTRL+ALT+SHIFT+<c>

TrMen:
# Context
Nushell exposes `control_alt` and similar keybinding modifiers, but currently it's not possible to actually use those mappings in nushell (at least on none of my keyboards).

From some limited testing, it appears like GNU readline also doesn't allow mapping `CTRL+ALT+f` and similar keybinds. I don't know if there's some deeper reason behind this, or just a limitation of readline itself. I can't think of any edgecases my change would break, since it's only if you actively choose to add a binding.

I use those keybinds quite extensively in other shells, so it'd be nice to have in nu as well.

# Commit message
Combinations like `Ctrl+Alt+a` could not be mapped previously, since some non-us keyboard layouts use those combinations for sending characters like `@`.

Now, first check if there a binding was defined using that combination, else interpret as raw characters. This should not affect those special keyboard layouts, since they would be sending `Ctrl+Alt+<special char>`, rather than what would be on the US layout.

E.g.
- Map `Ctrl+Alt+a` to an action
- `Ctrl+Alt+a` on (some) german keyboards would send `@`
- A german layout pressing those keys would send a key combination of `Ctrl+Alt+@`, so they are unaffected by this change (they would have to specifically map `Ctrl+Alt+@`)
- For any layout actually sending `Ctrl+Alt+a`, they get the mapped action

This affects emacs mode and vi insert mode (vi normal mode already allows such keybindings).

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -127,24 +127,28 @@ impl EditMode for Emacs {
                         _ => c.to_ascii_lowercase(),
                     };
 
-                    if modifier == KeyModifiers::NONE
-                        || modifier == KeyModifiers::SHIFT
-                        || modifier == KeyModifiers::CONTROL | KeyModifiers::ALT
-                        || modifier
-                            == KeyModifiers::CONTROL | KeyModifiers::ALT | KeyModifiers::SHIFT
-                    {
-                        ReedlineEvent::Edit(vec![EditCommand::InsertChar(
-                            if modifier == KeyModifiers::SHIFT {
-                                c.to_ascii_uppercase()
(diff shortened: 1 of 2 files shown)


sholderbach (approved):
Good call!

Yes with your change we should be able to relax the rules around both the `AltGr` maps as well as if someone tries to do some more insidious binding of keys without modifiers. (e.g. clear the buffer if you enter a `!` :P )

Closes #546 in that latter aspect.
Closes https://github.com/nushell/nushell/issues/8977

PR #583, opened 2023-05-18
Add quick input mode for menu

woxiaosa:
In my scenario, commands selected in menu can be executed directly without additional parameters. Therefore, it is unnecessary for user to press 'enter'.

I add a quick input mode for menu. In quick input mode, when user select a menu item, `read_line` return directly instead of waiting for 'Enter'.

usage:
```Rust
let completion_menu = Box::new(ColumnarMenu::default().with_name("completion_menu").with_quick_input());
```


fdncred:
Can you add an example that demonstrates how this works? I'd like to see some testing too if we can test something like this. Not sure if we can.

woxiaosa:
I've provided examples in `examples/menu_quick_input.rs`, and I've made demo videos for them.

When we select item, we simply press Enter once instead of twice.


https://github.com/nushell/reedline/assets/39560876/9e6359cc-95ba-4604-952d-18e79d4e3fee

PR #588, opened 2023-05-30
Update to make work with new lines embedded in sent string.

fnordpig:
Currently if a sent string doesn't contain \r, it just prints on a new line without the carriage return:

This is the expected output:

This fixes that by always splitting the received line into lines.  The example is updated to demonstrate.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1669,7 +1669,8 @@ impl Reedline {
             let result = external_printer.receiver().try_recv();
             match result {
                 Ok(line) => {
-                    messages.push(line);
+                    let lines = line.lines().map(String::from).collect::<Vec<_>>();
+                    messages.extend(lines);
                 }
                 Err(TryRecvError::Empty) => {
                     break;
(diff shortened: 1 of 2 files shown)


src/engine.rs
```diff
@@ -1669,7 +1669,8 @@ impl Reedline {
             let result = external_printer.receiver().try_recv();
             match result {
                 Ok(line) => {
-                    messages.push(line);
+                    let lines = line.lines().map(String::from).collect::<Vec<_>>();
```
fdncred, line 1672: Will this prevent streaming since you're collecting here? or is this only collecting one line or something small that won't artificially slow things down?

fnordpig: It depends on how big a string blob is sent through the channel to the printer.  Generally it shouldn't impact anything - it'll be a single line of output and a single line coming out of lines().  For something massive with lots of new lines, it'll split the large string up into lines and pass them through.  The alternative would be to parse the string and replace all \n with \r\n, possibly with some regex to not end up with \r\r\n's, which I don't think would be much faster as you'll need to create a new string in memory.

fdncred:
ok, let's try it out. thanks!

PR #591, opened 2023-06-06
Upgrade `nu-ansi-term` to 0.48.0 release

sholderbach:


sholderbach:
Holding of on landing this one for the `0.20.0` release as nushell also needs to have the same version in its interactions with https://github.com/sharkdp/lscolors

sholderbach:
Also see the accidental breaking change that slipped in with this release we should probably resolve https://github.com/nushell/nu-ansi-term/issues/46

PR #592, opened 2023-06-09
add a new disable_bracketed_paste method

WindSoilder:
Want to add `disable_bracketed_paste` method for reedline engine, hope it can make nushell works with `bracketed_paste` better.

Relative issue: https://github.com/nushell/nushell/issues/9218
I think paste password inside `gpg` should work fine even when `bracketed_paste` is enabled.

To fix the issue, we can have something like a `DisableBracketedPaste` guard during nushell repl.  But it requires reedline to provide this funciton.


fdncred:
I think if you merge main this typos thing should be fixed.

PR #598, opened 2023-06-28
always disable bracketed paste after read_line

WindSoilder:
According to @sholderbach 's comment: https://github.com/nushell/nushell/pull/9399#issuecomment-1585692641

It's good to disable bracketed paste in reedline rather than nushell

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -573,8 +573,9 @@ impl Reedline {
 
         let result = self.read_line_helper(prompt);
 
+        #[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Sounds good! Thanks @WindSoilder

LevitatingBusinessMan:
The fact that this was merged makes me seriously reconsider using this project.

LevitatingBusinessMan:
I think the problem is that in your previous PR you were re-setting the bracketed paste. Now it's just being disabled indefinitely.

PR #601, opened 2023-07-11
Replace crossterm ScrollUp with universal workaround

NotLebedev:
Replace scrolling via escape sequence with workaround that prints newlines to retain scrollback regardless of terminal implementation of scroll sequence. See [nushell/9166](https://github.com/nushell/nushell/issues/9166) for discussion and related info.

This PR replaces using ScrollUp sequence `"\[e<num>S"` with printing desired amount of blank lines at the last line of terminal window. This workaround functions on all terminals that have scrollback buffer regardless of their ability to handle and implementation of ScrollUp sequence.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -10,7 +10,7 @@ use {
     crossterm::{
         cursor::{self, MoveTo, RestorePosition, SavePosition},
         style::{Attribute, Print, ResetColor, SetAttribute, SetForegroundColor},
-        terminal::{self, Clear, ClearType, ScrollUp},
+        terminal::{self, Clear, ClearType},
         QueueableCommand, Result,
     },
     std::io::Write,
@@ -156,7 +156,7 @@ impl Painter {
             self.prompt_start_row = 0;
         } else if required_lines >= remaining_lines {
             let extra = required_lines.saturating_sub(remaining_lines);
-            self.stdout.queue(ScrollUp(extra))?;
+            self.queue_universal_scroll(extra)?;
             self.prompt_start_row = self.prompt_start_row.saturating_sub(extra);
         }
 
@@ -475,7 +475,7 @@ impl Painter {
         let final_row = self.prompt_start_row + self.last_required_lines;
         let scroll = final_row.saturating_sub(self.screen_height() - 1);
         if scroll != 0 {
-            self.stdout.queue(ScrollUp(scroll))?;
+            self.queue_universal_scroll(scroll)?;
         }
         self.stdout
             .queue(MoveTo(0, final_row.min(self.screen_height() - 1)))?;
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Thanks for resolving this!

This sounds reasonable and your helper is amazingly documented.

NotLebedev:
Also anybody interested can test this changes by compiling nushell against  this revision of reedline and running two following nushell scripts:
1. `$env.PROMPT_COMMAND = {|| "\n>\n>" }; clear; for i in (term size | get rows | 1..$in) {print $i}` It will clear screen. Fill it with numbers and then scroll two lines. Scroll your terminall up a bit, all lines should be there. Before fix it looked like this on my windows terminal:
  image
after change all numbers are present:
  image
2. `$env.PROMPT_COMMAND = {|| ">" }; clear; for i in (term size | get rows | 1..($in - 5)) {print $i}` and then press `F1` or `ctrl + R` (or invoke any menu with your preferred shortcut) This will scroll terminal several lines up. Difference is that this time scroll is initiated when cursor is not at the bottom of window, but it still should work. Before fix i could not scroll up at all (notice absence of scroll bar on right) because nothing was moved into scrollback buffer by ScrollUp:
  image
after change all numbers are available and I can scroll up:
  image

fdncred:
Love it! Nice work @NotLebedev!

fdncred:
@NotLebedev I'm seeing missing lines in Windows Terminal but seemingly only when maximized.
```
❯ detect columns --help
```
produces this problem on the first example.
```
Examples:
  Splits string across multiple columns
  > 'a b c' | detect columns -n
  │ # │ column0 │ column1 │ column2 │
  ├───┼─────────┼─────────┼─────────┤
  │ 0 │ a       │ b       │ c       │
  ╰───┴─────────┴─────────┴─────────╯
```
The top of the table is cutoff and it's the very first line.
image

Then I scroll up and see it cut off
image

NotLebedev:
> @NotLebedev I'm seeing missing lines in Windows Terminal but seemingly only when maximized.

Is master branch (or your branch) of nushell using this patch? It seems like at least master does not. Here in `Cargo.lock` https://github.com/nushell/nushell/blob/545697c0b20b58f13b8dfbbefa700e2dbd08d864/Cargo.lock#L4215 revision is [cf841](https://github.com/nushell/reedline/commit/cf841beb92affc7cb5cdc16bddefdebcd001f8c9) and this change is one commit later on [4f31e](https://github.com/nushell/reedline/commit/4f31e200e46b621e96b313ad334d50375fcf624d)

fdncred:
rebuilding on latest main to make sure. stay tuned.

fdncred:
I think you're right @NotLebedev. I got the latest main, checkout the detect columns pr, merged main, then did cargo update -p reedline and now my reedline is on 4f31e. I missed doing the cargo update last time so my mistake. It's working great now. Sorry for the false alarm!

Issue #602, opened 2023-07-13
Panic when accepting element from help menu


dedebenui:
**Platform** macOS 13.4
**Terminal software** kitty 0.29.0

I don't really know in what broader context it occurs, but here's a sequence of input that will consistently make nushell crash. From the trace, it looks like the panic occurs in Reedline.

## Steps to reproduce
1. new shell
2. type
```
def fn [] {
return 
```

5. while still on the second line (ends with a space), hit F1 to get the help menu
6. start typing the name of a command
7. accept with enter

At this point, nushell panics.

<details>

<summary>Backtrace</summary>

```
thread 'main' panicked at 'assertion failed: self.is_char_boundary(idx)', /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/alloc/src/string.rs:1532:9
stack backtrace:
   0:        0x10a827f7b - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hae51cb91d407e2ef
   1:        0x109c7261b - core::fmt::write::h746bc0969202388b
   2:        0x10a82252c - std::io::Write::write_fmt::h4098c2c7437a0bd7
   3:        0x10a827d5a - std::sys_common::backtrace::print::h8360bf0158e89b36
   4:        0x10a82c461 - std::panicking::default_hook::{{closure}}::hedf04c568eb6e0bc
   5:        0x10a82c23f - std::panicking::default_hook::h62889b2c29e2347d
   6:        0x10a82caba - std::panicking::rust_panic_with_hook::h22edd02828aff274
   7:        0x10a82c838 - std::panicking::begin_panic_handler::{{closure}}::h6895f44a9b7e2caa
   8:        0x10a82a9e9 - std::sys_common::backtrace::__rust_end_short_backtrace::h013519e9b77978c5
   9:        0x10a82c5e2 - _rust_begin_unwind
  10:        0x10a9fa763 - core::panicking::panic_fmt::h2a51b60c12d11322
  11:        0x10a9fa7f7 - core::panicking::panic::h00b4c20719c0abc2
  12:        0x10a76a0e8 - reedline::core_editor::line_buffer::LineBuffer::insert_char::hf2a552f7a55a11e7
  13:        0x10a745e39 - reedline::core_editor::editor::Editor::run_edit_command::h624514b8504fc16c
  14:        0x10a75efe3 - reedline::engine::Reedline::run_edit_commands::hd8b36992cc5f2d1d
[...]

PR #612, opened 2023-07-21
Fix a crash when enabling multi-line editing and help menu

nibon7:
This is an attempt to fix a crash when enabling multi-line editing and help menu.

Related #602 nushell/nushell#9627

PS: Can't reproduce nushell/nushell#9627 on linux after this change.

[asciicast](https://asciinema.org/a/598319)

### DIFF
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -161,7 +161,8 @@ impl LineBuffer {
 
     /// Cursor position *in front of* the next unicode grapheme to the left
     pub fn grapheme_left_index(&self) -> usize {
-        self.lines[..self.insertion_point]
+        let index = std::cmp::min(self.insertion_point, self.lines.len());
+        self.lines[..index]
             .grapheme_indices(true)
             .last()
             .map(|(i, _)| i)


sholderbach (commented):
Thanks for hardening this particular function. Still a bit curious which places may corrupt the insertion point.

nibon7:
Seems the root cause is https://github.com/nushell/nushell/blob/0b1e368cea8fe921716ac3bd6353cbe29d46c700/crates/nu-cli/src/menus/description_menu.rs#L648-L650, when `start != 0`, `offset` may be greater than `lb.len()`  @sholderbach 

PS:  Maybe we can add `debug_assert` in [set_insertion_point](https://github.com/nushell/reedline/blob/ddec668b9258ad8452669dc7872814c2fba0be44/src/core_editor/line_buffer.rs#L71) to check whether the insertion point is valid.

* A possible quick fix:
```
diff --git a/crates/nu-cli/src/menus/description_menu.rs b/crates/nu-cli/src/menus/description_menu.rs
index 5f7434bb0..26db0afb7 100644
--- a/crates/nu-cli/src/menus/description_menu.rs
+++ b/crates/nu-cli/src/menus/description_menu.rs
@@ -646,7 +646,10 @@ impl Menu for DescriptionMenu {
                 |lb| {
                     lb.replace_range(start..end, replacement);
                     let mut offset = lb.insertion_point();
-                    offset += lb.len().saturating_sub(end.saturating_sub(start));
+                    offset += lb
+                        .len()
+                        .saturating_sub(end.saturating_sub(start))
+                        .saturating_sub(start);
                     lb.set_insertion_point(offset);
                 },
                 UndoBehavior::CreateUndoPoint,
```

sholderbach:
Mhh the fix for nushell in nushell/nushell#9784 seems pretty subtle (still need to wrap my brain fully around it :D) I think we should keep some protections on the reedline side. Maybe `debug_assert!` is to lenient :P Either full assert to panic early or a return value that forces you to think about it.

PR #615, opened 2023-07-31
turn off default prompt styling (bold) to prevent leakage

fdncred:
Before this PR, the bold attribute was hard coded and therefore leaked through. Example:

### Before
image
Notice that the prompt is green bold even though I told it just green. Also, when i draw the prompt with the `do` command, it's drawn right. This is because the `do` command is rendering a buffer and the PROMPT_COMMAND is going through this code in reedline and hard-coding bold everywhere.

### After
image
Notice that when i tell it green, the prompt is drawn in green and with the `do` command. When I tell it green_bold the prompt is drawn in green_bold as well as with the `do` command.

This is meant to close https://github.com/nushell/nushell/issues/9123 when it's implemented in nushell.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -256,8 +256,7 @@ impl Painter {
         // print our prompt with color
         if use_ansi_coloring {
             self.stdout
-                .queue(SetForegroundColor(prompt.get_prompt_color()))?
-                .queue(SetAttribute(Attribute::Bold))?;
+                .queue(SetForegroundColor(prompt.get_prompt_color()))?;
         }
 
         self.stdout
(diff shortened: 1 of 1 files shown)


src/painting/painter.rs
```diff
@@ -270,22 +269,22 @@ impl Painter {
...
+                .queue(SetForegroundColor(prompt.get_prompt_right_color()))?;
         }
 
         self.print_right_prompt(lines)?;
 
         if use_ansi_coloring {
-            self.stdout.queue(ResetColor)?;
+            self.stdout
+                .queue(SetAttribute(Attribute::Reset))?
+                .queue(ResetColor)?;
```
sholderbach, line 287: Let's keep this reset, so a user provided prompt style doesn't leak into the main buffer.

sholderbach (commented):
Yeah this styling should have no business being part of the `Painter`. With `Prompt`s being also just strings as in the nushell case we should not set style here.

I think we could reconsider even the whole prompt color thing.

Before we land please keep the reset defensive.

src/painting/painter.rs
fdncred: What do you mean? The reset is in there still, in fact I augmented it to reset the style attribute and reset the color. Are you waying you want to change this back to only the color reset?

sholderbach: Ah my mental pattern matching was reading it wrong with all the `SetAttribute` deletions. My bad.

Thanks for adding this!

sholderbach (approved):

Issue #348, opened 2022-03-14
Transient prompt support


jmoore34:
Possible alternative to #195 

Powerline prompts can be nice, but they can also fill the scrollback history with a lot of visual noise. Some readline implementations support transient prompts, in which after the user hits 'enter', the fancy powerline prompt is replaced with a much simpler prompt. 

Here is an example with zsh's powerlevel10k:
![transient 2](https://user-images.githubusercontent.com/1783464/158272359-15571bf6-ccec-4f38-aa56-84adfea438bf.gif)

sholderbach:
This definitively needs additional support from reedlines side to paint a different prompt after submission of an entry. Good catch!

Cactus-man:
I feel like before this is done the prompting system in general should be reevaluated before, as behavior in some possible situations is not well-defined. Most importantly, how should a prompt behave on line break characters in certain parts? In its current state it behaves unexpectedly (though I'd argue there's no "expected" behavior in this case anyway).  

![image](https://user-images.githubusercontent.com/64769435/211161766-bd9032d0-f03f-49be-8d79-bc8225dad80e.png)
_Fig 1. `Prompt::render_prompt_left` returns a string with line break_

![image](https://user-images.githubusercontent.com/64769435/211161808-94a4f41a-7cfd-4a3a-af81-29266c953489.png)
_Fig 2. the same as Fig 1. but in a different terminal_

This might seem obvious to not do if encountered by a developer but it reflects an asymmetry within the prompt rendering code which currently inhibits arbitrary Unicode prompts, which could be right-to-left.

fdncred:
transient prompt was implemented a while back, thanks.

PR #627, opened 2023-08-23
Let prompts choose to repaint on enter

ysthakur:
This PR goes towards addressing #348. It adds `fn repaint_on_enter(&self) -> bool` to the `Prompt` trait so that a prompt can choose to be repainted on enter. I figured an option could be added to `$env.config` on the Nushell side to enable a transient prompt. After that, the hooks described in [this issue](https://github.com/nushell/nushell/issues/7843) will hopefully start working.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -714,7 +714,14 @@ impl Reedline {
                 match self.handle_event(prompt, event)? {
                     EventStatus::Exits(signal) => {
                         if prompt.repaint_on_enter() {
-                            self.repaint(prompt)?;
+                            println!("{:?}", self.editor.line_buffer());
+                            if let Some(id) = self.history_last_run_id {
+                                if let Ok(last) = self.history.load(id) {
+                                    self.editor.edit_buffer(|buf| buf.insert_str(&last.command_line), UndoBehavior::UndoRedo);
+                                    self.repaint(prompt)?;
+                                    self.editor.edit_buffer(|buf| buf.clear(), UndoBehavior::UndoRedo);
+                                }
+                            }
                         }
                         // Move the cursor below the input area, for external commands or new read_line call
                         self.painter.move_cursor_to_end()?;
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1712,12 +1712,10 @@ impl Reedline {
         let buffer = self.editor.get_buffer().to_string();
         self.hide_hints = true;
         // Additional repaint to show the content without hints etc.
-        match &mut self.transient_prompt {
-            Some(transient_prompt) => {
-                let prompt = transient_prompt.clone();
-                self.repaint(prompt.as_ref())?
-            }
-            None => self.repaint(prompt)?,
+        if let Some(transient_prompt) = &self.transient_prompt {
+            self.repaint(transient_prompt.clone().as_ref())?
+        } else {
+            self.repaint(prompt)?;
         }
         if !buffer.is_empty() {
             let mut entry = HistoryItem::from_command_line(&buffer);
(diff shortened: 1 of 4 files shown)


fdncred:
If reedline is going to support a transient prompt, which would be totally rad, I'd love to see an example in this repo that shows folks how to make a basic one.

fdncred:
> Let me know if it'd be better to make a completely separate transient_prompt example (and restore the custom_prompt example).

Yes, another example would be best. It also has to be confirmed that your change doesn't break anything else or the other examples.

fdncred:
I ran this PR example. I'm not sure what it's supposed to do or how to make it work. I just get this
```
    Finished dev [unoptimized + debuginfo] target(s) in 3.55s
     Running `target/debug/examples/transient_prompt`
Transient prompt demo:
Abort with Ctrl-C or Ctrl-D
(transient_prompt example)>ls
::: a
::: b
::: d
:::
:::
:::
```

src/prompt/base.rs
```diff
@@ -97,6 +97,10 @@ pub trait Prompt: Send {
         &self,
         history_search: PromptHistorySearch,
     ) -> Cow<str>;
+    /// Whether to repaint the prompt after the user hits enter (for transient prompts)
+    fn repaint_on_enter(&self) -> bool {
+        false
+    }
```
sholderbach, line 103: Mhh I don't like this as a silently stateful API.

I would prefer this to be explicit. Three ideas that come to mind:

- making this an *explicit* setter, by preferrably having `&mut self` and either a descriptive name or an additional parameter setting the post submit state. Could still have No-Op default impl.
- passing the information that the prompt should take its post submit form directly to the `render_*` methods as a parameter. This has the downside of being a breaking change and increasing the API beginners have to learn
- adding additional `render_post_submit_*` methods that have a default implementation which would just invoke their regular form, but could be overridden by the more complex `Prompt` implementations.

sholderbach (changes requested):
Thanks for giving this a shot and working up an example and doing due diligence on the testing.

From a long term maintenance perspective I think this API is a bit too bug prone and can cause confusion.

fdncred:
I was getting ready to object to the `unwrap()` in the engine.rs, but it appears that we have many unwraps in this code base. Seems like we should fix those eventually. (not in this pr, of course).

fdncred:
Wait a second, now that I look at these unwraps, most of them are in tests, which is fine. I found 10 that are not in tests that we need to fix. @ysthakur, would you mind adding error handling to the one you have in engine.rs? I really don't think we want to continue adding unwraps in our code.

fdncred:
> there's also a bunch of .expect("todo: error handling")

good point. sounds like something for a followup pr, if anyone is interested. hint hint. 😆

src/engine.rs
```diff
@@ -434,6 +439,13 @@ impl Reedline {
         self
     }
 
+    /// Set a different prompt to be used after submitting each line
+    #[must_use]
+    pub fn with_transient_prompt(mut self, transient_prompt: Box<dyn Prompt>) -> Self {
+        self.transient_prompt = Some(Arc::from(Mutex::from(transient_prompt)));
+        self
+    }
```
sholderbach, line 447: If you can get by with this `fn (mut self, Box<>) -> Self` signature, then I don't think the whole `Arc<Mutex<Box<>>>` is needed, right?

sholderbach: @ysthakur sorry for not getting back to you!

Yes I like the variant with `.take()`. To me that is pretty idiomatic.

sholderbach (approved):
Thanks for sticking with us!

Let's land this, so the road is clear to figure out the best way to configure this in nushell.

(with https://github.com/nushell/nushell/blob/ba6d8ad2617c2100021d6e3c949c0b9088f2667b/Cargo.toml#L167 you can just `cargo update -p reedline` to get the latest unreleased version for testing)

fdncred:
Can't wait to see what this looks like in nushell. It should be interesting. Good work @ysthakur!

PR #629, opened 2023-09-01
Fix `read_line` docs

Hofer-Julian:
The result type is not `crossterm::Result` (anymore), but `std::io::Result`


sholderbach (approved):
Thanks!

PR #631, opened 2023-09-09
Cut/Yank multiple words

alsuren:
Currently, using control-w to cut multiple words leaves only
the most recently cut word in your cut buffer for yanking with
control-y.

This patch makes reedline behave like readline: it restores all
concecutive cuts as a single buffer. This means that typing
`control-w` `control-w` `control-y` gets you back where you started.

Before this is merged, it would be useful to:

- [x] ~make a `trait ClipboardExt` with `fn update(&mut self, content: &str, mode: ClipboardMode, operation: ReplaceOrAppendOrPrepent);` rather than copy-pasting the same 5 lines everywhere~ turns out there's a lot of repeated code to do with cutting, and it was easy enough to DRY out. It also made more sense to keep everything internal to Editor rather than pushing it out into the Clipboard trait.
- [x] write some unit tests
  - ~[ ] it may be possible to write a proptest/fuzz test that generates a random buffer and a random position and a random set of cut commands~
- [x] ~follow UX_TESTING.md to make sure I've not broken anything.~
  - [x] here the behavior in VI mode and probably also what happens when you undo/redo certain operations will be the most critical to me.
  - [ ] I also made a quick stab at a kind-of integration test suite for the vi edit mode keybindings, but I'm not really sure what `UntilFound([MenuUp, Up])` really means (as generated by the k key in vi), so I didn't test multiline editing that well. Feedback welcome, or we can just ship it as-is and improve it later.
- [x] double check whether each of the things should be a prepend or append operation. I only use control-w and control-y in readline, so I don't really know how the other edit commands are supposed to behave.
  - [x] with `1 2 3 4 5 6 7 8 9` in my editor, and the cursor in the middle, C-w followed by C-k or C-u then C-y does the same thing as bash, which is good enough for me. https://www.masteringemacs.org/article/keyboard-shortcuts-every-command-line-hacker-should-know-about-gnu-readline lists a whole bunch more keybindings, but I don't even know how to type meta on macos and I don't have a physical linux box around to test, so I didn't try all keybindings.
- ~[ ] should I actually check what readline's logic is? If I read the source code, does my contribution here become a derived work?~ I decided to treat it as a black box.

~There is a known edge case that if you delete a word and then delete the whole line that you're on, you won't get back where you started, because the line will be appended to the word.~ As far as I can tell, bash's emacs mode doesn't have any whole-line-cutting shortcuts, and vi now sends the StopCutting command, so it is impossible to hit this edge case. If it ever *becomes* a problem in the future then we could change the LastCutCommand state machine to include a ClipboardMode, and reset the cut buffer if ClipboardMode changes from Normal to Lines.

~If we decide that we don't want `control-w` `control-w` `control-y` to get you back where you started, I think the first few commits are probably worth having. I'm happy to clean them up for inclusion first if you prefer.~

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -264,4 +264,60 @@ mod test {
 
         assert_eq!(result, ReedlineEvent::None);
     }
+
+    #[allow(non_snake_case)]
+    mod integration {
+        use super::*;
+        use crate::{Editor, ReedlineEvent, UndoBehavior};
+        use pretty_assertions::assert_eq;
+        use rstest::rstest;
+
+        fn unwrap_edits(event: ReedlineEvent) -> Vec<EditCommand> {
+            match event {
+                ReedlineEvent::Edit(c) => c,
+                // It's not clear what the difference is between these, but vi mode often returns Multiple
+                ReedlineEvent::Multiple(events) => {
+                    events.into_iter().flat_map(unwrap_edits).collect()
+                }
+                other => panic!("unexpected event ${other}"),
+            }
+        }
+
+        #[rstest]
+        #[case::edit_dw_P(&["^", "dw", "P"])]
+        #[case::edit_3dw_P(&["^", "3dw", "P"])]
+        #[case::edit_d3w_P(&["^", "d3w", "P"])]
+        #[case::edit_dw_dw_P(&[
+            // duplicate the first word, because dw dw P should drop the first word
+            "^", "dw", "P", "P", "b",
+            // run the actual test, and it should put us back where we started.
+            "dw", "dw", "P"])]
+        #[case::edit_ddu(&["dd", "u"])]
+        fn sum_to_zero(#[case] commands: &[&str]) {
+            let initial_input = "the quick brown fox jumps over the lazy dog";
+            let keybindings = default_vi_normal_keybindings();
+            let mut vi = Vi {
+                insert_keybindings: default_vi_insert_keybindings(),
+                normal_keybindings: keybindings,
+                mode: ViMode::Normal,
+                ..Default::default()
+            };
+
+            let mut editor = Editor::default();
+            editor.set_buffer(initial_input.to_string(), UndoBehavior::CreateUndoPoint);
+
+            for command in commands {
+                let command: Vec<char> = command.chars().collect();
+                let parsed = parse(&mut command.iter().peekable());
+                let commands = unwrap_edits(parsed.to_reedline_event(&mut vi));
+                // FIXME: use run_edit_commands() instead
+                for command in commands {
+                    editor.run_edit_command(&command)
+                }
+            }
+
+            // FIXME: use Engine::current_buffer_contents()
+            assert_eq!(initial_input, editor.get_buffer());
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


amtoine:
@alsuren 
you have a bunch of unticked tasks, is this ready for review?

alsuren:
> you have a bunch of unticked tasks, is this ready for review?

Hrm. It's certainly not ready for merging (apart from maybe the first two commits). I will put it back in draft.

If anyone has any early comments on the general shape, or edge cases that I've missed (I'm sure there will be loads), I would still appreciate comments on those things.

sholderbach:
> it may be possible to write a proptest/fuzz test that generates a random buffer and a random position and a random set of cut commands, and

I would consider proptest and fuzzing to be outside the scope of the PR. Fuzzing as in just looking for broken invariants is independent, and proptests looking trying to emulate a ground truth also require some good design of those invariants to be worth the runs (they would be still appreciated but not necessary for this PR)

> follow `UX_TESTING.md` to make sure I've not broken anything.

The checklist may be quite out of date. Here the behavior in VI mode and probably also what happens when you undo/redo certain operations will be the most critical to me.

> double check whether each of the things should be a prepend or append operation. 

Not sure what you mean by that?

> should I actually check what readline's logic is?

We are not a GNU `readline` clone so you don't need to match edge case for edge case. We should make our users happy but that's about it.

>  If I read the source code, does my contribution here become a derived work?

I am not a layer :D

sholderbach:
@alsuren To help with the review of added tests I set up a coverage analysis job. If you merge main, you can see if particular lines are hit by our tests.

src/edit_mode/vi/mod.rs
```diff
@@ -264,4 +264,84 @@ mod test {
...
+        #[case::edit_d3w_P(&["^", "d3w", "P"])]
+        #[case::edit_dtz_P(&["^", "dtz", "P"])]
+        #[case::edit_dTz_P(&["dTz", "P"])]
+        #[case::edit_dfz_P(&["^", "dfz", "P"])]
+        #[case::edit_dFz_P(&["dFz", "P"])]
+        #[case::edit_dw_dw_P(&[
+            // duplicate the first word, because dw dw P should drop the first word
+            "^", "dw", "P", "P", "Fj",
+            // run the actual test, and it should put us back where we started.
+            "dw", "dw", "P"])]
```
alsuren, line 301: This rstest is me trying to implement something like https://matklad.github.io/2021/05/31/how-to-test.html#Data-Driven-Testing

I tried to avoid writing test cases in terms of internal data structures like EditCommand because they tend not to be refactor-proof. Vim commands are already text based, so they feel like a natural base for a testing microformat/DSL. Might be worth making an equivalent DSL for Emacs using the `C-w` notation from the Emacs/bash docs. Probably something for another PR though.

I think this particular example might be stretching the usefulness of the interface I built a bit though. A better way to write this test would probably be on top of a `check_commands(before, commands, after)` helper, probably using $0/$1 to encode the cursor position, as the rust-analyzer tests do. The rest of the `sum_to_zero` test cases could then be rebased on top of such an abstraction.

I'm not sure when I will have another opportunity for recreational hacking. It's been wiping me out recently. I am planning to take some time off in October, so I will have another stab at it then if I haven't finished it before.

alsuren:
I've switched back to using bash, and I've lost momentum on this.

I'm going to close this PR. I will leave the branch up in case anyone wants to cherry-pick from it.

sholderbach:
That is unfortunate to hear. Thanks for giving it a shot! This PR looked very promising, hopefully someone can take it as inspiration.

PR #632, opened 2023-09-09
make Control-w and alt-backspace behave more like bash

alsuren:
see https://github.com/nushell/nushell/pull/10288 for more context

From the commit messages:
```
    make control-w behave like bash
    
    'man bash' documents this as:
    
           unix-word-rubout (C-w)
                  Kill the word behind point, using white space as a word boundary.
                  The killed text is saved on the kill-ring.
    
    zsh does not have a concept of small/big words, but its 'word' is slightly
    bigger than nushell's small word, because it includes -./_ and a few other
    punctuation marks.
    
    fish's control-w is slightly larger than nushell's small word in that it
    seems to delete up to the next / (and doesn't need an extra control-w to
    delete the '/' 'thing/').

```

```
    make alt+backspace work in vscode by default
    
    On MacOS, vscode generates control-alt-h in response to alt+backspace.
    This is handled in bash as cut-smallword-left and in zsh as cut-word-left
    (zsh doesn't seem to have a concept of small/big words). Fish treats this
    as cut small word left (fish bigwords include anything up to /, and its
    small words stop on any punctuation).
```

As mentioned in my nushell PR, I realise that keybindings are the kind of thing that can cause holy wars, so I'm happy to back out any changes that people dislike.

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -50,7 +50,7 @@ pub fn default_emacs_keybindings() -> Keybindings {
         KC::Char('y'),
         edit_bind(EC::PasteCutBufferBefore),
     );
-    kb.add_binding(KM::CONTROL, KC::Char('w'), edit_bind(EC::CutWordLeft));
+    kb.add_binding(KM::CONTROL, KC::Char('w'), edit_bind(EC::CutBigWordLeft));
     kb.add_binding(KM::CONTROL, KC::Char('k'), edit_bind(EC::CutToEnd));
     kb.add_binding(KM::CONTROL, KC::Char('u'), edit_bind(EC::CutFromStart));
     // Edits
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -51,6 +51,11 @@ pub fn default_emacs_keybindings() -> Keybindings {
         edit_bind(EC::PasteCutBufferBefore),
     );
     kb.add_binding(KM::CONTROL, KC::Char('w'), edit_bind(EC::CutBigWordLeft));
+    kb.add_binding(
+        KM::CONTROL | KM::ALT,
+        KC::Char('h'),
+        edit_bind(EC::CutWordLeft),
+    );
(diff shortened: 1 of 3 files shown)


amtoine:
from https://github.com/nushell/nushell/pull/10288/files#diff-5c0320ebb54cd8ffd7059eb0178377c5139877d21457f6903439e7caf2e5b40eL585, looks like there might be an additional change to add here to be on par between the two PRs? :open_mouth:

amtoine:
okey, looks about right compared to https://github.com/nushell/nushell/pull/10288 :+1:

WindSoilder:
Currently in nushell, we think a word doesn't contains `-`, but other shell don't think so.
I tend to modify `word_left_index` function, make it behaves more like fish way(which moves up to /, and a word contains `-`)

https://github.com/nushell/reedline/blob/ebd653b8400837677343a6dec9e27a705e2f8f42/src/core_editor/line_buffer.rs#L262

alsuren:
> Currently in nushell, we think a word doesn't contains `-`, but other shell don't think so.

Would it be valuable to add a medium-word concept and leave it up to the end user to configure, or should nushell just "get it right" by default with no configuration?

(I think that making this change is out of scope for this pr, but if there is an easy way to avoid people carrying patches locally, it is worth exploring)

WindSoilder:
I think we can just change the definition of our "word", if we make it configurable, there are too much things we needs to do: should a word contains `-`, a word contains `_`.

I'd like to borrow the idea from fish (https://fishshell.com/docs/current/interactive.html#shared-bindings):
> Alt+← and Alt+→ move the cursor one word left or right (to the next space or punctuation mark), Alt+Backspace removes one word backwards.

I've played with fish, and found in fish, a word doesn't contain `_` either.

> Control+W removes the previous path component (everything up to the previous “/”, “:” or “@”) (moving it to the [Copy and paste (Kill Ring)](https://fishshell.com/docs/current/interactive.html#killring)).

alsuren:
I've switched back to using bash, and I've lost momentum on this.

I'm going to close this PR. I will leave the branch up in case anyone wants to cherry-pick from it.

Issue #480, opened 2022-09-16
Occasional panics with SQLite history


kubouch:
**Platform** Linux
**Terminal software** tmux

Describe the problem you are observing.

## Steps to reproduce
Happens with history set to sqlite.

Unfortunately, I don't have 100% repro steps but reedline seems to panic when I have one Nushell open, then open another one in a [Slurm](https://slurm.schedmd.com/overview.html) interactive session, work for a while, then try to bring up a history entry with the "up" key.

The panic happens at this line https://github.com/nushell/reedline/blame/70118f732c974fd953356020c6b3eb9b6ebe1ee0/src/engine.rs#L1007 where it says "todo: error handling". The expect() should be cleared up.

## Screenshots/Screencaptures

```
thread 'main' panicked at 'todo: error handling: ReedlineError(HistoryDatabaseError("SqliteFailure(Error { code: NotADatabase, extended_code: 26 }, Some(\"file is not a database\"))"))', /home/kubouch/.cargo/git/checkouts/reedline-e42026a78d91c510/70118f7/src/engine.rs:1007:14
```

kubouch:
Seems to be introduced with https://github.com/nushell/reedline/pull/401

kubouch:
It seems the history database somehow gets corrupted. When I try to `open-db` it, it shows an error "Not a SQLite file".

kubouch:
Trying to run `nu` with the corrupted database results in
```
Error:
  × error within history database: SqliteFailure(Error { code: NotADatabase, extended_code: 26 },
  │ Some("file is not a database"))
```

phiresky:
Is your home directory on a network drive by any chance? If so, which one and with which mount options. It should be pretty hard to corrupt an SQLite, the only instances I've myself had before were when on a network drive, opened  from multiple computers, with a NFS mount option such as `nolock`.

kubouch:
It was on a compute cluster (that I do not administer) that has many nodes so it is quite possible it was due to some mounting etc.

kubouch:
Since it might be hard to track down the cause, I think the panic should first be patched ASAP with some error handling, then we can move on to figuring out why the database got corrupted.

phiresky:
Could there also maybe be some corruption resulting due to threading or something within reedline? I'm not sure if rusqlite is really safe regarding that.

kronberger-droid:
panic cannot occur since #1173

PR #647, opened 2023-10-17
add cwd aware hinter

p00f:
towards https://github.com/nushell/nushell/issues/8883

### DIFF
--- a/src/hinter/cwd_aware.rs
+++ b/src/hinter/cwd_aware.rs
@@ -1,0 +1,94 @@
+use crate::{history::SearchQuery, Hinter, History};
+use nu_ansi_term::{Color, Style};
+
+/// A hinter that use the completions or the history to show a hint to the user
+///
+/// Similar to `fish` autosuggestins
+pub struct CwdAwareHinter {
+    style: Style,
+    current_hint: String,
+    min_chars: usize,
+}
+
+impl Hinter for CwdAwareHinter {
+    fn handle(
+        &mut self,
+        line: &str,
+        #[allow(unused_variables)] pos: usize,
+        history: &dyn History,
+        use_ansi_coloring: bool,
+    ) -> String {
+        self.current_hint = if line.chars().count() >= self.min_chars {
+            history
+                .search(SearchQuery::last_with_prefix_and_cwd(
+                    line.to_string(),
+                    history.session(),
+                ))
+                .expect("todo: error handling")
+                .get(0)
+                .map_or_else(String::new, |entry| {
+                    entry
+                        .command_line
+                        .get(line.len()..)
+                        .unwrap_or_default()
+                        .to_string()
+                })
+        } else {
+            String::new()
+        };
+
+        if use_ansi_coloring && !self.current_hint.is_empty() {
+            self.style.paint(&self.current_hint).to_string()
+        } else {
+            self.current_hint.clone()
+        }
+    }
+
+    fn complete_hint(&self) -> String {
+        self.current_hint.clone()
+    }
+
+    fn next_hint_token(&self) -> String {
+        let mut reached_content = false;
+        let result: String = self
+            .current_hint
+            .chars()
+            .take_while(|c| match (c.is_whitespace(), reached_content) {
+                (true, true) => false,
+                (true, false) => true,
+                (false, true) => true,
+                (false, false) => {
+                    reached_content = true;
+                    true
+                }
+            })
+            .collect();
+        result
+    }
+}
+
+impl Default for CwdAwareHinter {
+    fn default() -> Self {
+        CwdAwareHinter {
+            style: Style::new().fg(Color::LightGray),
+            current_hint: String::new(),
+            min_chars: 1,
+        }
+    }
+}
+
+impl CwdAwareHinter {
+    /// A builder that sets the style applied to the hint as part of the buffer
+    #[must_use]
+    pub fn with_style(mut self, style: Style) -> Self {
+        self.style = style;
+        self
+    }
+
+    /// A builder that sets the number of characters that have to be present to enable history hints
+    #[must_use]
+    pub fn with_min_chars(mut self, min_chars: usize) -> Self {
+        self.min_chars = min_chars;
+        self
+    }
+}
--- a/src/hinter/cwd_aware.rs
+++ b/src/hinter/cwd_aware.rs
@@ -1,19 +1,19 @@
-#[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
-use crate::{history::SearchQuery, Hinter, History};
-#[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
+use crate::{
+    history::SearchQuery,
+    result::{ReedlineError, ReedlineErrorVariants::HistoryFeatureUnsupported},
+    Hinter, History,
+};
 use nu_ansi_term::{Color, Style};
 
 /// A hinter that uses the completions or the history to show a hint to the user
-/// NOTE: Only use this with `SqliteBackedHistory`!
-/// Similar to `fish` autosuggestins
-#[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
+///
+/// Similar to `fish` autosuggestions
 pub struct CwdAwareHinter {
     style: Style,
     current_hint: String,
     min_chars: usize,
 }
 
-#[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
 impl Hinter for CwdAwareHinter {
     fn handle(
         &mut self,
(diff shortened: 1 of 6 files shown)


p00f:
I tried this out and it works - and I love the experience. This was the only thing I missed from fish.

Help with writing tests would be appreciated

fdncred:
Nice! It would be good to have an example that demonstrates its use in the examples folder. Can you add one of those please?

p00f:
> Nice! It would be good to have an example that demonstrates its use in the examples folder. Can you add one of those please?

I tried running the existing one (`cargo run --example hinter`) and don't actually see hints - do you want me to just copy-paste it into a new example and just change the constructor like so:

```rust

// Create a reedline object with in-line hint support.
// cargo run --example cwd_aware_hinter
//
// Fish-style cwd history based hinting
// assuming history ["abc", "ade"]
// pressing "a" hints to abc.
// Up/Down or Ctrl p/n, to select next/previous match

use nu_ansi_term::{Color, Style};
use reedline::{CwdAwareHinter, DefaultPrompt, Reedline, Signal};
use std::io;

fn main() -> io::Result<()> {
    let mut line_editor = Reedline::create().with_hinter(Box::new(
        CwdAwareHinter::default().with_style(Style::new().italic().fg(Color::LightGray)),
    ));
    let prompt = DefaultPrompt::default();

    loop {
        let sig = line_editor.read_line(&prompt)?;
        match sig {
            Signal::Success(buffer) => {
                println!("We processed: {buffer}");
            }
            Signal::CtrlD | Signal::CtrlC => {
                println!("\nAborted!");
                break Ok(());
            }
        }
    }
}
```

fdncred:
I think so, but you'll need to add some items to the history in order to get hints to work.

p00f:
right - how do I change the cwd inside the repl though? One idea is to alternate between original cwd and `std::env::home_dir()` on every iteration.

p00f:
How does the example look? I tested the example and it works as intended.

p00f:
I think I need help writing tests as I'm not familiar with the code too much and there are no tests for the existing`DefaultHinter`

p00f:
On the nushell side I just use `DefaultHinter` when the file format is `FileBackedHistory` (I'm assuming `engine_state` is updated at runtime, so the behaviour is that it is just not used)

p00f:
https://paste.sr.ht/~p00f/5752fc99ba31990710f08c2a85de897a5b15bf54#cwd.diff-L48

@sholderbach this is my nushell patch

p00f:
Note that this causes no undesirable behaviour - if you change the history format you don't get hints from the other history file anyway

p00f:
> (I'm assuming engine_state is updated at runtime, so the behaviour is that it is just not used)

this is correct - the `Hinter` is created [inside](https://github.com/nushell/nushell/blob/5d8763ed1d14048ac9fc09eb21be36e6e622f0d7/crates/nu-cli/src/repl.rs#L302) the [REPL loop](https://github.com/nushell/nushell/blob/5d8763ed1d14048ac9fc09eb21be36e6e622f0d7/crates/nu-cli/src/repl.rs#L193)

sholderbach:
My caution is driven by the fact that we have a [substantial number of non-nushell users](https://github.com/nushell/reedline/network/dependents) of reedline. And since we introduced the new history search API (#401) we had a few panics based on the capabilities or assumptions what info is accessible (e.g. #480).

I would like to avoid a situation where you would encounter runtime panics when the `Hinter`/`History` combination gets used.

Things should be either abundantly clear from documentation, failure should happen either at the earliest point possible or handled gracefully.

> [paste.sr.ht/~p00f/5752fc99ba31990710f08c2a85de897a5b15bf54#cwd.diff-L48](https://paste.sr.ht/~p00f/5752fc99ba31990710f08c2a85de897a5b15bf54#cwd.diff-L48)

This would bring every user of `file_format: "sqlite"` over to the behavior that prioritizes hints for the current directory.
Not sure if this is universally liked. So there might be someone proposing a decoupling between the two down the line.

p00f:
I see - there's no way to find out the type of `History`, the trait has no such methods. Do I create a `HistoryType` enum and the implement such a method? If I do end up doing this, why not just do away with dynamic dispatch and use enums instead?

imo I should just document that it should always be used with `SqliteBackedHistory`.

>  Not sure if this is universally liked. So there might be someone proposing a decoupling between the two down the line.

If they do, I'll add a config option. At least for now, I think everyone will love this - fish has this by default without a config option, nobody complains.

fdncred:
> I think everyone will love this ...

Personally, I pass on this behavior. I'll play with it and test it out, but I don't think I'll be using it. Let's make it configurable. 

I think it's great to have it, and there's no doubt that some people will love it. However, I've worked on nushell long enough to know that you will never please everyone. 

On a separate note, maybe I'll change my nick from "fdncred" to "everyone" then we can list the features that 'everyone' loves. 🤣

p00f:
> I'll play with it and test it out.

of course, I haven't created a nushell PR yet.

> On a separate note, maybe I'll change my nick from "fdncred" to "everyone" then we can list the features that 'everyone' loves. 🤣

Please do, I hate when people ping `@everyone` on discord - this way it won't actually ping everyone 😛

fdncred:
> Please do, I hate when people ping @everyone on discord - this way it won't actually ping everyone.

LOL - me too. @here and @everyone kind of drive me crazy on discord.

src/hinter/cwd_aware.rs
```diff
@@ -0,0 +1,100 @@
+#[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
+use crate::{history::SearchQuery, Hinter, History};
+#[cfg(any(feature = "sqlite", feature = "sqlite-dynlib"))]
+use nu_ansi_term::{Color, Style};
+
+/// A hinter that uses the completions or the history to show a hint to the user
+/// NOTE: Only use this with `SqliteBackedHistory`!
```
p00f, line 7: @sholderbach How does this look?

sholderbach:
Briefly, I don't think just documenting it is a good approach for the public API.

With a `FileBackedHistory` things can be compiled, the builder executes and then things panic on the first keypress.
<details>
<summary>Panicking patch</summary>

```
cargo run --features sqlite --example cwd_aware_hinter
```

```patch
diff --git a/examples/cwd_aware_hinter.rs b/examples/cwd_aware_hinter.rs
index 17aa7dd..35f9c2f 100644
--- a/examples/cwd_aware_hinter.rs
+++ b/examples/cwd_aware_hinter.rs
@@ -49,16 +49,14 @@ fn main() -> io::Result<()> {
         #[allow(deprecated)]
         let home_dir = std::env::home_dir().unwrap();
 
-        let history = create_filled_example_history(
-            &home_dir.to_string_lossy().to_string(),
-            &orig_dir.to_string_lossy().to_string(),
-        );
+        let history = reedline::FileBackedHistory::new(100); 
+
 
         let mut line_editor = Reedline::create()
             .with_hinter(Box::new(
                 CwdAwareHinter::default().with_style(Style::new().italic().fg(Color::Yellow)),
             ))
-            .with_history(history);
+            .with_history(Box::new(history));
 
         let prompt = DefaultPrompt::default();
``` 
</details>

Either we replace the current `.expect()`s in a way we can implement fall backs through recoverable error handling or we block at construction in `.with_hinter`/`.with_history` and provide a recoverable error there.

### fallback/local error handling
As `History::search` already returns a result turning the expect here into a fall-back to a simpler query would be an option.
https://github.com/p00f/reedline/blob/00c59e4ca2762d5a12579b68de5dd505318fbfc4/src/hinter/cwd_aware.rs#L31

### check at construction time

You are right that trait objects are limited on its own, but for "RTTI we have at home" the cost of adding a method that advertises a particular capability doesn't sound too bad if it is only checked in the constructors.

**P.S.:** just to disagree with @fdncred I really like this feature as a default and would be ok with saving ourselves another config option (just strawmanning it in for API design purposes)

p00f:
Sorry, it took me a while to realize how "just don't use it" sounds.

I went with the first option as the second one requires more thought. Thanks for the help!

p00f:
re: config option, we can introduce this without an option on `main`, then add it if people want to disable it. There's some time before the next release.

p00f:
@sholderbach i fixed the warning causing the CI failure, can you re-run it?

sholderbach:
Thanks for implementing the fallback! 

> re: config option, we can introduce this without an option on `main`, then add it if people want to disable it. There's some time before the next release.

Agree, your nushell patch looked already pretty good.

sholderbach (approved):

p00f:
sorry, I think this one should pass

p00f:
should I squash?

sholderbach:
No need for manual squashing, we happily use the Github button for that.

Thanks for improving the quality of the hints. Looking forward to using them in Nushell

p00f:
How do I do the nushell PR? Uncommenting https://github.com/nushell/nushell/blob/d0dc6986dd14e7ea5443f3dc0eb5b101c5de5efc/Cargo.toml#L167 ?

sholderbach:
Yup uncommenting that line and running `cargo update -p reedline` to guarantee you get the most recent version on that branch

p00f:
Yeah - I was wondering if it was acceptable to commit into main, then I read the diff - it was uncommented in `main` when I started working  😄

Issue #570, opened 2023-04-13
No difference between "cutwordleft" and "cutbigwordleft"


ineu:
From the discussion: https://github.com/nushell/nushell/discussions/8859

I'm trying to remove the last part of the long_token_separated_by_underscores. For this purpose I tried binding the actions in question:
```
    {
      name:     delete_word
      modifier: control
      keycode:  char_g
      mode:     emacs
      event:    { edit: cutwordleft }
    }
```
The same for `cutbigwordleft`.

Expected result: 
- `cutwordleft` removes text to the previous underscore and/or a capital letter
- `cutbigwordleft` removes the whole token

Actual result:
- both behave the same and remove the whole token

kronberger-droid:
fixed with new resolver #1100

Issue #563, opened 2023-03-30
Incorrect word boundary in VI mode


morzel85:
**Platform** Ubuntu 20.04
**Terminal software** gnome-terminal
**Version** Reedline from Nushell 0.77.0

The _word_ boundary (used in `b` and `e`) allows too many characters.

In VI mode a _word_ should "_consists of sequence of letters, digits and underscores_" according to VIM docs, this is also how classic Readline works.

## Steps to reproduce
1. Run Reedline in VI mode.
2. Type _abc.def_ or _abc:def_
3. Use `b` and `e` motions and notice that the dot and colon are treated as parts of _word_

mjs:
I've had a look at this as I keep getting caught out by `dw` and `cw` not doing what I expect in nushell :grin:. I'm totally new to the reedline code base and am relatively new to Rust so bear with me.

The way `unicode-segmentation` splits words is super nuanced. It might be a shame to lose the "correct" handling of unicode words that it does.  It also doesn't look like there's any way to influence `unicode-segmentation`'s word splitting behaviour from the outside.

I can see a few possible solutions to fixing word splitting in reedline to better match what Vi and Emacs do:

1. Continue using `unicode-segmentation` for word splitting but then add an extra step which splits the words it generates further by looking through them for additional characters we want to split on. This is a bit clunky but at least we wouldn't be throwing away all the subtle handling of unicode that `unicode-segmentation` does. This might work because `unicode-segmentation`'s concept of words tends to be broader than what we want in reedline.
2. Implement independent word splitting in reedline by walking through the line's graphemes and applying our own rules. Treat all/most characters beyond ASCII as word characters. This should work fine in many cases but there's probably all sorts of unicode edge cases we'd be ignoring that would cause odd behaviour for people not working in straight ASCII.
3. Vendor/fork `unicode-segmentation`  and add support for customising word splitting.

I'm tempted to try option 1 first as it's the least invasive but I'm open to other ideas / dissenting opinions.

sholderbach:
Thanks for looking into it!

Yeah 1.) seems pretty reasonable as a starting point. The vim documentation feels a bit sparse to parse out the rules for a 2.) that would behave sensibly with more complex Unicode characters/graphemes. 
With a bit of effort this should be doable as well, but `unicode-segmentation` has certainly done some of the heavy lifting already. We certainly don't need to do all the look-up-table work `unicode-segmentation` has done for ultimate performance.

kronberger-droid:
fixed in #1100

PR #650, opened 2023-10-28
Split hinter tokens at Unicode word boundaries

stfacc:
Partial completion behaves now like usual word movements by C-Left and C-Right.

This is also useful to accept only parts of a suggested path.

### DIFF
--- a/src/hinter/cwd_aware.rs
+++ b/src/hinter/cwd_aware.rs
@@ -1,14 +1,10 @@
 use crate::{
+    hinter::get_first_token,
     history::SearchQuery,
     result::{ReedlineError, ReedlineErrorVariants::HistoryFeatureUnsupported},
     Hinter, History,
 };
 use nu_ansi_term::{Color, Style};
-use unicode_segmentation::UnicodeSegmentation;
-
-pub fn is_whitespace_str(s: &str) -> bool {
-    s.chars().all(char::is_whitespace)
-}
 
 /// A hinter that uses the completions or the history to show a hint to the user
 ///
@@ -68,22 +64,7 @@ impl Hinter for CwdAwareHinter {
     }
 
     fn next_hint_token(&self) -> String {
-        let mut reached_content = false;
-        let result: String = self
-            .current_hint
-            .split_word_bounds()
-            .take_while(|word| match (is_whitespace_str(word), reached_content) {
-                (_, true) => false,
-                (true, false) => true,
-                (false, false) => {
-                    reached_content = true;
-                    true
-                }
-            })
-            .collect::<Vec<&str>>()
-            .join("")
-            .to_string();
-        result
+        get_first_token(&self.current_hint)
     }
 }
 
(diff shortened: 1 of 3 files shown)


amtoine:
i'm not familiar with Reedline but for people that might want to review this PR, could you add a bit more information? :innocent: 

1. is this addressing a bug?
2. how to test the fix / the feature?

stfacc:
Sure, the main issue for me is the following:

* Suppose you typed this command
`> ls /some/very/long/path[ ]`
where `[]` indicates the cursor. Hitting C-Left and C-Right you can move between all elements of the path, one step at a time, for instance
`> ls /some/[v]ery/long/path`
* Now you start typing `ls` at a new prompt and you get
`> ls[ ]/some/very/long/path`
where everything after the cursor is the completion hint.
After hitting C-Right **once**, the entire path is now accepted:
`>ls /some/very/long/path[ ]`
This is different from the above situation and, at least for me, unexpected: what I usually want is to accept only part of the suggested path and modify e.g. the final elements.

This happens because currently the next history token is identified using only whitespace, while movements in `src/core_editor/line_buffer.rs` take into account Unicode words.

* A similar problem happens with long pipelines without spaces around `|`, which is common in interactive usage. These two history elements behave differently:
`> some|very|long|pipe`
`> some | very | long | pipe`


As additional context, the proposed changed is the mostly the same as `fish` history completion.

stfacc:
A few issues related to this one:

* https://github.com/nushell/reedline/issues/563
https://github.com/nushell/reedline/issues/570
Here it's pointed out that `.`, `:`, `_` should probably also be used as word separators.

* https://github.com/nushell/nushell/pull/10288
Points out that in `fish`, separators are included in the token. With current `unicode_segmentation` implementation each separator is its own word, which is not super useful: in the path example above, the cursor would stop at each `/`

amtoine:
thanks @stfacc :pray: :blush:

sholderbach (commented):
Thanks for tackling this. Sounds like a good possible quality of life improvement.

Looks pretty good based on how we currently deal with words in `src/core_editor/line_buffer.rs`.

I think it would be great if we could reduce the copy pasted code a bit here. I think we can make the `is_whitespace_str` a crate level helper, and maybe you could figure out a good place to share the word advancing for the hinters.

https://github.com/nushell/reedline/blob/973dbb5f5f2338c18c25ce951bfa42c8d8cacfdf/src/core_editor/line_buffer.rs#L751-L754

stfacc:
Hi,
I moved the helper functions to the base crate.

Since this is basically the same logic used for the LineBuffer, would it be useful to refactor even more and share the same code?
Not sure how easy. But could save some duplication if we added more logic to tackle the issues pointed out in the other PRs.

sholderbach (approved):
Thanks for the refactor! Looks good in testing. We can refine the word boundaries later. (e.g. `..` becomes two words, but that is something specific to e.g. Nushell where `..` are expected for paths or ranges)

PR #652, opened 2023-10-31
enable paste buffer to be changed before pasting

fdncred:
This PR is just a thought starter. What would it be like to be able to paste data into reedline and have it manipulate that data?

In this example, I paste some shell script syntax and it's changed to nushell syntax.

### Input from clipboard:
```
ENV_VAR1=var1
export ENV_VAR2=var2
line one \
line two \
line three \
line four
```
### Output in reedline command buffer 
(note the different env var syntax and the line continuation characters are removed)
image

I was thinking it would be cool to have some type of hook/callback where in nushell one could provide a custom command/script/module to manipulate data that is pasted. I'm not even sure that it's possible, but like I said, this is a thought starter.

I do not expect this PR to land. Just wanted to use it as an example.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -421,9 +421,9 @@ dependencies = [
 
 [[package]]
 name = "memchr"
-version = "2.5.0"
+version = "2.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


fdncred:
This was fun and has potential but there doesn't seem to be anyone who's interested right now. So, I'll just cloes this for now.

PR #651, opened 2023-10-30
Avoid consuming CPU when waiting for input.

sunfishcode:
When reading input, start by calling `event::read`, which efficiently blocks until input arrives, rather than polling for input every 100 milliseconds.

This preserves the existing `POLL_WAIT` polling behavior by moving the poll after the first `read` call returns.

Fixes #521.

PR #655, opened 2023-11-02
Avoid waiting 10ms after normal key events

### LINKED ISSUE #643: POLL_WAIT hurts user experience
**Platform** macOS
**Terminal software** Apple Terminal.app

- [Original bug report I filed against Nushell](https://github.com/nushell/nushell/issues/10484)
- [Test script](https://gist.github.com/comex/380653fefb3312b24da5d66ed5ee5be0) to measure Nushell's typing latency (runs on macOS or Linux)

To quote `src/engine.rs`:
```rust
// The POLL_WAIT is used to specify for how long the POLL should wait for
// events, to accelerate the handling of paste or compound resize events. Having
// a POLL_WAIT of zero means that every single event is treated as soon as it
// arrives. This doesn't allow for the possibility of more than 1 event
// happening at the same time.
const POLL_WAIT: u64 = 10;
```

In effect this adds 10ms of latency to every keystroke, by design.  But this hurts the user experience for two reasons.

- First, 10ms is a human-perceivable amount of lag by itself.  On my 120Hz display, 10ms will delay the output by 1-2 frames depending on how far into the frame the keystroke was received.  On a 60Hz display it would delay the output by 0-1 frames.  10ms is really not the end of the world, but when you add that to other sources of latency such as display latency and SSH connection latency, it starts to add up.

- Second, there is a [bug in Apple's Terminal.app](https://github.com/nushell/nushell/issues/10484#issuecomment-1732446649) where if a terminal program takes more than 1ms to respond to keystrokes, the output is delayed by an additional ~33ms.  This is not `reedline`'s fault, and I have filed a bug report with Apple about it, but removing the artificial latency significantly mitigates the bug.  I tried setting `POLL_WAIT` to 0 and running Nushell, and it often is capable of responding within the 1ms deadline.

Still, the justification given for `POLL_WAIT` makes sense: it should improve performance when there's a very fast stream of events such as with pasted text.  I suggest making the wait time dynamic.  When receiving the first few events after a period of inactivity, handle them instantly.  Then, if events continue to pile up quickly (more quickly than a human can type), start adding a delay.  (If needed, resize events could also be treated specially and always given the delay.)  That should preserve most of the benefit of the `POLL_WAIT` scheme, while eliminating the latency for normal typing.

Alternately, `POLL_WAIT` could be made runtime configurable.

sunfishcode:
Avoid waiting 10ms after normal key events, so that we process normal key events quickly when they arrive.

To handle pastes and resize bursts, introduce a "busy" state. When we see events immediately ready after an event, assume it's a paste, and enter the "busy" state. Similarly, after a resize, enter the "busy" state. When the paste or resize burst appears to be done, leave the "busy" state and resume processing normal key events quickly. See the comments in the code for more details.

Also, bump `POLL_WAIT` to 100ms. The previous change means we're not waiting for `POLL_WAIT` after normal keyboard input events, so we can afford to make this a little longer, and empirically, 100ms seems to be needed to avoid processing events during a resize burst on my machine.

And, remove the `EVENTS_THRESHOLD`, which now apppears to be redundant.

Fixes #643.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -774,12 +774,10 @@ impl Reedline {
                         busy = false;
                         break;
                     }
+                } else if event::poll(Duration::from_millis(0))? {
+                    busy = true;
                 } else {
-                    if event::poll(Duration::from_millis(0))? {
-                        busy = true;
-                    } else {
-                        break;
-                    }
+                    break;
                 }
             }
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -47,16 +47,13 @@ use {
 };
 
 // The POLL_WAIT is used to specify for how long the POLL should wait for
-// events, to accelerate the handling of paste or compound resize events. Having
-// a POLL_WAIT of zero means that every single event is treated as soon as it
-// arrives. This doesn't allow for the possibility of more than 1 event
-// happening at the same time.
-const POLL_WAIT: u64 = 10;
-// Since a paste event is multiple Event::Key events happening at the same time, we specify
-// how many events should be in the crossterm_events vector before it is considered
-// a paste. 10 events in 10 milliseconds is conservative enough (unlikely somebody
-// will type more than 10 characters in 10 milliseconds)
-const EVENTS_THRESHOLD: usize = 10;
+// events when we're "busy", that is, when we think we're in the middle of a
+// paste or a resize burst, so that we don't do lots of work painting output
+// that's about to be discarded.
+//
+// The value 100 here was chosen by looking at how quickly window a burst of
+// resize events coming from a window manager seem to arrive in practice.
+const POLL_WAIT: u64 = 100;
 
 /// Determines if inputs should be used to extend the regular line buffer,
 /// traverse the history in the standard prompt or edit the search string in the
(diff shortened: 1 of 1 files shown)


sunfishcode:
This is a more subtle change than #651, and will take more care to confirm that it works reliably for everyone.

So far, I've tested this on Linux a few different terminals. My understanding of the platform APIs used by `crossterm` on macOS and Windows is that this should work on them too, however I have not tested them.

fdncred:
What's the best way to test this. I have a Mac and can help test.

fdncred:
I tried pasting 16 lines and it's super-fast and worked great. However, when I pasted 165 lines the last 31 lines showed up in the repl and lines before that disappeared. Also, I couldn't get out of the paste mode. I hit return about 10 times and still had the multiline indicator. Seems like it wouldn't accept the line.

sunfishcode:
Yeah, for testing, the main things to test are: simple typing, small and large (> 1024 bytes, which is crossterm's buffer size) pastes, and window resizing. Ideally they should all feel fast.

In theory, the PR here is just changing the timing when events are processed and when repaints happen. It's possible that this is uncovering a pre-existing bug somewhere.

fdncred:
Any more thoughts on this @sunfishcode?

Feel-ix-343:
Any updates here?

sunfishcode:
No; I'm not likely to get back to this for a while now, so I'll close this. If anyone else wants to continue this, feel free to.

PR #656, opened 2023-11-03
CwdAwareHinter: remove cwd filter when there are no results

p00f:
closes https://github.com/nushell/nushell/issues/10921

### DIFF
--- a/src/hinter/cwd_aware.rs
+++ b/src/hinter/cwd_aware.rs
@@ -24,7 +24,7 @@ impl Hinter for CwdAwareHinter {
         use_ansi_coloring: bool,
     ) -> String {
         self.current_hint = if line.chars().count() >= self.min_chars {
-            history
+            let with_cwd = history
                 .search(SearchQuery::last_with_prefix_and_cwd(
                     line.to_string(),
                     history.session(),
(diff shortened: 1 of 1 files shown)


src/hinter/cwd_aware.rs
```diff
@@ -39,15 +39,29 @@ impl Hinter for CwdAwareHinter {
...
-                .expect("todo: error handling")
-                .get(0)
-                .map_or_else(String::new, |entry| {
-                    entry
-                        .command_line
-                        .get(line.len()..)
-                        .unwrap_or_default()
-                        .to_string()
-                })
+                .unwrap_or_else(|_| vec![]);
```
sholderbach, line 42: Good call! That should eliminate a possible failure path! And yeah having no hint wouldn't be a problem if there is an error internally.

As `Vec` should implement `Default` you could also write

```suggestion
                .unwrap_or_default();
```

sholderbach (approved):
Nice!

src/hinter/cwd_aware.rs
p00f: Thanks, done

p00f:
I just saw that the next release is tomorrow so I wanted to get this in

sholderbach:
Tested it with the `cwd_aware_hinter` demo and works like a charm. 

As we made it compulsory for the sqlite history let's get this qol improvement in.

PR #657, opened 2023-11-05
re-enable bracketed paste on not(windows) (fixes #9944) (fixes #648)

### LINKED ISSUE #648: bracketed paste only works the first time
**Platform** Ubuntu 22.04
**Terminal software** gnome-terminal

bracketed paste only works the first time. 

This bug does not exist on v0.19.1 but does on v0.25.0

## Steps to reproduce

1. run `cargo run --example demo`
2. paste multi-line content, successful, press `ctrl+c`, paste multi-line content again, it failed

## Screenshots/Screencaptures

reedline

LevitatingBusinessMan:
Due to a supposed issue (https://github.com/nushell/nushell/issues/9218) in nushell regarding commands that read standard input, a fix (https://github.com/nushell/nushell/pull/9399) was proposed that temporarily disables bracketed_paste during the execution of [eval_source](https://github.com/WindSoilder/nushell/blob/cf7040a215fa8d4008844f7fefc2abacc7482546/crates/nu-cli/src/util.rs#L205-L230). 

Instead, this issue was moved over to reedline and a new PR was created (https://github.com/nushell/reedline/pull/598). In this version of the "fix", bracketed_paste is disabled completely after the first execution of reedline. This has led to issue https://github.com/nushell/reedline/issues/648, where multiline paste is now being disabled on non-windows environments.

This PR effectively reverts https://github.com/nushell/reedline/pull/598. This might cause https://github.com/nushell/nushell/issues/9218 to reappear in nushell.

I have personally not been able to reproduce https://github.com/nushell/nushell/issues/9218,  modern versions of gpg use a GUI prompt for the password, or alternatively a curses window. I did try to replicate the issue using GNU readline and other prompting methods but I wasn't succesful.

If https://github.com/nushell/nushell/issues/9218 is a reproducible issue, I think the best course of action is to merge https://github.com/nushell/nushell/pull/9399, which supposedly fixes the issue, but on the nushell side. It seems like this issue might be specific to nushell, and thus it should be fixed in that repository. No need to affect all other reedline dependent packages with it.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -630,9 +630,6 @@ impl Reedline {
 
         let result = self.read_line_helper(prompt);
 
-        #[cfg(not(target_os = "windows"))]
-        self.disable_bracketed_paste()?;
-
         if self.use_kitty_protocol {
             let _ = execute!(io::stdout(), event::PopKeyboardEnhancementFlags);
         }


src/engine.rs
```diff
@@ -630,9 +630,6 @@ impl Reedline {
 
         let result = self.read_line_helper(prompt);
 
-        #[cfg(not(target_os = "windows"))]
```
fdncred, line 633: I kind of wonder if this was originally just a typo and should've really been `#[cfg(target_os = "windows"))]` because bracketed paste doesn't work in windows since crossterm doesn't support it yet.

LevitatingBusinessMan: The author of the original issue was using Linux. Though he was using the Windows Terminal (using debian via WSL2).

Thus it might be a Windows Terminal issue? I don't think anyone tried it to reproduce it with Windows Terminal specifically. And it seems nobody was actually able to reproduce it accept #theAkito himself.

theAkito: I can try to reproduce with [WezTerm](https://wezfurlong.org/wezterm/) on Windows. Just tell me which commits to try & any additional useful information...

LevitatingBusinessMan:
@theAkito 
> I can try to reproduce with [WezTerm](https://wezfurlong.org/wezterm/) on Windows. Just tell me which commits to try & any additional useful information...

I think the easiest way to test it would be to compile nushell with:
```toml
[patch.crates-io]
reedline = { git = "https://github.com/LevitatingBusinessMan/reedline.git", branch = "re-enable-bracketed-paste" }
```

And then try to reproduce the error. Have you seen this error only with gpg or also other commands that take input?

theAkito:
> And then try to reproduce the error.

What's the point of trying to reproduce it in Windows Terminal? If it's gonna happen again, it still could be due to this terminal.

LevitatingBusinessMan:
> If it's gonna happen again, it still could be due to this terminal.

Exactly. You can first verify the bug is still present on Windows Terminal. And then you can remove variables to find the root cause. Like if it is related to the terminal by switching to another.

You can also test for the bug in WezTerm first, but if that fails you'd have to test it in Windows Terminal anyway.

I am curious if the bug is specific to gpg somehow. If it isn't than I think more people would experience it. If the bug does appear in gpg try using passwd.

If it is specific to gpg than it must be related to [pinentry](https://www.gnupg.org/related_software/pinentry/index.html).

fdncred:
also related https://github.com/nushell/nushell/issues/9944

LevitatingBusinessMan:
I believe https://github.com/nushell/nushell/issues/9944 is actually fixed by this PR.

@amtoine is on Linux so it makes sense.

theAkito:
> Exactly. You can first verify the bug is still present on Windows Terminal. And then you can remove variables to find the root cause. Like if it is related to the terminal by switching to another.
> 
> You can also test for the bug in WezTerm first, but if that fails you'd have to test it in Windows Terminal anyway.

Makes sense. Will try, when I get the time.

Am following this pull request.

WindSoilder:
Thanks, generally I'm ok with your point.

I believe it will cause https://github.com/nushell/nushell/issues/9218 to reappear.  But we can bring https://github.com/nushell/nushell/pull/9399 back, so both reedline and nushell are useful to us again.

theAkito:
> You can also test for the bug in WezTerm first, but if that fails you'd have to test it in Windows Terminal anyway.

Done, did it.

Here is what I did, so there are no misunderstandings.

1. Compiled Nushell `master` branch on WSL: `cargo build --release --workspace; cargo run --release`
2. Testing GPG signing, which requires the passphrase. `gpg --yes -s s.txt`
3. Same issue, as in https://github.com/nushell/nushell/issues/9218.
4. Connected to WSL via [WezTerm](https://wezfurlong.org/wezterm/).
5. Same issue, once again.



> I am curious if the bug is specific to gpg somehow. If it isn't than I think more people would experience it. If the bug does appear in gpg try using passwd.
> 
> If it is specific to gpg than it must be related to [pinentry](https://www.gnupg.org/related_software/pinentry/index.html).

There are probably tons of other programs requiring password entry in a similar fashion, but right now none come to my mind. If you can tell me an example program, I can test, I will test the password entry there.

theAkito:
Tried `passwd` & works perfectly fine without issues.

These are the literal characters presented to me on `pinentry`.

```
                           ┌────────────────────────────────────────────────────────────────┐
                           │ Please enter the passphrase to unlock the OpenPGP secret key:  │
                           │ "Name <mail@mail.com>"                                         │
                           │ 4096-bit RSA key, ID HEXHEXHEXHEXHEHX,                         │
                           │ created 2001-11-11.                                            │
                           │                                                                │
                           │                                                                │
                           │ Passphrase: __________________________________________________ │
                           │                                                                │
                           │         <OK>                                    <Cancel>       │
                           └────────────────────────────────────────────────────────────────┘
```

GPG version

```
gpg (GnuPG) 2.2.27
libgcrypt 1.8.8
Copyright (C) 2021 Free Software Foundation, Inc.
License GNU GPL-3.0-or-later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

fdncred:
Are we saying this PR is ready to land or is it going to create other problems?

LevitatingBusinessMan:
Thanks to @theAkito I can now confirm that this issue is specific to curses.

I was able to replicate it with a password prompt written in python (with curses).

```PYTHON
import curses

def main():
    # Initialize curses
    stdscr = curses.initscr()

    # Prompt for password
    stdscr.addstr("Paste text using bracketed_paste: ")
    password = ""
    while True:
        key = stdscr.getch()

        password += chr(key)

        stdscr.clear()  # Clear screen
        stdscr.addstr("Paste text using bracketed_paste: " + password)

if __name__ == "__main__":
    main()
```

Escape characters `200` and `201` are prefixed and suffixed respectively. Curses captures the key and converts it to a character, and apparently nushell is sending some escape characters that get converted like this.

LevitatingBusinessMan:
I personally think this PR should be merged, so #9944 and #648 can be closed.

After this is merged, I will open a new issue at nushell, covering this issue with bracketed paste and curses.

fdncred:
ok @LevitatingBusinessMan let's try your solution. I'll be looking for that nushell PR. Appreciate it.

theAkito:
> I was able to replicate it with a password prompt written in python (with curses).

Weird. Cannot reproduce it in Nim.

```nim
##[
  Works on Nim 2.0.0

  nimble install ncurses # https://github.com/walkre-niboshi/nim-ncurses
  nim c -r reedline_657.nim ; rm reedline_657
]##

import pkg/ncurses

when isMainModule:
  let stdscr = initscr()
  stdscr.waddstr("Paste text using bracketed_paste: ")
  var password = ""
  while true:
    var key = stdscr.wgetch()

    if key == 0: break

    password &= key.chr

    stdscr.werase
    stdscr.waddstr(cstring("Paste text using bracketed_paste: " & password.repr))
```

Output just shows the characters I entered, which is, I suppose, the correct behaviour.

This program's `ncurses` library uses `libncursesw.so(.5|.6|)` in the background, so it's the behaviour I get from my system's `ncurses` library.

LevitatingBusinessMan:
@theAkito Did you test is using reedline compiled from git? Including this merge?

theAkito:
> @theAkito Did you test it using reedline compiled from git? Including this merge?

I tested with the product of yesterday's compilation, where I did reproduce it with GPG, earlier.

LevitatingBusinessMan:
I can reproduce it just fine, by having `reedline` pointed to the git's main branch. And doing `cargo update` to update the dependencies.

PR #658, opened 2023-11-05
don't allow certain examples to compile without their features

fdncred:
This PR forces examples that need certain features to nag about needing those features and not compile without specifying them.

image


amtoine:
that's pretty cool :yum:

PR #659, opened 2023-11-08
Properly handle optional event modes

sholderbach:
Both bracketed paste and the kitty keyboard enhancement flag are not always supported.
This goes both for terminals that may not emit events according to the extensions thus we already provide the means to only optionally enable them.
But also the applications that may run between each `Reedline::read_line` call are in no way guaranteed to supported any enhanced terminal events coming in.
It should thus be part of reedline's contract to leave `read_line` in a clean state with the canonical terminal behavior.

The API for `Reedline::enable_bracketed_paste` was misleading as it for the most part acted independent of the `Reedline` state and immediately globally enabled bracketed paste, potentially polluting any application during evaluation. `disable_bracketed_paste` equally globally modified state but was only active if the bracketed paste state was entered with `enable_bracketed_paste` and would only run once.

For better readability transfer this pattern to the kitty keyboard enhancement protocol. (which was already properly handled but was overly eager to check if the enhancements are supported on every `read_line`)

Make sure we enter the mode when starting to take control and exit the mode when leaving the mode or being dropped.
The settings exposed to the user will just internally enable a setting (in the case of the kitty keyboard enhancement flags crossterm provides an easy way to check for support that we take into account with that)

Close https://github.com/nushell/nushell/issues/10982 and related issues for non-nushell projects using reedline

As a follow up we can simplify the setters to ignore the `Result` or collapse it to a simpler `fn use_...(self, enable: bool)`

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -88,6 +88,7 @@ fn main() -> std::io::Result<()> {
         .with_partial_completions(true)
         .with_cursor_config(cursor_config)
         .use_bracketed_paste(true)
+        .use_kitty_keyboard_enhancement(true)
         .with_highlighter(Box::new(ExampleHighlighter::new(commands)))
         .with_hinter(Box::new(
             DefaultHinter::default().with_style(Style::new().fg(Color::DarkGray)),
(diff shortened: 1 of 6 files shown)


fdncred:
Nice work @sholderbach!!

fdncred:
are we going to land this before the release or after?

PR #661, opened 2023-11-11
Make `Result` and `ReedlineError` public

### LINKED ISSUE #545: Cannot implement custom History: `ReedlineError` is private
Currently you can't implement your own `History` type as multiple trait functions return a `Result<_, ReedlineError>`, but `ReedlineError` (which comes from the private `result` module) is not available to use outside of the crate.

Would it be possible to make it available publicly?

Thanks!

ClementNerma:
Allows implementing custom `History` types

Closes #545

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -239,7 +239,7 @@ mod engine;
 pub use engine::Reedline;
 
 mod result;
-pub use result::{Result, ReedlineError};
+pub use result::{ReedlineError, Result};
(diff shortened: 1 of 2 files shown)


fdncred:
Seems reasonable, just need to fix the CI errors.

ClementNerma:
@fdncred fixed!

fdncred:
Thanks

PR #662, opened 2023-11-11
Add more derive impl for HistoryItemId

ClementNerma:
As the item's documentation indicates, the history item's internal ID is auto-incremented.

Which means we can use it to order entries, as well as using it as e.g. an `HashMap`s key.


fdncred:
Seems reasonable. Thanks

PR #666, opened 2023-11-14
Also run clippy on examples

Hofer-Julian:
This PR also fixes existing clippy lints

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Clippy
-        run: cargo clippy ${{ matrix.flags }} --examples --all -- -D warnings
+        run: cargo clippy ${{ matrix.flags }} --all-targets --all -- -D warnings
 
 
       - name: Tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Clippy
-        run: cargo clippy ${{ matrix.flags }} --all -- -D warnings
+        run: cargo clippy ${{ matrix.flags }} --examples --all -- -D warnings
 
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Clippy
-        run: cargo clippy ${{ matrix.flags }} --all -- -D warnings
+        run: cargo clippy ${{ matrix.flags }} --examples --all -- -D warnings
```
sholderbach, line 47: Sadly this only clippies the examples so here we need at least
```suggestion
        run: cargo clippy ${{ matrix.flags }} --examples --lib --all -- -D warnings
```

Simpler and more expansive would be:

```suggestion
        run: cargo clippy ${{ matrix.flags }} --all-targets --all -- -D warnings
```

(where I am not sure which impact it has on the compilation flags/feature unification or if it has to rerun compilations a bunch)

sholderbach (changes requested):
Thanks for trying to improve our tooling here, and fixing the examples.

The wording around the build target flags is a bit unclear so no foul.

.github/workflows/ci.yml
Hofer-Julian: > Sadly this only clippies the examples so here we need at least

I was wondering that too since there is zero docs for `--examples`, `--all` or `all-targets` as far as I can see.
However, I've added a lint to lib.rs locally, and it was still detected correctly when running `cargo clippy --all --examples`.

sholderbach: Ah I ran my testing on `src/main.rs` of nushell. Here the `--bins` or `--bin nu` would not be checked with just `--examples`. The minimal docs for that are on `cargo check`

sholderbach: I would guess for future defensiveness we can go with `--all-targets`

sholderbach (approved):

PR #668, opened 2023-11-15
vi mode: map `cw`/`cW` to `ce`/`cE` by default

crides:
make `cw`/`cW` delete uptil the next space (like `ce`/`cE`). See [the vim help page on `cw`](https://vimhelp.org/change.txt.html#cw). This behavior is in both `zsh` and `bash`


sholderbach:
The title was at first a little confusing but you are correct when it comes to the behavior until which character things are changed (if starting on a word. Thanks for linking to the documentation as well! As described in the docs there is a slight difference between `ce` and `cw` when on whitespace. so this change wouldn't reflect that, but it would be a judgemeent call if we want to implement that as a separate op before simplifiying things internally for text objects (`ciw` my beloved).

PR #670, opened 2023-11-18
vi mode: Introduce alternate escape sequence for insert mode

benvansleen:
This PR introduces a new argument to the `Vi::new` constructor: the `alternate_esc_seq`. When set to `(KeyCode::Null, KeyCode::Null)`, this does nothing.

`alternate_esc_seq` can be set to something like `(KeyCode::Char('j'), KeyCode::Char('k'))` -- allowing the user to press "jk" (without modifiers) to move from Vi insert mode to Vi normal mode.

I wrote this because I would love to be able to use "jj" to enter normal mode in Nu (like I do for all other similar vi-modes).

(I am very new to Rust, Nu, and Reedline; if something is non-idiomatic, please tell me!)

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -114,7 +114,11 @@ fn main() -> std::io::Result<()> {
 
         add_newline_keybinding(&mut insert_keybindings);
 
-        Box::new(Vi::new(insert_keybindings, normal_keybindings))
+        Box::new(Vi::new(
+            insert_keybindings,
+            normal_keybindings,
+            (KeyCode::Null, KeyCode::Null),
+        ))
     } else {
         let mut keybindings = default_emacs_keybindings();
         add_menu_keybindings(&mut keybindings);
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -30,6 +30,8 @@ pub struct Vi {
     previous: Option<ReedlineEvent>,
     // last f, F, t, T motion for ; and ,
     last_char_search: Option<ViCharSearch>,
+    alternate_esc_seq: (KeyCode, KeyCode),
+    most_recent_keycode: KeyCode,
 }
 
 impl Default for Vi {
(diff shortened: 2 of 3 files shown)


benvansleen:
Addressing: https://github.com/nushell/reedline/issues/499

kuchta:
Actually all other major vi modes (vim, bash, zsh) uses Meta (Alt) + whatever for entering normal mode. I you get used to it, it's indispensable, but reedline is the only one that doesn't support that...

tamlok:
Can we use `ctrl+[` to exit insert mode?

benvansleen:
@tamlok 
Yes, the keycode C-[ is equivalent to ESC.

benvansleen:
@fdncred 

I believe the `config.nu` configuration is defined within the `nushell` crate. In `reedline`, this configuration is exposed by the `Vi` constructor.

If this PR is approved, I will write an accompanying `nushell` PR.

fdncred:
> I believe the config.nu configuration is defined within the nushell crate.

yes, of course. What I'm saying is will there be some way to define it in the keybinings section like this below but changed somehow to allow for other key combinations to be used for `alternate_esc_seq` instead of just j/k.

```
  {
    name: abbr
    modifier: control
    keycode: space
    mode: [emacs, vi_normal, vi_insert]
    event: [
    { send: menu name: alias_menu }
    { edit: insertchar, value: ' '}
    ]
  }
```

benvansleen:
Ah, I see what you mean. Currently, on the `reedline` side of things, the keys composing the sequence are configured by the (new) 3rd argument to `Vi::new`: 
```
Vi::new(insert_mode_keybinds, normal_mode_keybinds, (KeyCode::x, KeyCode::y))
```
This means the user can press "x" then "y" to exit insert-mode. (Passing `(KeyCode::Null, KeyCode::Null)` instead recovers the current behavior where only `ESC` exits insert-mode.)

I was thinking of implementing the `nushell` PR such that `config.nu` would look something like this:
```
...
bracketed_paste: true # enable bracketed paste, currently useless on windows
edit_mode: vi # emacs, vi
vi_alternate_esc_seq: { first: x, second: y }
shell_integration: true # enables terminal shell integration. Off by default, as some terminals have issues with this.
...

```

Achieving this with the existing `keybindings` section would require implementing something akin to emacs key-chords (since you press one key, release, then press the next). That would be a much, much wider-reaching change.

eggcaker:
any news?

ck3mp3r:
this is one of the first settings I tweak in neovim and also zsh vim mode...

omerxx:
Hi any news on this?
Hitting esc breaks my fingers 😅

Jasha10:
It seems that this PR may be blocked pending some design decisions.

I've opened issue https://github.com/nushell/reedline/issues/853 in an attempt to move discussion forward.

eggcaker:
any news?

PR #673, opened 2023-11-29
Respect termios `ECHO` flag

IanManske:
This PR makes it so that the `ECHO` flag given by [termios](https://www.man7.org/linux/man-pages/man3/termios.3.html) is respected. That is, the input characters are not printed/echoed back to the user if the `ECHO` flag is missing. This will help with using `rexpect` to test `reedline` and `nushell`, as the syntax highlighting, etc. does not need to be accounted for.
Related to: https://github.com/nushell/nushell/pull/11178

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ crossbeam = { version = "0.8.2", optional = true }
 crossterm = { version = "0.27.0", features = ["serde"] }
 fd-lock = "3.0.3"
 itertools = "0.10.3"
-nix = { version = "0.27.1", features = ["term"] }
 nu-ansi-term = "0.49.0"
 rusqlite = { version = "0.29.0", optional = true }
 serde = { version = "1.0", features = ["derive"] }
@@ -30,6 +29,9 @@ strum_macros = "0.25"
 thiserror = "1.0.31"
 unicode-segmentation = "1.9.0"
 unicode-width = "0.1.9"
+
+[target.'cfg(target_family = "unix")'.dependencies]
+nix = { version = "0.27.1", features = ["term"] }
 
 [dev-dependencies]
 gethostname = "0.4.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -19,6 +19,7 @@ crossbeam = { version = "0.8.2", optional = true }
 crossterm = { version = "0.27.0", features = ["serde"] }
 fd-lock = "3.0.3"
 itertools = "0.10.3"
+nix = { version = "0.27.1", features = ["term"] }
```
sholderbach, line 22: Not super critical but we could make this a unix only dependency for slightly faster downloads on windows.

src/engine.rs
```diff
@@ -149,6 +149,9 @@ pub struct Reedline {
     // Manage optional kitty protocol
     kitty_protocol: KittyProtocolGuard,
 
+    // Echo typed input
```
sholderbach, line 152: As this an absolutely obscure usecase I think expanding this to a small doccomment with pointers what you are trying to achieve and the relation to the `termios` flag makes sense for future generations.

sholderbach (commented):
Not sure what is the best choice to test `reedline's` specifc behavior with `rexpect` or what conventions arose for other shells when given `~ECHO`. 
For your Nushell tests in https://github.com/nushell/nushell/pull/11178 having the hinter dump stuff from the history may possibly interfere.

Cargo.toml
IanManske: Thanks, forgot to do that.

IanManske:
> For your Nushell tests in https://github.com/nushell/nushell/pull/11178 having the hinter dump stuff from the history may possibly interfere.

Thankfully, I don't think this will be an issue:
- The tests so far only send whole lines, so there shouldn't? be an opportunity for the hints to show.
- `rexpect` will skip over output until it finds the expected output anyways, so this should include hints. (Syntax highlighting, on the other hand, is interspersed with the output which messes things up).

This is also part of the reason why I was wondering if there was a way to disable the history for nushell or start with a blank history file.

PR #676, opened 2023-12-05
Allow to construct ReedlineError values from the outside

ClementNerma:
This PR allows to build a custom `History`, which requires returning `Reedline<_, ReedlineError>` values ; the problem being that `ReedlineError` values aren't currently constructable from outside of the `reedline` crate itself.

This will require to carefully craft the variants once we decide to publish the 1.0 version


fdncred:
Thanks

PR #678, opened 2023-12-06
 Make history-related items (de-)serializable

ClementNerma:


fdncred:
I generally don't have problems with your recent PRs but I'd like to see something in the description that we can refer to later, even though these are quite simple changes. 

Where are you integrating reedline that is forcing you to make these changes? I'm just interested to see what you're doing.

ClementNerma:
> I generally don't have problems with your recent PRs but I'd like to see something in the description that we can refer to later, even though these are quite simple changes. Obviously the CI needs to be green too. :)

Just edited the description and updated the code so the CI should pass.

> Where are you integrating reedline that is forcing you to make these changes? I'm just interested to see what you're doing.

I'm building full-blown custom shell, and I'm currently polishing the last parts, like the history. I want to store specific informations, maybe perform history synchronization between shell instances etc. So I absolutely need a custom `History` for that.

fdncred:
thanks for the info. i landed your pr that had a green ci.

ClementNerma:
CI is now green.

fdncred:
one of the other prs caused a conflict here

ClementNerma:
Fixed.

fdncred:
good luck, have fun!

fdncred:
> I'm building full-blown custom shell, and I'm currently polishing the last parts, like the history. I want to store specific informations, maybe perform history synchronization between shell instances etc. So I absolutely need a custom History for that.

Is this public somewhere that i can look and play with it?

ClementNerma:
Sure: https://github.com/ClementNerma/ReShell

Please note that there is no tutorial or documentation yet, so you'll have to decode the [`parser`](https://github.com/ClementNerma/ReShell/blob/main/crates/parser/src/program.rs) and [`AST`](https://github.com/ClementNerma/ReShell/blob/main/crates/parser/src/ast.rs) files yourself currently :p

Here is a little example (note that they are still some bugs, especially on the stack):

```rust
fn fibonacci(value: int) -> int {
    return if $value < 1 {
        0
    } else if $value <= 2 {
        1
    } else {
        fibonacci($value - 1) + fibonacci($value - 2)
    }
}

for i in range(1, 100) {
    echo "fib $i = ${fibonacci($i)}"
}
```

Issue #380, opened 2022-04-05
Select with shift-cursor and cut, copy and paste


DejayRezme:
I'm really missing features to select text with keys to mark for cut, copy and pasting. Not sure if this is the right place to suggest this or if this is planned or if there is some other way of doing this with bash or emacs knowledge. 

* shift-cursor left or right to select (mark) chars
* control-shift-cursor left or right to select words
* control-x should cut selected text
* delete should delete selected text
* control-a should selects whole line (or use CutCurrentLine? instead)
* All CutXXX commands like "CutFromStart" or "CutWordLeft" should use the OS copy/paste buffer not an internal buffer

Maybe there could also be a mode where I can move around in the terminal with the cursor and mark and copy text?

The new powershell terminal does all this really nice now. Coming from windows I'm really missing these features since they work in all text editors and I have muscle memory for them. I

## References
#10

sholderbach:
A selection mode inside the line is definetely in scope. And having a set of keybindings, which you are implying with ctrl-a etc., that are closer to the experience using Windows/windowed applications instead of emacs would also be nice to make it more attractive to people that previously were maybe not unix/terminal power users.

Doing selection beyond the current input buffer (e.g. previous output) is not possible with the current design as reedline is not managing the output of other programs (Basically still writes to the virtual teletype paper)

Could you move the other issues to their own? (`keybindings default` is implemented in part in nushell/nushell maybe file it there)

EmilyisWIP:
Any updates on this? *Boop*

sachinjoseph:
@sholderbach @jntrnr this would be a great addition! :-)

fdncred:
Adding [LineEditor](https://github.com/AmrDeveloper/Lineeditor) that implements this selection functionality and is inspired by reedline. They have found a way to do this with crossterm and we could use their code for inspiration ourselves.

Tastaturtaste:
I started to implement the requested functionality in #689. For now only selecting with Shift + Arrow, cut with Ctrl + X, and paste with Ctrl + V is implemented. Pasting with Ctrl + V is only possible (on windows) if `reedline` is compiled with the feature `system_clipboard`, otherwise the key combination is intercepted by windows itself and the contents of the system clipboard are pasted instead of the local contents. In nushell this feature is not enabled afaik, which I hope can be reconsidered to more seamlessly integrate with everything else.

Tastaturtaste:
I think this Issue can be closed, too. What do you think, @fdncred?

[...]

PR #689, opened 2023-12-27
Add ability to select and cut text in the input buffer

Tastaturtaste:
Addresses issue #380.

> All CutXXX commands like "CutFromStart" or "CutWordLeft" should use the OS copy/paste buffer not an internal buffer

This is already the case if the crate is compiled with the feature `system_clipboard`. This is, afaik, not (yet) done for nushell.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -61,8 +61,8 @@ impl Editor {
             EditCommand::InsertNewline => self.line_buffer.insert_newline(),
             EditCommand::ReplaceChar(chr) => self.replace_char(*chr),
             EditCommand::ReplaceChars(n_chars, str) => self.replace_chars(*n_chars, str),
-            EditCommand::Backspace => self.line_buffer.delete_left_grapheme(),
-            EditCommand::Delete => self.line_buffer.delete_right_grapheme(),
+            EditCommand::Backspace => self.backspace(),
+            EditCommand::Delete => self.delete(),
             EditCommand::CutChar => self.cut_char(),
             EditCommand::BackspaceWord => self.line_buffer.delete_word_left(),
             EditCommand::DeleteWord => self.line_buffer.delete_word_right(),
@@ -564,6 +564,28 @@ impl Editor {
             self.selection_anchor = Some(self.insertion_point());
         }
         self.line_buffer.move_word_right();
+    }
+
+    fn delete_selection(&mut self) {
+        self.line_buffer.clear_range_safe(
+            self.selection_anchor.expect("Deleting a selection only makes sense if we indeed have a selection. Calling without one is probably a bug."), 
+            self.line_buffer.insertion_point());
+    }
+
+    fn backspace(&mut self) {
+        if self.selection_anchor.is_some() {
+            self.delete_selection();
+        } else {
+            self.line_buffer.delete_left_grapheme();
+        }
+    }
+
+    fn delete(&mut self) {
+        if self.selection_anchor.is_some() {
+            self.delete_selection();
+        } else {
+            self.line_buffer.delete_right_grapheme();
+        }
     }
 }
 
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -9,9 +9,9 @@ use crate::{core_editor::get_default_clipboard, EditCommand};
 pub struct Editor {
     line_buffer: LineBuffer,
     cut_buffer: Box<dyn Clipboard>,
-
     edit_stack: EditStack<LineBuffer>,
     last_undo_behavior: UndoBehavior,
+    selection_anchor: Option<usize>,
 }
 
 impl Default for Editor {
@@ -21,6 +21,7 @@ impl Default for Editor {
             cut_buffer: Box::new(get_default_clipboard()),
             edit_stack: EditStack::new(),
             last_undo_behavior: UndoBehavior::CreateUndoPoint,
+            selection_anchor: None,
         }
     }
 }
@@ -96,6 +97,11 @@ impl Editor {
             EditCommand::CutLeftBefore(c) => self.cut_left_until_char(*c, true, true),
             EditCommand::MoveLeftUntil(c) => self.move_left_until_char(*c, false, true),
             EditCommand::MoveLeftBefore(c) => self.move_left_until_char(*c, true, true),
+            EditCommand::SelectMoveLeft => self.select_move_left(),
+            EditCommand::SelectMoveRight => self.select_move_right(),
+            EditCommand::SelectAll => self.select_all(),
+            EditCommand::CutSelection => self.cut_selection(),
+            EditCommand::CopySelection => self.copy_selection(),
         }
 
         let new_undo_behavior = match (command, command.edit_type()) {
@@ -112,6 +118,12 @@ impl Editor {
             (_, EditType::UndoRedo) => UndoBehavior::UndoRedo,
             (_, _) => UndoBehavior::CreateUndoPoint,
         };
+        if !matches!(
+            command,
+            EditCommand::SelectMoveLeft | EditCommand::SelectMoveRight | EditCommand::SelectAll
+        ) {
+            self.reset_selection();
+        }
         self.update_undo_state(new_undo_behavior);
     }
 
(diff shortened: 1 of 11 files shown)


src/core_editor/editor.rs
```diff
@@ -39,79 +40,91 @@ impl Editor {
     }
 
     pub(crate) fn run_edit_command(&mut self, command: &EditCommand) {
+        use EditCommand as EC;
```
fdncred, line 43: I'm not sure I see the point of this. I'd prefer to leave it EditCommand.

fdncred:
This looks like a promising start. Is it enabled in any of the reedline demos so we can try it out?

Tastaturtaste:
You can try it out in the basic example of reedline. It is not gated behind a feature. But if you want to use Ctrl+V you have to compile with feature system_clipboard, at least on windows.

src/core_editor/editor.rs
Tastaturtaste: Ok, no problem. In my opinion it's more noise than necessary, but I don't really care that much.

fdncred: Thanks. I'm really looking forward to playing with this.

fdncred:
Hmmm, I'm on macos and can't get it to work. I compiled with `cargo run --examples basic --features=system_clipboard`. I tried Ctrl+A, Shift+Left, Shift+Right and I'm not getting any selection.

Tastaturtaste:
I have not yet implemented highlighting of the selection. If you hold down Shift and move the cursor to over text and afterwards press Ctrl + X you should have the text in the clipboard. At least pasting inside the terminal again worked without problem for me.

fdncred:
oh, i see. seems to work great with that little tidbit. 😆 thanks.

Tastaturtaste:
Is there anything left to do before you can consider to merge it? The two failing clippy lints are not related to the code I touched, so I don't know what to do about them...

fdncred:
Just change the code like the clippy lints suggest, replacing `.get(0)` with `.first()`. The CI needs to be green prior to landing.

Tastaturtaste:
> Do you plan on making a PR for nushell based on this PR?

Yeah, that was my plan

fdncred:
I'd like to know/see what changes are required in nushell before landing this.

fdncred:
I've looked over this several times now and I'm about ready to land it. I see you have a builder pattern to implement it in nushell (and others). How are you feeling about this PR/Do you still have things you want to do or are you ready?

Tastaturtaste:
> I see you have a builder pattern to implement it in nushell

Yes, that should work just like any other EditCommand. However, the bindings `Ctrl + C` and `Ctrl + A` are used in the default `config.nu` for `cancel_command` and `move_to_line_start` respectively. I don't know how a differentiation between `Ctrl + C` to cancel a operation and `Ctrl + C` to copy a selection could be handled. Maybe there is some way to see if currently some text is selected and treat it as a copy command and otherwise as a cancel command?

> How are you feeling about this PR/Do you still have things you want to do or are you ready?
Actually, I am currently working on making selection an option for all of the previously existing `move` commands. This should afford much more flexibility and I believe is better extensible. 
Please take a look if you think this would be beneficial. I think it is an improvement.

fdncred:
Would you mind talking a bit about how the selection, etc works in relation to graphemes? Seems like we have a lot of grapheme work here but I'm not sure all that is accommodated in your new code or not and I'd hate to break that type of thing.

Tastaturtaste:
Of course, I will try to give a brief summary of the selection logic and how it relates to grapheme boundaries.

Any `EditCommand` of the `MoveCursor` type carries with it a parameter specifying if the move should select text that is moved over. This works as follows:
1. The `Editor` has a field `selection_anchor` of type `Option<usize>`, which tracks if a selection is currently happening, and if so, it carries the index of one end of the selected range into the line_buffer. The other end of the selected range is marked by the current cursor position.
2. When a `EditCommand` with a selection request (`select == true`) is processed, a check is made if we previously already started selecting text (indicated by `Some(usize)` in `selection_anchor`). If yes, the selection_anchor remains set as it is. If no, it is set to the current cursor position before moving the cursor.
3. When an `EditCommand` is processed that does anything else than select text by moving the cursor the `selection_anchor` is reset to `None`.

By this logic, the `selection_anchor` can only contain indices that were previously also cursor positions. Assuming
1. Cursor positions are always valid indices at grapheme boundaries
2. All commands that modify the line buffer also reset the `selection_anchor`

the indices bracketing the selected text in the line buffer are always valid indices at grapheme boundaries.
Assumption 2. is given because all `MoveCursor` edit commands do not modify the line buffer and conceptually probably never should. However, this *could* be a maintenance hazard if somehow a `MoveCursor` command with `select == true` were added that *does* modify the line buffer. This hazard could be contained if mutable access to the line_buffer would always have to be done through a custom type or function that could then make sure to always reset the `selection_anchor`.

fdncred:
Thanks so much for the detailed explanation :heart:. It seems like we'd want to remove this maintenance hazard if we could, although I believe that to be outside the scope of this PR.

fdncred:
how's things going here?

Tastaturtaste:
Yep, the clipboard branch is already in main and I rebased on it. So I think this is good to go unless I miss something.

Tastaturtaste:
However, I just saw the CI with nushell is red. In that PR I already use this branch of reedline and the test fail in the nushell pipeline seems so be related to the system clipboard. I don't yet know what the reason could be.

fdncred:
ok, let's go with this then!

fdncred:
now you can update your nushell PR and we can troubleshoot there. hopefully we don't have to come back here to make any changes.

Tastaturtaste:
Hmm, I just cloned a fresh copy of nushell and tried to reproduce the commands not working. But for me everything works as expected. Just to be sure I did the right steps to reproduce:
```
git clone https://github.com/nushell/nushell.git
cd nushell
cargo run -- -n
<Test commands here>
```
Are there specific steps to reproduce the issue?

fdncred:
@amtoine can you comment on this?

@Tastaturtaste It works on Windows for me with `nu -n`. I was having problems on mac. I can test there later. I'll try ubuntu on WSL and see in a minute.

fdncred:
@Tastaturtaste on WSL Ubuntu, when I typed `shift left-arrow` I got this
image

with full backtrace
image

Tastaturtaste:
I can't take a closer look today, but will look into it in the next few days. Sadly my weekend is packed already, so maybe after that. 

The backtrace gives a good hint already. Somewhere in `reedline::painting::styled_text::StyledText::style_range` the indexing of the styled buffer (or rather of one styled fragment of the buffer) goes wrong. The idea in that function is to store each continuous fragment of the buffer with the same style as a pair of `String` and `Style`. When something is selected the style has to be updated to show the selection effect.

I can reproduce the crash with a shorter prompt on windows:
1. Type `cargo run` (do not press enter)
2. press left once (cursor is now between u and r
3. press shift + left

fdncred:
@Tastaturtaste Sounds good. We appreciate your help whenever you can get to it!

kubouch:
@Tastaturtaste Could you put that panic with repro steps in a separate issue and slap the "Critical" label on it? So that we can track easier in case you won't have time in the next few days.

fdncred:
I think this is a similar panic in nushell https://github.com/nushell/nushell/issues/11813

Tastaturtaste:
https://github.com/nushell/reedline/pull/689#issuecomment-1934978686
I got this fixed on [this branch](https://github.com/Tastaturtaste/reedline/tree/bug/style-text). Now I need to try and reproduce the error @fdncred got on WSL.

Tastaturtaste:
> @Tastaturtaste on WSL Ubuntu, when I typed `shift left-arrow` I got this. Note that this happens when I accept an item out of my history. e.g. i typed `nu` and the history hint was `nu -c "def table [] { \"hi\" }; table"` so I right-arrowed to complete it, hit left-arrow once to get inside the trailing double quote, then hit shift left-arrow for the crash. image
> 
> with full backtrace image

I reproduced the error on WSL. The fix for my subtraction with overflow error also fixes this one. Unless someone says otherwise I consider this bug fixed with https://github.com/nushell/reedline/pull/751.

fdncred:
Thanks @Tastaturtaste. Do you have a nushell branch that I can easily test this in? I'm not quite sure how to reproduce it in reedline. I've tried but can't make it fail.

Tastaturtaste:
I don't have one pushed. Locally I tested with the nushell main branch and set reedshell patch to my git branch. Do you need me to push a nushell branch to test?

fdncred:
> Do you need me to push a nushell branch to test?

If it's not too much trouble. It would make it easier for us to test.

Tastaturtaste:
https://github.com/Tastaturtaste/nushell/tree/test-reedline-style-text-fix
Here, have a go. Should I make a quick PR against nushell so you can run CI tests?

fdncred:
Let me try to reproduce the problem and try your branch and see how far that gets. Thanks so much! Gimme a few and I'll ping you back.

fdncred:
Looks like it fixes it to me. Thanks

PR #690, opened 2023-12-27
Use empty history if search query fails

ysthakur:
There's a bunch of places with `.expect("todo: error handling")`, and this PR tries to remove most of them. Here's the four different kinds:

- There are a couple inside `examples/demo.rs` that I didn't bother with.
- Some of these are for unwrapping the result of running a search query to look through the history. I figured that instead of panicking, it may be better to silently treat `Err`s as the history being empty.
- One of them comes after [`history.save(entry)`](https://github.com/nushell/reedline/blob/main/src/engine.rs#L1757), in `engine.rs`. This has been edited to silently ignore an `Err`.
- The rest of the `.expect`s are all inside `engine.rs` and are for handling the result of moving the `history_cursor` forward or backward. Unfortunately, all of them appear inside methods that return either `()` or an `io::Result`. These have also been edited to silently ignore `Err`s.

A disadvantage of silently handling `Err`s is that users won't know that something went wrong, but if the `.expect`s are to be kept, they should at least be changed to have more useful messages to help users file bug reports.

There shouldn't be any user-facing changes from this.

### DIFF
--- a/src/completion/history.rs
+++ b/src/completion/history.rs
@@ -23,7 +23,7 @@ impl<'menu> Completer for HistoryCompleter<'menu> {
             .search(SearchQuery::all_that_contain_rev(
                 parsed.remainder.to_string(),
             ))
-            .expect("todo: error handling");
+            .unwrap_or(vec![]);
 
         values
             .into_iter()
@@ -40,7 +40,7 @@ impl<'menu> Completer for HistoryCompleter<'menu> {
             .count(SearchQuery::all_that_contain_rev(
                 parsed.remainder.to_string(),
             ))
-            .expect("todo: error handling");
+            .unwrap_or(0);
         count as usize
     }
 }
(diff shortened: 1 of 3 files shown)


amtoine:
silent errors sound worse to me than panicking :thinking: 
would be nice to give a proper error to the users

ysthakur:
@amtoine Hmm, I guess so. I'm going to try fuzzing later to try to find cases that trigger panics, and if I can't find anything, I'll just change the expect messages to be more descriptive

ysthakur:
I did some fuzzing in [this branch](https://github.com/ysthakur/reedline/tree/fuzz) and couldn't find any real defects, so I guess I'll just change the error messages to be more descriptive, there's probably no need to do actual error handling.

ysthakur:
Actually, I have no idea what kind of message would be helpful there, so I think I'll just close this for now, lol

fdncred:
darn, i was hoping this pr would get rid of some of the non-error-handling we have. maybe in the future sometime?

Issue #473, opened 2022-09-06
f/F/t/T motions don't work for characters such as b or 0


morzel85:
**Platform** Ubuntu 20.04, Nushell 0.67.0
**Terminal software** gnome-terminal

I noticed problem with `f/F/t/T` motions to characters such as: `b`, `w`, `1` or `5`. 

For example `fb `doesn't move cursor from beginning of `aaabbbccc` to the letter `b` (but `fc `would correctly move to letter `c`). Reedline must be getting confused with special meaning of word motions or counts.

Please note that result of this issue is similar to https://github.com/nushell/reedline/issues/449 but root cause is different.
If `f/F/t/T` target is pending, parser should **not** treat following character as count or motion.

## Steps to reproduce
1. Run Reedline in `vi` mode: `cargo run -- --vi`
2. Type some text such as `aaabbbccc`
3. Try to execute `f/F/t/T` motions targeting the letter `b`.
4. Notice that cursor doesn't move (try another `f/F/t/T` motion on some other letter such as `c` and notice that it works ok).

## Screenshots/Screencaptures
Some log from `src/edit_mode/vi/mod.rs`: 
![image](https://user-images.githubusercontent.com/7968512/188705896-da2e0568-2a8e-4f8c-8317-fa949f0adbe5.png)

Issue #449, opened 2022-07-06
Vi-normal mode f<char> doesn't work for period character


aikow:
**Platform**: Manjaro

**Window Manager**: awesomewm

**Terminal software**: alacritty

Used with nushell

## Problem

Trying to jump to the next or previous period character (`.`) does not work,
even though jumping to all other characters is possible. With 

```nu
open part_a.part_b.part_3.csv
```

typing `f.` in normal mode does absolutely nothing. I tested it with most other
characters, and they all seem to work fine.

Also the word boundary definition for period seems to be very inconsistent as
well, internally, and with `vi` and `readline`.

Given the above input navigating with `w` the jump forward word character puts
the cursor into the following positions

```nu
open part_a.part_b.part_3.csv
^    ^                   ^^  ^
```

Readline stays consistent by jumping like this

```nu
open part_a.part_b.part_3.csv
^    ^     ^^     ^^    ^^
```

which is predictable. `fish`'s vi mode has the different behavior again and
jumps straight to each part like so

```nu
open part_a.part_b.part_3.csv
^    ^    ^ ^    ^ ^    ^ ^
```

## Steps to reproduce

`config.nu`

```nu
let-env config = {
  edit_mode: vi
}
```

## Solution

Adding a setting to allow users to configure their own word separators would be
a nice feature.

sholderbach:
The `f.` case is definetely a bug. It also seems to get stuck in the parser state machine.
If somebody wants to pick that up, I think it is not too hard to solve. (Related issue that can be solved by lifting the dot into the full command parser instead of a special case, would be that the dot currently doesn't support multipliers)

For the word definition (if you consider particularly pressing maybe lift it to a different issue): In my research on different word definitions I came accross many inconsistencies like that (e.g. emacs evil mode disagreeing with vim. bash, zsh, fish all doing slightly different things also for the emacs bindings) As a result I for now settled on the Unicode Annex 29 word definitions (they consider `snake_case` a single word, if I recall basically UAX31 identifiers are vaild words) with a special case to jump over whitespace. 
In general we should strive to be relatively faithful to the precedent but there were a few annoyances we were able to solve with this definition (e.g. deleting separate special characters like the `|` without touching a preceding or following word that is not nice in some other line editors. #346)

aikow:
My main complaint with the current word boundary definition is that numbers seem to count as a word boundary even if they are part of `snake_case1word`, which seems a bit unintuitive.

[...]

PR #693, opened 2024-01-02
incomplete key sequences should be completed first

crides:
Fixes an issue like #473 and #449, but for custom keybindings. Fixed by checking if the current key sequence is complete, and if not, complete it before checking for keybindings.

Added a test for this scenario

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -30,6 +30,7 @@ pub struct Vi {
     previous: Option<ReedlineEvent>,
     // last f, F, t, T motion for ; and ,
     last_char_search: Option<ViCharSearch>,
+    seq_completed: bool,
 }
 
 impl Default for Vi {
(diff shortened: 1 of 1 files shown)


fdncred:
the ci doesn't agree with you. 😆

crides:
Huh, a bunch of things I didn't touch. I can fix later today

crides:
aight there we go

crides:
hmm this is not actually general enough (not accounting for insert mode bindings) but I don't think such a case can exist in insert mode right now (not even for custom bindings)? Examples in normal vim/zsh vi mode will be `^V` and `^K` for inserting characters/digraphs; or `^O` if you wanna include that (temporary exit to normal mode; see [`:h i_^O`](https://vimhelp.org/insert.txt.html#i_CTRL-O))

kronberger-droid:
Closed in favor of #1103. Does essentially the same but rebased onto current editor.
Put @crides as co author.
Thanks.

PR #697, opened 2024-01-04
Try fix prompt_start_row reset to 0 when opening a file without newline in Nushell 

WindSoilder:
This pr is going to apply one of solution that @nibon7 attempted: https://github.com/nushell/nushell/issues/11399#issuecomment-1867921387

Relative: https://github.com/nushell/nushell/issues/11399


fdncred:
after talking to @WindSoilder, let's give this a try.

PR #584, opened 2023-05-23
Fix move to line start in multi-line history entries

boathouse2112:
Fixes #582 

Reverts part of #97 -- Normal traversal sets history string at cursor to line buffer. It's not clear to me what that was added in response to.

Added tests covering the problem and the fix in `engine.rs`. The bug occurs in `engine.rs`, so it's not possible to test in `core_editor/line_buffer.rs`, where the other edit command tests are.

Completed the UX checklist the best I could, and ran nushell with the fix for a bit.

## Manual checks

Relevant features tested (leave open if you did not consider those areas touched by your PR):

- [x] core editing and default Emacs keybindings
- [x] history
- [ ] syntax highlighting
- [x] completion/hinting
- [x] vi mode

### Info

Build: [x] debug / [ ] release

Platform:

Terminal emulator:

Inside a [ ] ssh,[ ] tmux or [ ] screen session?

### Basics

- [x] Typing of a short line containing both upper- and lowercase characters.
- [x] Movement left/right using the arrow keys
- [x] Word to the left with `Ctrl-b` or `Ctrl-Left`, Word to the right with `Ctrl-f`
- [x] `Enter` to complete entry

#### Clearing

- [x] Type something and abort the entry with `Ctrl-c`, you should end up on an empty prompt below.
- [x] Type something and press `Ctrl-l` to clear the screen. Your current entry should still be there and passed through when pressing `Enter`

#### Unicode and Emojis

- [x] Paste the line `Emoji test 😊 checks 🤦🏼‍♂️ unicode` and move the cursor over the emojis.
- [x] Are you able to delete the smiley?
- [x] `Home`/`End` at accurate positions
- [x] Check that the emoji containing line can be entered

## History

- [x] On the empty line press the `up-arrow` key to see if you can recall the previous entry
- [x] Press `Enter` to execute this line (it should *not* be duplicated in the history, after checking leave history recall by `down-arrow`)
- [x] On an empty line start typing the beginning of a line in the history. Hit the `up-arrow` to find the matching entry.
- [x] After that run `Ctrl-r` to start traditional reverse search. Type your initial search. Can you find more hits by pressing `Ctrl-r` or `up-arrow`?
- [x] Abort this search by pressing `Ctrl-c`

PR #704, opened 2024-01-11
Revert "Fix move to line start in multi-line history entries"

fdncred:
Reverts nushell/reedline#584

This is causing odd test behavior like ansi escapes, seeing reedline prompts in the middle of testing, leaving the terminal in a raw state.

@boathouse2112 If you can find a fix for this it would be great!!

These are the two test commands that I run and see the problems.
```
cargo test --features=bashisms,sqlite,external_printer,system_clipboard
```
and
```
cargo nextest run --all-features
```

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1403,6 +1403,15 @@ impl Reedline {
     /// Executes [`EditCommand`] actions by modifying the internal state appropriately. Does not output itself.
     pub fn run_edit_commands(&mut self, commands: &[EditCommand]) {
         if self.input_mode == InputMode::HistoryTraversal {
+            if matches!(
+                self.history_cursor.get_navigation(),
+                HistoryNavigationQuery::Normal(_)
+            ) {
+                if let Some(string) = self.history_cursor.string_at_cursor() {
+                    self.editor
+                        .set_buffer(string, UndoBehavior::HistoryNavigation);
+                }
+            }
             self.input_mode = InputMode::Regular;
         }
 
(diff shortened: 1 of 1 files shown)


stormasm:
We need to be able to run these tests on all (3) of our platforms including Windows and Mac...

Currently the CI is not set up to do this...

I would be willing to run the tests on my Mac once they are fixed and I am sure someone would be willing to run them on Windows as well prior to re-landing these tests...

Thanks !

stormasm:
I don't think these tests belong in [Engine](https://github.com/nushell/reedline/blob/main/src/engine.rs).

I think we should start a test directory under the top level Reedline and put them there...

The following comments were observed on a Mac.

There appears to be some strange race conditions which are caused by running the tests in parallel
with nextest...

They pass some of the time in the cargo test runner (with no issue) but have more serious issues in nextest which is not a good sign.

stormasm:
I am going to experiment with refactoring the tests and getting rid of the [rstest](https://docs.rs/rstest/latest/rstest/) functionality.  This may be blocking nextest which is what I am seeing...

I think by refactoring the tests to have individual tests with just the different type of input scenarios would be simpler to test longer term.  Right now all of the cases are bundled into two tests with 3 cases each...

Also --- for some reason and this is a question for folks reading this....

I can NOT FIGURE OUT how to IGNORE rstests --- if anyone knows how to do this let me know !  thanks...

```rust
#[ignore] 
```

IanManske:
> There appears to be some strange race conditions which are caused by running the tests in parallel

That was my suspicion as well (multiple tests all trying to read/write from the terminal). But even when running the tests serially (`cargo test -j 1`), the terminal output is still garbled. I assume our testing structure isn't meant to interact with the terminal which is why there previously weren't any tests for the `Reedline` struct.

A more future-proof solution is to use [`rexpect`](https://crates.io/crates/rexpect) for testing, as it will create terminal files to isolate each test/process's stdin and stdout. Unfortunately, I think it only supports unix systems. PR #673 is related.

PR #606, opened 2023-07-18
Replace clipboard with copypasta

nibon7:
This PR tries to replace `clipboard` with [copypasta](https://github.com/alacritty/copypasta). `clipboard` is unmaintained and `xcb` that `clipboard` depends on has a rustsec advisory.

```
Crate:     xcb
Version:   0.8.2
Title:     Multiple soundness issues
Date:      2021-02-04
ID:        RUSTSEC-2021-0019
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0019
Solution:  Upgrade to >=1.0
Dependency tree:
xcb 0.8.2
└── x11-clipboard 0.3.3
    └── clipboard 0.5.0
        └── reedline 0.21.0

Crate:     clipboard
Version:   0.5.0
Warning:   unmaintained
Title:     clipboard is Unmaintained
Date:      2022-06-25
ID:        RUSTSEC-2022-0056
URL:       https://rustsec.org/advisories/RUSTSEC-2022-0056
Dependency tree:
clipboard 0.5.0
└── reedline 0.21.0

Crate:     xcb
Version:   0.8.2
Warning:   unsound
Title:     Soundness issue with base::Error
Date:      2020-12-10
ID:        RUSTSEC-2020-0097
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0097
Severity:  5.5 (medium)
```

PR #705, opened 2024-01-14
Replace clipboard with arboard, which is actively maintained

Tastaturtaste:
Adresses #606 and [this request](https://github.com/nushell/nushell/pull/11535#issuecomment-1890960094)

Replace `clipboard` dependency with [arboard](https://crates.io/crates/arboard). 

# Considerations
A look at crates.io shows that there are currently two major cross-platform clipboard crates to choose from:
1. [arboard](https://crates.io/crates/arboard)
2. [copypasta](https://crates.io/crates/copypasta)

`copypasta` is maintained by the `alacritty` project, which gives some confidence in the future maintenance story. Using this crate as a drop-in replacement for `clipboard` works just from an API standpoint, but leads to failure of `cargo test --features system_clipboard core_editor::clip_buffer::tests::reads_back` on windows. This is due [this bug](https://github.com/DoumanAsh/clipboard-win/pull/23) and `copypasta`s outdated version of `clipboard-win`. While this issue could be worked around, it does not instill much confidence in me that future issues will be resolved and dependencies kept up-to-date.

`arboard` is [actively maintained by a corporation depending on it](https://github.com/1Password/arboard/issues/26) and has roughly three times the number of downloads of `copypasta`. While the API is not a drop-in replacement for `clipboard`, the required changes are minimal. 

# Dependencies
`arboard` does pull in a few dependencies, mostly depending on the target OS. Many of the dependencies, like `log`, `thiserror`, `parking_lot` (on linux) are already present in reedline's dependency graph anyway. In addition 
However, as mentioned in #606, wayland support adds many dependencies to the list.

Number of unique dependencies for linux with system_clipboard enabled after cargo update:
current main: 53
arboard: 59
arboard + wayland: 99

Without having looked further into it `copypasta` also seems to pull in the majority of dependencies for wayland support. So I guess its mostly a wash in that department when comparing the alternatives.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,11 @@ version = "0.28.0"
 doctest = true
 
 [dependencies]
+arboard = { version = "3.3.0", optional = true, default-features = false, features = ["wayland-data-control"] }
 chrono = { version = "0.4.19", default-features = false, features = [
     "clock",
     "serde",
 ] }
-clipboard = { version = "0.5.0", optional = true }
 crossbeam = { version = "0.8.2", optional = true }
 crossterm = { version = "0.27.0", features = ["serde"] }
 fd-lock = "3.0.3"
@@ -44,7 +44,7 @@ bashisms = []
 external_printer = ["crossbeam"]
 sqlite = ["rusqlite/bundled", "serde_json"]
 sqlite-dynlib = ["rusqlite", "serde_json"]
-system_clipboard = ["clipboard"]
+system_clipboard = ["arboard"]
 
 [[example]]
 name = "cwd_aware_hinter"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,20 +4,21 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.8.3"
+version = "0.8.7"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -13,11 +13,11 @@ version = "0.28.0"
 doctest = true
 
 [dependencies]
+arboard = { version = "3.3.0", optional = true, default-features = false, features = ["wayland-data-control"] }
```
fdncred, line 16: Can we include this wayland feature only if it's not windows and not macos? At least that will cut the dependencies for some users.

fdncred:
Looks like pretty minimal changes, very nice! Thanks for this. I just had one further question above and then I think we could move forward.

Cargo.toml
Tastaturtaste: I made sure that is already the case. Take a look at https://github.com/1Password/arboard/blob/master/Cargo.toml.

I can do it explicitly in our Cargo.toml, too, if you want.

fdncred: ok, it looks like it's already that way in arboard. Thanks.

PR #696, opened 2024-01-03
ide style completions

maxomatic458:
#660 

demo clip: [asciicast](https://asciinema.org/a/urUsZ5DbRieBEY27mxfDTgd9S)

PR #708, opened 2024-01-19
get correct cursor pos when menu indicator contains newline

maxomatic458:
fixes a bug from #696

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -277,7 +277,7 @@ pub use validator::{DefaultValidator, ValidationResult, Validator};
 
 mod menu;
 pub use menu::{
-    menu_functions, ColumnarMenu, IdeMenu, ListMenu, Menu, MenuEvent, MenuTextStyle, ReedlineMenu,
+    menu_functions, ColumnarMenu, IdeMenu, DescriptionMode, ListMenu, Menu, MenuEvent, MenuTextStyle, ReedlineMenu,
 };
 
 mod terminal_extensions;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -277,7 +277,8 @@ pub use validator::{DefaultValidator, ValidationResult, Validator};
 
 mod menu;
 pub use menu::{
-    menu_functions, ColumnarMenu, IdeMenu, DescriptionMode, ListMenu, Menu, MenuEvent, MenuTextStyle, ReedlineMenu,
+    menu_functions, ColumnarMenu, DescriptionMode, IdeMenu, ListMenu, Menu, MenuEvent,
+    MenuTextStyle, ReedlineMenu,
 };
 
 mod terminal_extensions;
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -543,7 +543,7 @@ impl Menu for ColumnarMenu {
             // editing a multiline buffer.
             // Also, by replacing the new line character with a space, the insert
             // position is maintain in the line buffer.
-            let trimmed_buffer = editor.get_buffer().replace('\n', " ");
+            let trimmed_buffer = editor.get_buffer().replace("\r\n", "  ").replace('\n', " ");
             completer.complete(
                 &trimmed_buffer[..editor.insertion_point()],
                 editor.insertion_point(),
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -543,7 +543,7 @@ impl Menu for ColumnarMenu {
             // editing a multiline buffer.
             // Also, by replacing the new line character with a space, the insert
             // position is maintain in the line buffer.
-            let trimmed_buffer = editor.get_buffer().replace("\r\n", "  ").replace('\n', " ");
+            let trimmed_buffer = editor.get_buffer().replace('\n', " ");
             completer.complete(
                 &trimmed_buffer[..editor.insertion_point()],
                 editor.insertion_point(),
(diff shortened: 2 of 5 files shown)


fdncred:
it would be great if we had a test for this to ensure no breakage in the future.

/cc @stormasm

stormasm:
I am not sure this is working correctly as I tested it by modifying your ide_completions example with the following code

```rust
fn add_newline_keybinding(keybindings: &mut Keybindings) {
    // This doesn't work for macOS
    keybindings.add_binding(
        KeyModifiers::ALT,
        KeyCode::Enter,
        ReedlineEvent::Edit(vec![EditCommand::InsertNewline]),
    );
}
```

```rust
add_newline_keybinding(&mut keybindings);
```

similar to the way it works in the demo example...

and then tried typing one letter basically just the "t"

and then add three insert new lines and then hit tab...

See if it works properly on your machine / I am on a Mac

maxomatic458:
should be fixed now 

i found another problem (not specific to my menu)
if you enter a bunch of newlines (so many until the prompt is not visible anymore), then the cursor will enter the menu

here is a screenshot of the columnar menu, basically using the same code as above
grafik

maxomatic458:
> After the third new line is entered I am not seeing test anymore

im not to sure what you mean by that?
can you replicate this in other menus?

stormasm:
So I am seeing strange behavior in the demo example yes....

at the prompt type "h" and then type tab and you see everything as expected....

Now do another test and type "h"

then enter about 5 or 6 new lines and then type "h"

now not everything is there....

stormasm:
So check out the *demo example*

Type an "h" at the prompt and you will see another "h" show up on the screen many lines down probably where the drop down menu ends...

This is a separate issue than the one I mentioned earlier...
But lets address this one first.  Which seems more serious (possibly)...

maxomatic458:
@stormasm so i tried that
[asciicast](https://asciinema.org/a/8kFumdmfLafUSKhETJIVIzwTQ)

but i dont think i was able to replicate it. Did i do something wrong? (did you also try the latest commit?)

if you referred to the cut off at the bottom, then thats intended (you can scroll throught the options that dont fit on the screen)

maxomatic458:
> Type an "h" at the prompt and you will see another "h" show up on the screen many lines down probably where the menu ends...

this might be somewhat intentional, i think thats just a multiline history suggestion. but you cant hit tab to complete it, so i guess thats bugged, and it probably shouldnt even show up if you are not in multiline mode

maxomatic458:
[asciicast](https://asciinema.org/a/zEGYfIVLIlajazUTnKQV7KX5W)

stormasm:
This is your example -> ide_completions.rs modified to show whats going on...

https://github.com/stormasm/nutmp/blob/main/reedline/ide_completions.rs

Note that after the third input line the first line "test" goes away --- this does not seem correct to me...

https://github.com/nushell/reedline/assets/1809991/b826cd5e-44f9-4df0-9139-89d470dddc66

stormasm:
Here is a slightly longer video --- you will note that after the third *InsertNewLine* that test for some reason disappears...


https://github.com/nushell/reedline/assets/1809991/6aeca9d5-40c6-4dee-9cb2-2fef9ea7e71c

maxomatic458:
as soon as i enter a newline i get "NO RECORDS FOUND" until i enter text in the current line,

i start the example
and i press these buttons
t, tab, ALT+ENTER

i also tried this on the columnar menu and it does the same

did you try if this happens for other menus as well?

maxomatic458:
ok i was able to replicate it, by using the unix linebreaks
the same thing happens on the columnar menu as well

maxomatic458:
@stormasm 
ok so it looks like this line here
https://github.com/nushell/reedline/blob/31eaeeb231664b889c92595ff38ead4ccb570766/src/menu/columnar_menu.rs#L546
causes this to only happen on unix (only replace "\n", not "\r\n")

changing the line to
```rs
let trimmed_buffer = editor.get_buffer()
    .replace("\r\n", "  ")
    .replace("\n", " ");
```
will lead to it to work on windows as well. but there is still the issue that some of the suggestion dissapear after a few lines (like your clip above)

i think this might be something for a seperate PR (as this isn't caused by the ide menu, and happens in the other menus as well)

should i open up a seperate issue for this?

fdncred:
Ah, line endings. we've seen that time and time again in reedline and nushell, since we're cross platform. nice catch.

Let's see what @stormasm says but imo, I think I'd put this change
```rust
let trimmed_buffer = editor.get_buffer()
    .replace("\r\n", "  ")
    .replace("\n", " ");
```
in a separate PR in case we have to revert it (or this current PR) for some odd reason.

maxomatic458:
just squeezed this in here as well in preperation for full support to nushell

stormasm:
> Ah, line endings. we've seen that time and time again in reedline and nushell, since we're cross platform. nice catch.
> 
> Let's see what @stormasm says but imo, I think I'd put this change
> 
> ```rust
> let trimmed_buffer = editor.get_buffer()
>     .replace("\r\n", "  ")
>     .replace("\n", " ");
> ```
> 
> in a separate PR in case we have to revert it (or this current PR) for some odd reason.

I will defer to @fdncred for the call on what to do here as I don't know what the best way to go is...

fdncred:
@maxomatic458 can you open another pr for that line ending change? We'll land that real quick.

maxomatic458:
@fdncred i just made #709

fdncred:
does this need to be rebased to accomodate the line ending pr?

maxomatic458:
@fdncred i don't think so

fdncred:
Ok. Let's give it a try.

PR #712, opened 2024-01-22
Clipboard graceful handling

abusch:
If the `system_clipboard` feature is enabled, gracefully handle the case where the system clipboard can't be accessed and default to `LocalClipboard` in that case instead of panicking. This should help with running nu scripts in a docker container for instance.

### DIFF
--- a/src/core_editor/clip_buffer.rs
+++ b/src/core_editor/clip_buffer.rs
@@ -56,21 +56,29 @@ pub use system_clipboard::SystemClipboard;
 #[cfg(feature = "system_clipboard")]
 /// Helper to get a clipboard based on the `system_clipboard` feature flag:
 ///
-/// Enabled -> [`SystemClipboard`], which talks to the system
+/// Enabled -> [`SystemClipboard`], which talks to the system. If the system clipboard can't be
+/// accessed, it will default to [`LocalClipboard`].
 ///
 /// Disabled -> [`LocalClipboard`], which supports cutting and pasting limited to the [`crate::Reedline`] instance
-pub fn get_default_clipboard() -> SystemClipboard {
-    SystemClipboard::new()
+pub fn get_default_clipboard() -> Box<dyn Clipboard> {
+    SystemClipboard::new().map_or_else(
+        |_e| {
+            eprintln!("Defaulting to local clipboard!");
+            Box::new(LocalClipboard::new()) as Box<dyn Clipboard>
+        },
+        |cb| Box::new(cb),
+    )
 }
 
 #[cfg(not(feature = "system_clipboard"))]
 /// Helper to get a clipboard based on the `system_clipboard` feature flag:
 ///
-/// Enabled -> `SystemClipboard`, which talks to the system
+/// Enabled -> `SystemClipboard`, which talks to the system. If the system clipboard can't be
+/// accessed, it will default to [`LocalClipboard`].
 ///
 /// Disabled -> [`LocalClipboard`], which supports cutting and pasting limited to the [`crate::Reedline`] instance
-pub fn get_default_clipboard() -> LocalClipboard {
-    LocalClipboard::new()
+pub fn get_default_clipboard() -> Box<dyn Clipboard> {
+    Box::new(LocalClipboard::new())
 }
 
 #[cfg(feature = "system_clipboard")]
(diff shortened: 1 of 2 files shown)


kubouch:
It seems like adding the system_clipboard feature job to the CI caused the failure. I'm not sure why the Wayland dependency requires a nightly, not how to fix it. Is the extra CI step necessary? It would be great to land this because currently the latest Nushell built from main panics when launched without a window manager.

Tastaturtaste:
Hey, I had the same problem in nushell. See this comment: https://github.com/nushell/nushell/pull/11535#issuecomment-1901096877

Tastaturtaste:
> and default to LocalClipboard in that case instead of panicking

Please provide a big fat warning for the user in that case, I would be quite surprised if that happened silently to me

abusch:
> Is the extra CI step necessary?

No, I thought it was an oversight that the feature had been missed, but I think it might have been on purpose. There are several tests that use the clipboard, so when run in parallel they step on each other's toes and cause failures. The system clipboard is basically a big global mutable state. I think I'll remove the feature from the CI matrix.

abusch:
> > and default to LocalClipboard in that case instead of panicking
> 
> Please provide a big fat warning for the user in that case, I would be quite surprised if that happened silently to me

I do have a `eprintln!("Defaulting to local clipboard!");` when that happens, but I'm not sure if that's the best way.

Also, I rebased my branch on `main` and now I can't test it with nushell due to conflicting versions of `nu-ansi-term`. I'll have to wait until that's resolved to see if the warning gets printed.

nibon7:
> I do have a `eprintln!("Defaulting to local clipboard!");` when that happens, but I'm not sure if that's the best way.

Nushell prints this message every time when `XWayland` is not started, I think it's quite annoying :(

abusch:
> > I do have a `eprintln!("Defaulting to local clipboard!");` when that happens, but I'm not sure if that's the best way.
> 
> Nushell prints this message every time when `XWayland` is not started, I think it's quite annoying :(

oh, does nushell recreate the whole reedline instance for every prompt? 🤔

kubouch:
Yes, I believe that it does (you can check `nu-cli/src/repl.rs` in Nushell). I'd also prefer to have this message removed because there seems to be no way to turn it off. I haven't tested it myself, though. @nibon7 does it print the line on every prompt?

Instead of eprintln!, it would be better to be able to somehow query information about Reedline where you could check what is your current clipboard, for example.

abusch:
> Yes, I believe that it does (you can check `nu-cli/src/repl.rs` in Nushell). I'd also prefer to have this message removed because there seems to be no way to turn it off. I haven't tested it myself, though. @nibon7 does it print the line on every prompt?

From what I can tell, it doesn't (the reedline instance seems to be created outside of the repl loop).

> Instead of eprintln!, it would be better to be able to somehow query information about Reedline where you could check what is your current clipboard, for example.

To be honest, I don't think I meant to commit that `eprintln!` 😆 so I'm happy to remove it.

I think the only potentially confusing case would be if you _are_ running `nu` in a graphical environment but there is something broken in your setup somehow which would make it silently default to the local clipboard. I guess in this case being able to query the clipboard type could be useful...

abusch:
An alternative might be to expose a `no_system_clipboard` config option / CLI switch in `nushell` that would make it always use the local clipboard (even if compiled with the `system_clipboard` feature), which would be how you'd turn the warning message off. But that's likely quite a bit more work 😅

kubouch:
> From what I can tell, it doesn't (the reedline instance seems to be created outside of the repl loop).

There is `Reedline::create()` also inside the loop, but I haven't checked the code carefully.

Could you then remove the eprintln! so that we can merge it? We can then iterate on some better diagnostics, but let's have the unwrap fix merged first because it's urgent.

abusch:
Done.

fdncred:
I'm wondering if all this is because of our choice to use `arboard` vs `copypasta`? Any guess if `copypasta` would have the same issues? @Tastaturtaste did you happen to test with both?

Tastaturtaste:
If by "all of this" you mean having to handle the absence of an available system clipboard and in particular on Linux CI, then I am pretty certain it would be the same with `copypasta`. 
I didn't test with both, but I took a look at both of their CIs and they both specifically exclude Linux. I don't think they do that just for giggles.

fdncred:
ya, sorry i wasn't more clear, "all of this" means the ci failures we're having. thanks for the info.

Issue #706, opened 2024-01-18
Refactor Menu System with Composition of Menu Functions


stormasm:
#### columnar_menu.rs

```rust
    pub fn with_selected_text_style(mut self, selected_text_style: Style) -> Self {
        self.color.selected_text_style = selected_text_style;
        self
    }
```

#### list_menu.rs

```rust
   pub fn with_selected_text_style(mut self, selected_text_style: Style) -> Self {
        self.color.selected_text_style = selected_text_style;
        self
    }
```
#### ide_menu.rs

```rust
    pub fn with_selected_text_style(mut self, selected_text_style: Style) -> Self {
        self.color.selected_text_style = selected_text_style;
        self
    }
```

All three of our menu types have the exact same code in their implementations...

And this is true across all of our menu_functions lending itself to *massive* code duplication in our menu system.

As we continue to add new types of menus its only going to get worse, meaning in the future when we want to
refactor the menu system its going to be a lot more work...

Ideally we would compose this functionality so we only have one copy of the above code instead of three copies
and the composition style would lead to just adding new code for the functions that are unique to a particular menu
type.

ysthakur:
To add on to this, `only_buffer_difference` is also handled pretty much the same way in all the menus, including `description_menu`, I think

ysthakur:
I don't know if there's already a discussion going somewhere, but in case there isn't, I made a [HackMD](https://hackmd.io/@ysthakur/H1vZgA0Fp/edit) for brainstorming, would love to get more of your opinions @stormasm (cc: @maxomatic458, since you're working on improving completers) (also, just to clarify, I wasn't planning on making a PR all by myself anytime soon, just wanted to be involved in the discussion 🙂)

Moving the description menu into the same crate as the other menus makes sense to me too. Is there any reason to make a separate crate just for the `Menu` trait and its implementations, though? (I figured such a crate wouldn't be used by anyone other Reedline and Reedline users)

stormasm:
I added your HackMd document here as well so it can be found easier in the next period of time...

https://hackmd.io/gDNAvI-8TDyubogworDyCA

ysthakur:
Ah, a line-editor-agnostic menu system sounds cool

maxomatic458:
> cc: @maxomatic458, since you're working on improving completers

im probably not making more changes to the completer anytime soon. I just added a function for convenience that menus can make use of.

All the menus also have a lot of fields in common (which could be compacted in a struct).

[...]

PR #713, opened 2024-01-22
get back the ranges of the strings from the completer used for generating completions

### LINKED ISSUE #711: make the completer return the strings (or positions of them) that are used for completing
I think this [the complete method](https://github.com/nushell/reedline/blob/d9db6a84b70d182c11c1a657d363c914eb3921d1/src/completion/base.rs#L32) could be changed, so it will also return the strings that were used to generate the Suggestions.

something like this:
```rs
fn complete(&mut self, line: &str, pos: usize) -> (Vec<Suggestion>, Vec<String>);
```

for example:
buffer: "this is t"
calling .complete could return the suggestions ("this is the reedline crate", "test", ...) alongside the strings that were used to calculate the suggestions (in this case "this is t", "t")

this info would be useful for menus (for example for greying out text already typed), or in the case of the new ide menu, change the position so it lines up with the typed text (instead of just the cursor)

maxomatic458:
closes #711
adds the function ``complete_with_base_ranges`` to the Completer trait, that returns the ranges, of the strings that the Suggestions were calculated from.

the ide menu can now use of this, to adjust the position of the suggestions to the typed text matches the suggestion.
grafik

### DIFF
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -679,7 +679,7 @@ impl Menu for IdeMenu {
             // editing a multiline buffer.
             // Also, by replacing the new line character with a space, the insert
             // position is maintain in the line buffer.
-            let trimmed_buffer = editor.get_buffer().replace("\r\n", "  ").replace('\n', " ");
+            let trimmed_buffer = editor.get_buffer().replace('\n', " ");
             completer.complete(
                 &trimmed_buffer[..editor.insertion_point()],
                 editor.insertion_point(),
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -679,7 +679,7 @@ impl Menu for IdeMenu {
             // editing a multiline buffer.
             // Also, by replacing the new line character with a space, the insert
             // position is maintain in the line buffer.
-            let trimmed_buffer = editor.get_buffer().replace('\n', " ");
+            let trimmed_buffer = editor.get_buffer().replace("\r\n", "  ").replace('\n', " ");
             completer.complete(
                 &trimmed_buffer[..editor.insertion_point()],
                 editor.insertion_point(),
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -11,6 +11,7 @@ use nu_ansi_term::{ansi::RESET, Style};
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
+/// The direction of the description box
 pub enum DescriptionMode {
     /// Description is always shown on the left
     Left,
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -75,6 +75,14 @@ struct DefaultIdeMenuDetails {
     pub max_description_height: u16,
     /// Offset from the suggestion box to the description box
     pub description_offset: u16,
+    /// If true, the cursor pos will be corrected, so the suggestions match up with the typed text
+    /// ```text
+    /// C:\> str
+    ///      str join
+    ///      str trim
+    ///      str split
+    /// ```
+    pub correct_cursor_pos: bool,
 }
 
 impl Default for DefaultIdeMenuDetails {
@@ -91,6 +99,7 @@ impl Default for DefaultIdeMenuDetails {
             max_description_width: 50,
             max_description_height: 10,
             description_offset: 1,
+            correct_cursor_pos: false,
         }
     }
 }
@@ -114,6 +123,9 @@ struct IdeMenuDetails {
     pub space_right: u16,
     /// Corrected description offset, based on the available space
     pub description_offset: u16,
+    /// The ranges of the strings, the suggestions are based on (ranges in [`Editor::get_buffer`])
+    /// This is required to adjust the suggestion boxes position, when `correct_cursor_pos` in [`DefaultIdeMenuDetails`] is true
+    pub base_strings: Vec<String>,
 }
 
 /// Menu to present suggestions like similar to Ide completion menus
@@ -662,16 +674,16 @@ impl Menu for IdeMenu {
 
     /// Update menu values
     fn update_values(&mut self, editor: &mut Editor, completer: &mut dyn Completer) {
-        self.values = if self.only_buffer_difference {
+        let (values, base_ranges) = if self.only_buffer_difference {
             if let Some(old_string) = &self.input {
                 let (start, input) = string_difference(editor.get_buffer(), old_string);
                 if !input.is_empty() {
-                    completer.complete(input, start + input.len())
+                    completer.complete_with_base_ranges(input, start + input.len())
                 } else {
-                    completer.complete("", editor.insertion_point())
+                    completer.complete_with_base_ranges("", editor.insertion_point())
                 }
             } else {
-                completer.complete("", editor.insertion_point())
+                completer.complete_with_base_ranges("", editor.insertion_point())
             }
         } else {
             // If there is a new line character in the line buffer, the completer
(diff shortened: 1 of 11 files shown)


stormasm:
https://github.com/nushell/reedline/issues/706#issuecomment-1904433003

This is another reason why consolidating all of our menu code in one place is going to make the testing and landing of PRs much simpler going forward...

This change will probably impact https://github.com/nushell/nushell/blob/main/crates/nu-cli/src/menus/description_menu.rs

Meaning once we land this PR nushell may no longer compile ---- having all of our menu code in one place is probably the way to go...

maxomatic458:
> This change will probably impact https://github.com/nushell/nushell/blob/main/crates/nu-cli/src/menus/description_menu.rs
> 
> Meaning once we land this PR nushell may no longer compile ---- having all of our menu code in one place is probably the way to go...

I didnt really mess with the menus in this one, i added https://github.com/maxomatic458/reedline/blob/5f65d771873b3a15b862604746106bd2bfec4a84/src/completion/base.rs#L36-L50

and then used that in the ide menu

nushell should compile fine

but i agree in the future its probably better to first rework the menus before adding new ones

src/menu/mod.rs
stormasm, line 1: Nice addition for simplicity....

stormasm:
> > This change will probably impact https://github.com/nushell/nushell/blob/main/crates/nu-cli/src/menus/description_menu.rs
> > Meaning once we land this PR nushell may no longer compile ---- having all of our menu code in one place is probably the way to go...
> 
> I didnt really mess with the menus in this one, i added https://github.com/maxomatic458/reedline/blob/5f65d771873b3a15b862604746106bd2bfec4a84/src/completion/base.rs#L36-L50
> 
> and then used that in the ide menu
> 
> nushell should compile fine
> 
> but i agree in the future its probably better to first rework the menus before adding new ones

OK, now I understand this better !  Thanks...

The description_menu will still compile fine --- I can get in there after this PR lands and modify the description menu in nushell similar to the way you modified  the columnar and list menu removing the unused code ?

maxomatic458:
@stormasm yes, that should work just fine

stormasm:
@maxomatic458 Great work on this PR as it simplifies the code further as well as makes the end user of your API have more options for tweaking the menus they implement...

Can you please add in some more "scenarios" to ide_completions.rs example....

```rust
   let commands = vec![
        "test".into(),
        "hello world".into(),
        "hello world reedline".into(),
        "this is the reedline crate".into(),
    ];
 ```

just so I can test some other types of cases as well...   I saw the test you added in too which is great...

Also --- maybe we can add in the ability to have some new lines just so we can see how the code behaves if you throw some new lines in there as well ??

Besides the tests you added --- its kind of cool and helpful to be able to play with your code a bit more with an expanded example that gives the end user more options to understand and test further the cool concepts you have implemented...

```rust
fn add_newline_keybinding(keybindings: &mut Keybindings) {
    // This doesn't work for macOS
    keybindings.add_binding(
        KeyModifiers::ALT,
        KeyCode::Enter,
        ReedlineEvent::Edit(vec![EditCommand::InsertNewline]),
    );
}
```

similar to -> https://github.com/stormasm/nutmp/blob/main/reedline/ide_completions.rs

stormasm:
Once your PR lands obviously we will have a lot more abilities to test it with the nushell code...

But for the folks who just use reedline as end users and are not using nushell at all this makes the example more robust...

stormasm:
So here is the idea that I have --- let me know what you think...

In your expanded example or new example outline and show the scenario to the end user of two cases:

1) what is a situation where they will need to have 

```rust
           correct_cursor_pos: false
```

2) what is a situation where 

```rust
           correct_cursor_pos: true
```

This way the end user can have a better mental model of whats going on...  Thanks !

maxomatic458:
can i just add those variables in the example, like so

```rs

    let min_completion_width = 0;
    let max_completion_width = 50;
    let max_completion_height = u16::MAX;
    let padding = 0;
    let border = false;
    let cursor_offset = 0;
    let description_mode = DescriptionMode::PreferRight;
    let min_description_width = 0;
    let max_description_width = 50;
    let description_offset = 1;
    let correct_cursor_pos = false;

    let commands = vec![
        "test".into(),
        "hello world".into(),
        "hello world reedline".into(),
        "this is the reedline crate".into(),
    ];
    let completer = Box::new(DefaultCompleter::new_with_wordlen(commands, 2));

    let mut ide_menu = IdeMenu::default()
        .with_name("completion_menu")
        .with_min_completion_width(min_completion_width)
        .with_max_completion_width(max_completion_width)
        .with_max_completion_height(max_completion_height)
        .with_padding(padding)
        .with_cursor_offset(cursor_offset)
        .with_description_mode(description_mode)
        .with_min_description_width(min_description_width)
        .with_max_description_width(max_description_width)
        .with_description_offset(description_offset)
        .with_correct_cursor_pos(correct_cursor_pos);
        
   ...
```

so someone could just change them? (ill also add some comments then)

src/menu/ide_menu.rs
```diff
@@ -680,12 +699,18 @@ impl Menu for IdeMenu {
             // Also, by replacing the new line character with a space, the insert
             // position is maintain in the line buffer.
             let trimmed_buffer = editor.get_buffer().replace("\r\n", "  ").replace('\n', " ");
-            completer.complete(
+            completer.complete_with_base_ranges(
```
stormasm, line 702: The expanded example with the added ability to throw some new lines in there will test this part of the code better

stormasm:
> can i just add those variables in the example, like so
> 
> ```rust
>     let min_completion_width = 0;
>     let max_completion_width = 50;
>     let max_completion_height = u16::MAX;
>     let padding = 0;
>     let border = false;
>     let cursor_offset = 0;
>     let description_mode = DescriptionMode::PreferRight;
>     let min_description_width = 0;
>     let max_description_width = 50;
>     let description_offset = 1;
>     let correct_cursor_pos = false;
> 
>     let commands = vec![
>         "test".into(),
>         "hello world".into(),
>         "hello world reedline".into(),
>         "this is the reedline crate".into(),
>     ];
>     let completer = Box::new(DefaultCompleter::new_with_wordlen(commands, 2));
> 
>     let mut ide_menu = IdeMenu::default()
>         .with_name("completion_menu")
>         .with_min_completion_width(min_completion_width)
>         .with_max_completion_width(max_completion_width)
>         .with_max_completion_height(max_completion_height)
>         .with_padding(padding)
>         .with_cursor_offset(cursor_offset)
>         .with_description_mode(description_mode)
>         .with_min_description_width(min_description_width)
>         .with_max_description_width(max_description_width)
>         .with_description_offset(description_offset)
>         .with_correct_cursor_pos(correct_cursor_pos);
>         
>    ...
> ```
> 
> so someone could just change them? (ill also add some comments then)

sounds perfect, thank you !

maxomatic458:
@stormasm quick question
```rs
    /// Menu builder with the default border value
    #[must_use]
    pub fn with_default_border(mut self) -> Self {
        self.default_details.border = Some(BorderSymbols::default());
        self
    }
```
i have this function in the ide menu builder, do you think i should change it to this?

```rs
    /// Menu builder with the default border value
    #[must_use]
    pub fn with_default_border(mut self, default_border: bool) -> Self {
        self.default_details.border = if default_border {
            Some(BorderSymbols::default())
        } else {
            None
        }
        self
    }
```
would be a breaking change for nushell, but i would probably do a PR there anyways to load the new config setting for the ide menu

stormasm:
> @stormasm quick question
> 
> ```rust
>     /// Menu builder with the default border value
>     #[must_use]
>     pub fn with_default_border(mut self) -> Self {
>         self.default_details.border = Some(BorderSymbols::default());
>         self
>     }
> ```
> 
> i have this function in the ide menu builder, do you think i should change it to this?
> 
> ```rust
>     /// Menu builder with the default border value
>     #[must_use]
>     pub fn with_default_border(mut self, default_border: bool) -> Self {
>         self.default_details.border = if default_border {
>             Some(BorderSymbols::default())
>         } else {
>             None
>         }
>         self
>     }
> ```
> 
> would be a breaking change for nushell, but i would probably do a PR there anyways to load the new config setting for the ide menu

Lets leave it alone for now and not make the change....
We can add it in later if we want...

maxomatic458:
ok, then i think im done with the PR,
i also changed the default completions example (if we get more menus or more config settings, it might be a good idea to just have one example for them)

stormasm:
@maxomatic458 thanks for making the requested changes in the examples... good job 😄

PR #715, opened 2024-01-22
Remove conflicting keybindings added with selection feature

Tastaturtaste:
## Removes some conflicting keybindings introduced with https://github.com/nushell/reedline/pull/689
Fixes issue https://github.com/nushell/nushell/issues/11600

The issue was that Ctrl+C was bound to copy the selected test after the already existing CtrlC command.

### DIFF
--- a/src/edit_mode/keybindings.rs
+++ b/src/edit_mode/keybindings.rs
@@ -156,11 +156,6 @@ pub fn add_common_navigation_bindings(kb: &mut Keybindings) {
         edit_bind(EC::MoveToLineStart { select: false }),
     );
     kb.add_binding(
-        KM::CONTROL,
-        KC::Char('a'),
-        edit_bind(EC::MoveToLineStart { select: false }),
-    );
-    kb.add_binding(
         KM::NONE,
         KC::End,
         ReedlineEvent::UntilFound(vec![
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks!

src/edit_mode/keybindings.rs
```diff
@@ -214,10 +209,18 @@ pub fn add_common_edit_bindings(kb: &mut Keybindings) {
...
+        KC::Char('x'),
+        edit_bind(EC::CutSelection),
+    );
+    kb.add_binding(
+        KM::CONTROL | KM::SHIFT,
+        KC::Char('c'),
+        edit_bind(EC::CopySelection),
+    );
+    kb.add_binding(
+        KM::CONTROL | KM::SHIFT,
```
sholderbach, line 223: While not in conflict with any of the reedline default emacs-inspired bindings, most GUI terminal emulators that provide their own independent means of selection will intercept `Ctrl-Shift-C` and `Ctrl-Shift-V`.

fdncred: What is your suggestion @sholderbach?

Here's my brief survey of the terminals I have installed on my mac. Some of them do selection, some won't. Some see ctrl-shift-anything as ctrl-anything according to keybindings listen. I think some could be made to work by disabling built-in ctrl-shift-anything.

Ghostty - works
Alacritty - works
iterm2 - does not work
wezterm - does not work
terminal.app - does not work
tabby - does not work
warp - does not work, selection doesn't even work
cool-retro-term - does not work, selection doesn't even work
kitty - does not work
rio - does not work
zed built-in terminal - does not work
vscode built-in terminal - does not work

sholderbach: The number of free keybinds which are part of the lowest common denominator accross the different terminals is certainly limited. Pretty likely that we would have to settle for something pretty unidiomatic/non-mnemonic if we want to squeeze it among everything people expect from bash readline/emacs. Maybe we would at some point add an additional editmode (that is not modal) to cover more GUI-like keybindings.

Tastaturtaste: Would it make sense to support a "pure" shell mode? Without emacs or vim bindings? I personally want to use nushell first and foremost as a shell. For my use case it would prioritise stuff like selection, history and stuff like that much higher than replicating terminal editor functionality.

fdncred: That sounds interesting. How you would do that?

PR #722, opened 2024-01-25
add style from #691 to ide menu

maxomatic458:


maxomatic458:
@fdncred here is a screenshot
grafik
(my terminal colors make it look a bit dark, but its the same as in #691)

stormasm:
@maxomatic458 just curious to know why when you do PRs do you work off your main branch ?

I think it would be easier if you created a branch first and then started working on the PR...

This way if one of your PRs doesn't land right away you can update your main to the latest version of Reedline (for example) and then start another PR by branching off main....

maxomatic458:
@stormasm 
Yea i was to lazy to make a branch for a small pr like this.

And i forgot that i was still on main when starting to work on refactoring the menus.

I should probably start doing that.

stormasm:
@maxomatic458 can you please post the style setting along with the config.nu stuff surrounding it so I can test it with my nushell config.nu 

Thanks !

maxomatic458:
@stormasm i think this only works with external completers (so nothing needs to be changed in nushells config) i used https://carapace.sh

src/menu/ide_menu.rs
```diff
@@ -559,8 +559,13 @@ impl IdeMenu {
         if use_ansi_coloring {
             if index == self.index() {
                 format!(
-                    "{}{}{}{}{}{}{}",
+                    "{}{}{}{}{}{}{}{}",
                     vertical_border,
+                    suggestion
+                        .style
+                        .unwrap_or(self.color.text_style)
+                        .reverse()
```
fdncred, line 567: Seems like this has the same problem that I had with rsteube's PR, it's hard coding `reverse()` which would ignore any setting you have in nushell. it should also be implemented in nushell since it's just adding a style.

Shouldn't it be implemented more like this? https://github.com/nushell/reedline/pull/720/files

maxomatic458: thanks for catching that!
i just made #725

fdncred: thank you!

PR #723, opened 2024-01-26
menu refactor

maxomatic458:
remove code duplication in the menus using the ``MenuCommon`` struct and a ``MenuBuilder`` trait.
related #706.

### DIFF
--- a/src/completion/default.rs
+++ b/src/completion/default.rs
@@ -356,10 +356,10 @@ impl CompletionNode {
 
 #[cfg(test)]
 mod tests {
+    use super::*;
+    use pretty_assertions::assert_eq;
     #[test]
     fn default_completer_with_non_ansi() {
-        use super::*;
-
         let mut completions = DefaultCompleter::default();
         completions.insert(
             ["ｎｕｓｈｅｌｌ", "ｎｕｌｌ", "ｎｕｍｂｅｒ"]
@@ -393,6 +393,53 @@ mod tests {
                     append_whitespace: false,
                 },
             ]
+        );
+    }
+
+    #[test]
+    fn default_completer_with_start_strings() {
+        let mut completions = DefaultCompleter::default();
+        completions.insert(
+            ["this is the reedline crate", "test"]
+                .iter()
+                .map(|s| s.to_string())
+                .collect(),
+        );
+
+        let buffer = "this is t";
+
+        let (suggestions, ranges) = completions.complete_with_base_ranges(buffer, 9);
+        assert_eq!(
+            suggestions,
+            [
+                Suggestion {
+                    value: "test".into(),
+                    description: None,
+                    extra: None,
+                    span: Span { start: 8, end: 9 },
+                    append_whitespace: false,
+                },
+                Suggestion {
+                    value: "this is the reedline crate".into(),
+                    description: None,
+                    extra: None,
+                    span: Span { start: 8, end: 9 },
+                    append_whitespace: false,
+                },
+                Suggestion {
+                    value: "this is the reedline crate".into(),
+                    description: None,
+                    extra: None,
+                    span: Span { start: 0, end: 9 },
+                    append_whitespace: false,
+                },
+            ]
+        );
+
+        assert_eq!(ranges, [8..9, 0..9]);
+        assert_eq!(
+            ["t", "this is t"],
+            [&buffer[ranges[0].clone()], &buffer[ranges[1].clone()]]
         );
     }
 }
--- a/src/completion/default.rs
+++ b/src/completion/default.rs
@@ -55,17 +55,17 @@ impl Completer for DefaultCompleter {
     /// assert_eq!(
     ///     completions.complete("bat",3),
     ///     vec![
-    ///         Suggestion {value: "batcave".into(), description: None, extra: None, span: Span { start: 0, end: 3 }, append_whitespace: false},
-    ///         Suggestion {value: "batman".into(), description: None, extra: None, span: Span { start: 0, end: 3 }, append_whitespace: false},
-    ///         Suggestion {value: "batmobile".into(), description: None, extra: None, span: Span { start: 0, end: 3 }, append_whitespace: false},
+    ///         Suggestion {value: "batcave".into(), description: None, style: None, extra: None, span: Span { start: 0, end: 3 }, append_whitespace: false},
+    ///         Suggestion {value: "batman".into(), description: None, style: None, extra: None, span: Span { start: 0, end: 3 }, append_whitespace: false},
+    ///         Suggestion {value: "batmobile".into(), description: None, style: None, extra: None, span: Span { start: 0, end: 3 }, append_whitespace: false},
     ///     ]);
     ///
     /// assert_eq!(
     ///     completions.complete("to the bat",10),
     ///     vec![
-    ///         Suggestion {value: "batcave".into(), description: None, extra: None, span: Span { start: 7, end: 10 }, append_whitespace: false},
-    ///         Suggestion {value: "batman".into(), description: None, extra: None, span: Span { start: 7, end: 10 }, append_whitespace: false},
-    ///         Suggestion {value: "batmobile".into(), description: None, extra: None, span: Span { start: 7, end: 10 }, append_whitespace: false},
+    ///         Suggestion {value: "batcave".into(), description: None, style: None, extra: None, span: Span { start: 7, end: 10 }, append_whitespace: false},
+    ///         Suggestion {value: "batman".into(), description: None, style: None, extra: None, span: Span { start: 7, end: 10 }, append_whitespace: false},
+    ///         Suggestion {value: "batmobile".into(), description: None, style: None, extra: None, span: Span { start: 7, end: 10 }, append_whitespace: false},
     ///     ]);
     /// ```
     fn complete(&mut self, line: &str, pos: usize) -> Vec<Suggestion> {
@@ -102,6 +102,7 @@ impl Completer for DefaultCompleter {
                                     Suggestion {
                                         value: format!("{span_line}{ext}"),
                                         description: None,
+                                        style: None,
                                         extra: None,
                                         span,
                                         append_whitespace: false,
@@ -176,15 +177,15 @@ impl DefaultCompleter {
     /// completions.insert(vec!["test-hyphen","test_underscore"].iter().map(|s| s.to_string()).collect());
     /// assert_eq!(
     ///     completions.complete("te",2),
-    ///     vec![Suggestion {value: "test".into(), description: None, extra: None, span: Span { start: 0, end: 2 }, append_whitespace: false}]);
+    ///     vec![Suggestion {value: "test".into(), description: None, style: None, extra: None, span: Span { start: 0, end: 2 }, append_whitespace: false}]);
     ///
     /// let mut completions = DefaultCompleter::with_inclusions(&['-', '_']);
     /// completions.insert(vec!["test-hyphen","test_underscore"].iter().map(|s| s.to_string()).collect());
     /// assert_eq!(
     ///     completions.complete("te",2),
     ///     vec![
-    ///         Suggestion {value: "test-hyphen".into(), description: None, extra: None, span: Span { start: 0, end: 2 }, append_whitespace: false},
-    ///         Suggestion {value: "test_underscore".into(), description: None, extra: None, span: Span { start: 0, end: 2 }, append_whitespace: false},
+    ///         Suggestion {value: "test-hyphen".into(), description: None, style: None, extra: None, span: Span { start: 0, end: 2 }, append_whitespace: false},
+    ///         Suggestion {value: "test_underscore".into(), description: None, style: None, extra: None, span: Span { start: 0, end: 2 }, append_whitespace: false},
     ///     ]);
     /// ```
     pub fn with_inclusions(incl: &[char]) -> Self {
@@ -374,6 +375,7 @@ mod tests {
                 Suggestion {
                     value: "ｎｕｌｌ".into(),
                     description: None,
+                    style: None,
                     extra: None,
                     span: Span { start: 0, end: 3 },
                     append_whitespace: false,
@@ -381,6 +383,7 @@ mod tests {
                 Suggestion {
                     value: "ｎｕｍｂｅｒ".into(),
                     description: None,
+                    style: None,
                     extra: None,
                     span: Span { start: 0, end: 3 },
                     append_whitespace: false,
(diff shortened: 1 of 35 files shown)


maxomatic458:
the api for the menus is pretty much the same. the only breaking change is that the common builder functions (with_name, with_marker...) are now in a trait.

fdncred:
Thanks @maxomatic458. We'll take a look at this soon.

/cc @ysthakur I think you were thinking about going in this direction too. Your review would be helpful.

ysthakur:
I'll take a look too. I was actually working on that issue too, just hadn't made a PR yet, lol. I only made some helpers, though, so this PR and my branch aren't doing the same thing

maxomatic458:
i mostly got rid of code duplication by moving the builder methods in a trait (which i guess is somewhat a breaking change)
If you are ok with that, then i can move this PR over to @ysthakur's fork.

ysthakur:
That'd be cool. I'd made a `completer_input` helper function to help with `update_values` (gets the input that needs to be passed to the completer), a `replace_in_buffer` helper, and a `can_partially_complete` helper (for the columnar and IDE menus). They're all after [this line](https://github.com/ysthakur/reedline/blob/menu-duplication/src/menu/menu_functions.rs#L248) in `menu_functions.rs`. `completer_input` and `can_partially_complete` could now become methods on `MenuCommon`. Caveat: the `completer_input` helper I made doesn't replace newlines, because I think that might screw up the history completer. Instead, I changed the `DefaultCompleter` to replace `\r` and `\n` with spaces. Other completers would also have to be updated to handle newlines.

Also, to avoid breaking changes, would it be possible to keep `Menu` the same as before, but have a `MenuWithCommon` trait for `Menu`s that have a `MenuCommon`? And then there could be an implementation of `Menu` for any menu that implements `MenuWithCommon`. Not sure if that's doable in Rust. It would still mean that anyone building the columnar, list, or IDE menus would need to use the new `MenuBuilder`, but it'd be less of a breaking change, at least

ysthakur:
> If you are ok with that, then i can move this PR over to @ysthakur's fork.

@maxomatic458 No need, you can take the stuff from my fork instead since this PR is already up

stormasm:
@maxomatic458 @ysthakur once you all are in a good state with this PR I will review and test as well...
thanks for taking this one for a spin 😄 
it will certainly help clean up the code a bit...

stormasm:
@maxomatic458 I reviewed your code so far and it looks really good !  Great job...

maxomatic458:
> Also, to avoid breaking changes, would it be possible to keep `Menu` the same as before, but have a `MenuWithCommon` trait for `Menu`s that have a `MenuCommon`? And then there could be an implementation of `Menu` for any menu that implements `MenuWithCommon`. Not sure if that's doable in Rust. It would still mean that anyone building the columnar, list, or IDE menus would need to use the new `MenuBuilder`, but it'd be less of a breaking change, at least

I think we might just be able to do this here:
```rs
pub trait Menu: Send {
    fn common(&self) -> MenuCommon {
        panic!("Not implemented for this Menu");
    }
    
    fn name(&self) -> &str {
        self.common().name;
    }
    ...
}
```

this would not break any existing menus, (because the functions ``common`` is optional as it has a base implementation and we can expect the menus to already have the ``name`` function implemented, so it will be overwritten).

and this could also then be used in a seperate builder trait, which only only be breaking for the reedline included menus (and would be simply fixed by importing the MenuBuilder trait)

maxomatic458:
@stormasm do you think this panic would be ok here?
if we do it like this we make sure that we stay fully backwards compatible with existing menus

ysthakur:
@maxomatic458 Ah, whoops, I didn't read closely enough to see `name` and other methods were being kept in `Menu`.

I'm a bit nervous about the panicking. Someone could forget to implement `common()` as well as `name()`, `indicator()`, and the other getters. That said, making new menus is rare, and I think it's unlikely someone would implement an entire menu while forgetting to implement both `common` *and* the getters. They'd surely catch it while testing out their menu before the users get it, since nothing would work.

By the way, how about turning `MenuCommon` into `MenuSettings`, and only keeping the setting-like fields in there that don't need to be changed? It would mean each menu maintaining its own `input`, `values`, and `longest_suggestion` fields, but I don't think it's a huge deal, since those don't have setter/builder methods like `with_name()`. You'd then be able to do `self.values` directly instead `self.common.values`, and also get rid of `common_mut()`.

maxomatic458:
>I'm a bit nervous about the panicking. Someone could forget to implement common() as well as name(), indicator(), and the other getters. That said, making new menus is rare, and I think it's unlikely someone would implement an entire menu while forgetting to implement both common and the getters. They'd surely catch it while testing out their menu before the users get it, since nothing would work.

The idea is, for new menus you just implement common and common_mut; the other methods (including all the builder methods, and stuff like name(), marker()) dont need to be implemented again because they have their base implementation already;

old menus would continue to work just fine (because the base implementations are overwritten, and the panic basically makes the function optional)

> By the way, how about turning MenuCommon into MenuSettings, and only keeping the setting-like fields in there that don't need to be changed? It would mean each menu maintaining its own input, values, and longest_suggestion fields, but I don't think it's a huge deal, since those don't have setter/builder methods like with_name(). You'd then be able to do self.values directly instead self.common.values, and also get rid of common_mut().

we could do this (so we only have the values that we actually need for the builder functions for example), but this would still lead to the structs all having the same fields, but i guess thats not too bad.

ysthakur:
> The idea is, for new menus you just implement common and common_mut; the other methods (including all the builder methods, and stuff like name(), marker()) dont need to be implemented again because they have their base implementation already

Right, I was just afraid someone making a new menu wouldn't get warned by their IDE to implement `common` and `common_mut`. But it's probably not a big deal, it would panic immediately when they tested it so they can't really forget.

> but this would still lead to the structs all having the same fields

That's true, only 4 fields would be left in `MenuCommon` then and that does take away from the point of having it in the first place.

> And at least for the builder we need a way to get a mutable reference of this settings object

Ah, didn't realize `common_mut` was still needed for that

stormasm:
@maxomatic458 code is looking good --- if you are ready to start testing nushell I created a branch pointing to your code so you can see the compiler errors...

https://github.com/stormasm/nushell/tree/menu_refactor_01

After reviewing the compiler errors --- it might give you some more ideas as to how to clean up or improve your menu refactor code --- and then once the reedline code is good and you are happy with it...

If you feel like / or when you feel the time is right...

Go ahead and create your own branch off nushell main just like I did and start working on getting nushell to compile...

Once we have a built version of nushell on your own nushell branch then we can start testing the menu stuff in nushell.

Thanks !  This is looking great so far 😄

ysthakur:
@maxomatic458 I made a [PR](https://github.com/nushell/reedline/pull/726/files) from my branch in case you want to copy the helper functions from there into this PR. I haven't actually finished adding tests, though.

I kinda wish `active`, `event`, and `only_buffer_difference`/`input` weren't handled by the menu. The event unfortunately can't be passed straight into `update_working_details` because the list menu needs to know the event when `update_values()` is called (also, breaking change). And making the only buffer difference thing a responsibility of the completer would be pretty tough to implement :(. Removing `is_active()` is the only one that seems doable, although it would break code (but as far as breaking changes go, it's not the worst, since you'd only have to remove the `active` field from your menu). Maybe it could be handled by the engine instead, by keeping one list of inactive menus and a separate list of active menus instead of a single list of all menus.

stormasm:
@ysthakur all great suggestions --- thanks for those and I am sure @maxomatic458 and you will get the code to a place you are both happy with...

We are also not in any hurry to land this code --- its much more important that you feel comfortable with the changes you both are making --- so take as much time as you both need to figure out the best path forward...

I merely did the nushell branch just so we can see so far what compiler errors are showing up on the nushell side of things.

Thanks to both of you for pursuing the refactor --- its looking good so far...

maxomatic458:
Ok, so i made a ``MenuSettings`` struct that looks like this
```rs
pub struct MenuSettings {
    /// Menu name
    name: String,
    /// Menu coloring
    color: MenuTextStyle,
    /// Menu marker when active
    marker: String,
    /// Calls the completer using only the line buffer difference difference
    /// after the menu was activated
    only_buffer_difference: bool,
}
```

which gets rid of all the duplicate builder functions

and i made these changes here to the Menu trait
```rs
pub trait Menu: Send {
    /// Get MenuSettings
    fn settings(&self) -> &MenuSettings;

    /// Menu name
    fn name(&self) -> &str {
        &self.settings().name
    }

    /// Menu indicator
    fn indicator(&self) -> &str {
        &self.settings().marker
    }
    ...
```

to also get rid of those two functions, but this is a breaking change, because now each menu needs a implementation for settings (which could be made optional via the panic approach)

and i also added all the changes from @ysthakur 

and for nushell, you basically just need to bring the MenuBuilder trait into scope and thats pretty much it

stormasm:
@maxomatic458 @ysthakur Super fantastic job !!

I got everything compiling on my nushell branch --- and ran through all of the menu types on the Nushell side of the world...

Everything appears to be working fine...  Although it is preliminary testing, I haven't found any issues so far...

The nushell side of the world could probably use a bit more testing....
before we land this PR  on the Reedline side!

Would you all mind running nushell off my branch which is up and running and working...

And see if you see any oddities / weirdness on the nushell side of the world...

Once you both give me the thumbs up (that you have tested my nushell branch) I will land the Reedline PR...

Then we will finalize the nushell side of the world to make sure that is good...

And then I will land the nushell side after that....

Again --- wow !  really cool 😄 

@fdncred if you feel like doing some testing as well that would be cool...

maxomatic458:
found an unrelated bug  #727 (but thats on that also happens on the current reedline version)
tested all menu types with different configs and didnt find any bugs (apart from that)

fdncred:
Were we able to avoid breaking changes?

maxomatic458:
> Were we able to avoid breaking changes?

currently the only breaking changes are:
* ``MenuBuilder`` trait needs to be imported where a menu is built
* ``Menu`` trait has a new function that needs to be implemented (that could be fixed with a base implementation see above, e.g by panicing)

ysthakur:
@maxomatic458 Everything looks good to me too! I guess we'll have to go with panicking since there's no other obvious ways to avoid a breaking change.

> Were we able to avoid breaking changes?

@fdncred The `completer_input` helper that was added to help with `update_values` doesn't replace newlines with spaces anymore. Either the columnar, IDE, and description menus could do the newline-replacing before calling that helper, or we could leave it to be handled by completers downstream (`DefaultCompleter` in Reedline has already been changed to handle newlines)

src/completion/default.rs
```diff
@@ -75,6 +75,10 @@ impl Completer for DefaultCompleter {
         // `only_buffer_difference` is false
         let line = if line.len() > pos { &line[..pos] } else { line };
         if !line.is_empty() {
+            // When editing a multiline buffer, there can be new line characters in it.
+            // Also, by replacing the new line character with a space, the insert
+            // position is maintain in the line buffer.
+            let line = line.replace("\r\n", " ").replace('\n', " ");
```
ysthakur, line 81:
```suggestion
            let line = line.replace('\r', " ").replace('\n', " ");
```
I misread the [original code](https://github.com/nushell/reedline/blob/main/src/menu/columnar_menu.rs#L551), `\r\n` is supposed to be replaced with two spaces rather than one, sorry about that

maxomatic458: thanks for catching that!

stormasm:
> Were we able to avoid breaking changes?

Breaking changes were pretty much non-existent except for some very minor issues which I already dealt with in my Nushell branch....

maxomatic458:
@stormasm i added a panic, so existing menu implementations (outside of nushell) will not break.
should i leave it in or should i remove it?

stormasm:
> @stormasm i added a panic, so existing menu implementations (outside of nushell) will not break. should i leave it in or should i remove it?

Lets leave it in for now....
And see if anyone complains later 😄 
Then we can worry about it...
And deal with it...
Unless others feel strongly the other way...
@maxomatic458 What do you think we should do ?

stormasm:
@maxomatic458 @ysthakur 
I will plan on landing your PR sometime tomorrow morning (Sunday) my time --- I live in Oregon, USA

I will test your branch more this evening and in the morning... 

Then I will play around with the Reedline main branch in nushell tomorrow post landing...

And plan on updating Nushell with the Reedline main branch Sunday evening or Monday morning...
If no one who is testing finds any issues...

Unless someone disagrees --- or for some reason you all don't feel its ready...

This has been an incredible job !  I am impressed 😄 

Thanks kindly for both of your efforts...
I think the menu code base will be a lot more solid moving forward...
Meaning it will be a lot easier for us to maintain / others to use and understand better
and just overall a much cleaner menu design for the entire Reedline community...

ysthakur:
I'd like to add a couple test cases for the `can_partially_complete` helper before then. It's not a priority, since the columnar menu already has tests for that and we'd just be moving them from there into `menu_functions.rs`, but it might make more sense for tests for that helper to live with the other tests for helpers in `menu_functions.rs` rather than with the tests specific to the columnar menu.

ysthakur:
You know what, never mind, it's fine if the `can_partially_complete` tests live in `columnar_menu.rs`. I don't want to delay this PR further.

fdncred:
we don't have to do it in this PR but it would be great to have more tests. it seems that tests are the only thing these days that guarantees other changes don't affect previous changes. i know reedline is kind of a special case since it's a line editor but we need to figure out how to do more thorough testing and have more coverage in our tests. i don't want to hold this PR up either but it would be nice to get back to tests at some point.

ysthakur:
@fdncred Absolutely, it just so happens that the columnar menu is already thoroughly testing `can_partially_complete`, and I think the other two helpers we added have all their edge cases covered too. There is the edge case of calling `replace_in_buffer` where the insertion point isn't the same as the end of the span that's being replaced, but I'm not sure how that'd happen or what the expected behavior should be in that situation (currently, it moves the insertion point around according to the offset but doesn't mess anything up)

fdncred:
That's good to hear. Thanks.

src/menu/columnar_menu.rs
```diff
@@ -60,70 +60,41 @@ pub struct ColumnarMenu {
...
-    /// Calls the completer using only the line buffer difference difference
-    /// after the menu was activated
-    only_buffer_difference: bool,
 }
 
 impl Default for ColumnarMenu {
     fn default() -> Self {
         Self {
-            name: "columnar_menu".to_string(),
+            settings: MenuSettings::default(),
```
ysthakur, line 74:
```suggestion
            settings: MenuSettings::default().with_name("columnar_menu"),
```

stormasm:
@maxomatic458 @ysthakur We are up and running in nushell with your newest amazing menu code !! 😄

PR #725, opened 2024-01-26
Fix style in ide menu

maxomatic458:


fdncred:
Thanks!

stormasm:
@fdncred thank you for catching that as well...

@maxomatic458 I am not seeing the effect of your change show up on the nushell side with carapace...

Note you have to do a cargo update in nushell to pick up the latest Reedline...

Thanks for giving me an example of how to test your change....

PR #728, opened 2024-01-27
Print foreground color for indicator and right prompt in large buffers

### LINKED ISSUE #727: list menu marker changing color
windows 10; windows terminal.

https://github.com/nushell/reedline/assets/104733404/bb9b1fce-e061-4cb6-a1ce-19f86bbbd9eb

ysthakur:
Fixes #727

Previously, when printing large buffers, the foreground color for the prompt indicator and right prompt wasn't printed.

Below is the original behavior. You can see that with a small buffer that doesn't fill up the whole screen, the menu indicator (`|`) is blue and the right prompt is purple, as expected. But with a large buffer, both become green.

[asciicast](https://asciinema.org/a/D9bu9pnpHYziMyUsrAe8Azmoz)

Below is the new behavior. The menu marker and right prompt are the expected colors even with a large buffer.

[asciicast](https://asciinema.org/a/RHfh4dfkZttJ3VTAHekk69KND)

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -342,12 +342,21 @@ impl Painter {
         self.stdout.queue(Print(&coerce_crlf(prompt_skipped)))?;
 
         if extra_rows == 0 {
+            if use_ansi_coloring {
+                self.stdout
+                    .queue(SetForegroundColor(prompt.get_prompt_right_color()))?;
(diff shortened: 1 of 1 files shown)


fdncred:
great work! and that was really quick. thanks!

PR #729, opened 2024-01-28
Add a way to disable the prompt

rtpg:
This is part of work to simplify [this nushell PR](https://github.com/nushell/nushell/pull/11288). Clearing out the prompt removes a reference to the stack, letting us avoid a copy

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -487,6 +487,12 @@ impl Reedline {
         self.transient_prompt = Some(transient_prompt);
         self
     }
+    /// Remove the current prompt
+    #[must_use]
+    pub fn disable_transient_prompt(mut self) -> Self {
+        self.transient_prompt = None;
+        self
+    }
 
     /// A builder which configures the edit mode for your instance of the Reedline engine
     #[must_use]


stormasm:
@rtpg can you please add in a example that shows this working...  Thanks !

stormasm:
@rtpg as far as the example it can be as simple as the 

https://github.com/nushell/reedline/blob/main/examples/basic.rs

but call it *disable_prompt.rs*

fdncred:
> @rtpg as far as the example it can be as simple as the
> 
> https://github.com/nushell/reedline/blob/main/examples/basic.rs
> 
> but call it _disable_prompt.rs_

I don't really understand what you're saying here. Can you add an example and some test to ensure this behavior works in the future?

src/engine.rs
```diff
@@ -487,6 +487,12 @@ impl Reedline {
         self.transient_prompt = Some(transient_prompt);
         self
     }
+    /// Remove the current prompt
```
fdncred, line 490: Should this be
```suggestion
    /// Remove the current transient prompt
```

fdncred:
unresponsive.

PR #732, opened 2024-01-29
Fix quick completion

nibon7:
Fixes nushell/nushell#11209


nibon7:
It can be reproduced in nushell.

1. touch aaa aba aab
2. ls a<kbd>tab</kbd>b<kbd>tab</kbd>

step 2 doesn't auto select `aba` before this pr.

fdncred:
Makes sense. Thanks for the repro!

PR #736, opened 2024-01-31
Add `History::delete` for `FileBackedHistory`

Hofer-Julian:
- Add `History::delete` for `FileBackedHistory`
- Adapt `history` to print id rather than just consecutive numbers
- Add `history remove-item <id>` to demo

See also https://github.com/nushell/nushell/pull/11629

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -176,7 +176,7 @@ fn main() -> reedline::Result<()> {
                     continue;
                 }
                 // Delete history entry of a certain id
-                if buffer.trim().starts_with("history delete-item") {
+                if buffer.trim().starts_with("history remove-item") {
                     let parts: Vec<&str> = buffer.split_whitespace().collect();
                     if parts.len() == 3 {
                         if let Ok(id) = parts[2].parse::<i64>() {
@@ -184,7 +184,7 @@ fn main() -> reedline::Result<()> {
                             continue;
                         }
                     }
-                    println!("Invalid command. Use: history delete <id>");
+                    println!("Invalid command. Use: history remove-item <id>");
                     continue;
                 }
                 // Get this history session identifier
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -16,7 +16,7 @@ use {
 
 #[cfg(not(any(feature = "sqlite", feature = "sqlite-dynlib")))]
 use reedline::FileBackedHistory;
-use reedline::{CursorConfig, MenuBuilder};
+use reedline::{CursorConfig, HistoryItemId, MenuBuilder};
 
 fn main() -> reedline::Result<()> {
     println!("Ctrl-D to quit");
@@ -173,6 +173,18 @@ fn main() -> reedline::Result<()> {
                 // Get the history only pertinent to the current session
                 if buffer.trim() == "history session" {
                     line_editor.print_history_session()?;
+                    continue;
+                }
+                // Delete history entry of a certain id
+                if buffer.trim().starts_with("history delete-item") {
+                    let parts: Vec<&str> = buffer.split_whitespace().collect();
+                    if parts.len() == 3 {
+                        if let Ok(id) = parts[2].parse::<i64>() {
+                            line_editor.history_mut().delete(HistoryItemId::new(id))?;
+                            continue;
+                        }
+                    }
+                    println!("Invalid command. Use: history delete <id>");
                     continue;
                 }
                 // Get this history session identifier
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -536,8 +536,9 @@ impl Reedline {
             .search(SearchQuery::everything(SearchDirection::Forward, None))
             .expect("todo: error handling");
 
-        for (i, entry) in history.iter().enumerate() {
-            self.print_line(&format!("{}\t{}", i, entry.command_line))?;
+        for entry in history.iter() {
+            let Some(id) = entry.id else { continue };
+            self.print_line(&format!("{}\t{}", id, entry.command_line))?;
         }
         Ok(())
     }
--- a/src/history/file_backed.rs
+++ b/src/history/file_backed.rs
@@ -223,9 +223,11 @@ impl History for FileBackedHistory {
         Ok(())
     }
 
-    /// Writes unwritten history contents to disk.
+    /// Syncs current state with disk.
     ///
+    /// Normally, that means reading entries from the file and appending new entries to it.
     /// If file would exceed `capacity` truncates the oldest entries.
+    /// If necessary, the whole file is overwritten.
     fn sync(&mut self) -> std::io::Result<()> {
         if let Some(fname) = &self.file {
             if let Some(base_dir) = fname.parent() {
--- a/src/history/file_backed.rs
+++ b/src/history/file_backed.rs
@@ -205,13 +205,20 @@ impl History for FileBackedHistory {
         Ok(())
     }
 
-    fn delete(&mut self, _h: super::HistoryItemId) -> Result<()> {
-        Err(ReedlineError(
-            ReedlineErrorVariants::HistoryFeatureUnsupported {
-                history: "FileBackedHistory",
-                feature: "removing entries",
-            },
-        ))
+    fn delete(&mut self, h: super::HistoryItemId) -> Result<()> {
+        let id = h.0 as usize;
+        let num_entries = self.entries.len();
+        // Check if the id is valid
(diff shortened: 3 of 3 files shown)


fdncred:
Trying this out. It looks like `delete-item` may be 1 off?
```nushell
D:\reedline〉history
... others not listed
46      ls
47      ls -r
48      hello another very large option for hello word that will force one column
49      history
D:\reedline〉history delete-item 48                                                               01/31/2024 07:52:21 AM
D:\reedline〉history                                                                              01/31/2024 07:52:27 AM
... others not listed
45      ls
46      ls -r
47      hello another very large option for hello word that will force one column
48      history delete-item 48
49      history
```

Hofer-Julian:
> Trying this out. It looks like `delete-item` may be 1 off? or some other entry was deleted. I'm noticing now that they're not the same numbers anymore.

Ids are not preserved with the file-based history.
If that is of importance, some kind of tombstone system would have to be established.

fdncred:
Right, i understand that the ids are not preserved but if it's counting from 1 to 50 they should always be the same numbers unless something was deleted from within right?

Hofer-Julian:
> Right, i understand that the ids are not preserved but if it's counting from 1 to 50 they should always be the same numbers unless something was deleted from within right?

Yes, looking at your example again, it looks indeed like a one-off error.
Sigh, as they say:
> There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

fdncred:
diffing the two history outputs, when I said `history delete-item 48` it looks like it deleted item 0.

Hofer-Julian:
@fdncred I had another look (with a different machine on a different OS) and it still just works for me...

First, I run `cargo run --example demo`

```
〉history
0       history
1       history
2       history delete-item 1
3       history delete-item 3
4       history
〉history delete-item 0                                                                                                  
〉history
0       history
1       history delete-item 1
2       history delete-item 3
3       history
4       history delete-item 0
5       history
```

I can't seem to manage to get it into the broken state you had

Hofer-Julian:
Also, that just works:

```
〉history 
0       history
1       history delete-item 1
2       history delete-item 3
3       history
4       history delete-item 0
5       history
6       history delete-item 2
7       history
8       history delete-item 2
9       history
〉history delete-item 2
〉history 
0       history
1       history delete-item 1
2       history
3       history delete-item 0
4       history
5       history delete-item 2
6       history
7       history delete-item 2
8       history
9       history delete-item 2
10      history
```

Hofer-Julian:
Could you maybe try one more time? And double check that you run the correct command and have the correct branch checked out?

fdncred:
Sure, no worries.

```
history
0           let-env current_page = 0
1           alias pp = (let cmds = (help commands | range ($env.current_page - 10)..($env.current_page)); let-env current_page = ($env.current_page - 10); echo $cmds)
2           alias hh = (let cmds = (help commands | range ($env.current_page)..($env.current_page + 10));  echo $cmds)
3           alias nn = (let cmds = (help commands | range $env.current_page..($env.current_page + 10)); let-env current_page = ($env.current_page + 10); echo $cmds)
4       exit
5       pwd
...
44      history
45      ls
46      ls -r
47      hello another very large option for hello word that will force one column
48      history delete-item 48
49      history
```
i want to delete 47 which is that long one
```
D:\reedline〉history delete-item 47
D:\reedline〉history
0           alias pp = (let cmds = (help commands | range ($env.current_page - 10)..($env.current_page)); let-env current_page = ($env.current_page - 10); echo $cmds)
1           alias hh = (let cmds = (help commands | range ($env.current_page)..($env.current_page + 10));  echo $cmds)
2           alias nn = (let cmds = (help commands | range $env.current_page..($env.current_page + 10)); let-env current_page = ($env.current_page + 10); echo $cmds)
3       exit
4       pwd
5       exit
...
44      ls
45      ls -r
46      hello another very large option for hello word that will force one column
47      history
48      history delete-item 47
49      history
```
Notice that 47 is now 46 and 0 is different.

I'm testing on Windows. I wonder if that makes a difference due to line endings or something like that?

fdncred:
I just tested on WSL Ubuntu and it works there. Must be a Windows thing?

fdncred:
That worked. Here's my guess. My history is old so I had items in history with leading spaces. Leading spaces aren't permitted anymore. So, maybe we should trim the history items somewhere to prevent this from happening.

Hofer-Julian:
> That worked. Here's my guess. My history is old so I had items in history with leading spaces. Leading spaces aren't permitted anymore. So, maybe we should trim the history items somewhere to prevent this from happening.

Good to hear that this fixed it. Would be good to have a reproducer though to make sure that this is indeed the reason.

fdncred:
I'm not really sure what it is. I can reproduce it every time with that history file I attached above.

Hofer-Julian:
> That worked. Here's my guess. My history is old so I had items in history with leading spaces. Leading spaces aren't permitted anymore. So, maybe we should trim the history items somewhere to prevent this from happening.

Your guess is correct. I was able to reproduce the behavior with your history file and can confirm that the leading whitespace is to blame.

I also found another bug in my implementation, will convert this to a draft for now.

Hofer-Julian:
It should be fine now.

Here's what I changed:
- rename `history delete-item` to `history remove-item` since that's how it will be called on the nushell side https://github.com/nushell/nushell/pull/11629
- add option to overwrite the complete file (needed for the next two points)
- trim entries with leading whitespace and trigger overwrite
- overwrite when history items have been removed

fdncred:
I tested this PR this morning with this history.txt and couldn't get it to work. It just deleted item 0 off the history.
[history.txt](https://github.com/nushell/reedline/files/14221667/history.txt)

Hofer-Julian:
> I tested this PR this morning with this history.txt and couldn't get it to work. It just deleted item 0 off the history. [history.txt](https://github.com/nushell/reedline/files/14221667/history.txt)

I just checked, and the problem is that your history file is at capacity (50).
That means that when you run `history` it gets appended to your history backend and the first entry is pruned, which changes the ids of all entries.
When you then try to delete an entry it deletes a different entry than you intended.

At this point, I see only these ways forward:
- Refactoring the file based history backend altogether (just to be clear, I am not signing up for that :P)
- Add logic, so that the `history` command will not be added to history (a bit hacky but would be fine by me)
- Close this PR, but still merge https://github.com/nushell/nushell/pull/11629 even though only sqlite is supported (also fine by me)
- Close both this PR and https://github.com/nushell/nushell/pull/11629 (my least favorite option)

fdncred:
> That means that when you run history it gets appended to your history backend and the first entry is pruned, which changes the ids of all entries.
When you then try to delete an entry it deletes a different entry than you intended.

Seems like it's good that we caught this but bad because it means more changes are due.

I'm not really a fan of any option but the first one. Seems like this is close to being what we need but not quite. Sorry I found so many problems. It's not my intent to have continual criticisms of your PR. :(

Hofer-Julian:
> > That means that when you run history it gets appended to your history backend and the first entry is pruned, which changes the ids of all entries.
> > When you then try to delete an entry it deletes a different entry than you intended.
> 
> Seems like it's good that we caught this but bad because it means more changes are due.

I very much appreciate that you found these bugs.
For nushell this would have been much less obvious to debug.

> I'm not really a fan of any option but the first one. Seems like this is close to being what we need but not quite.

I don't think it's super close. We would have to introduce the concept of explicit indexes. But just thinking about all the edge cases makes not want to go further into this direction. Especially since I don't care too much about the file backed history, and would much rather have nushell switch completely to sqlite :D

fdncred:
for what it's worth, i'd rather have nushell just be sqlite. i just think there would be outrage if that happened.

Hofer-Julian:
> for what it's worth, i'd rather have nushell just be sqlite. i just think there would be outrage if that happened.

Just for context, I meant changing the default to sqlite, not removing the file-based backend (at least for now :P)
And accepting that not every feature, like deleting a specific entry, is also supported by the file-based backend

Hofer-Julian:
Anyway, my suggestion would be to merge https://github.com/nushell/nushell/pull/11629 and leave this PR open for now.
I assume https://github.com/nushell/reedline/pull/747 changes things too, and I'd hate if both PRs would bit-rot.

Agreed @fdncred?

kronberger-droid:
Closing this since the file backend still has no stable ids,
so a delete at capacity hits the wrong entry, as you found in Feb 2024.

PR #737, opened 2024-01-31
External printer API rework

IanManske:
This PR brushes up the external printer feature and is motivated by [this nushell PR](https://github.com/nushell/nushell/pull/11696). Notable changes include:
- Uses `std::sync::mpsc` instead of `crossbeam-channel`, since `crossbeam-channel` was merged into `std::sync::mpsc` in Rust 1.67.
- Users can send/print raw bytes instead of only `String`s.
- Reduced intermediate allocations using `Receiver::try_iter`.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -677,10 +677,9 @@ impl Reedline {
             #[cfg(feature = "external_printer")]
             if let Some(channel) = &self.external_printer_channel {
                 // get messages from printer as crlf separated "lines"
-                let messages = channel.messages();
-                if !messages.is_empty() {
+                if !channel.receiver().is_empty() {
                     self.painter.print_external_message(
-                        messages,
+                        channel,
                         self.editor.line_buffer(),
                         prompt,
                     )?;
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -676,7 +676,6 @@ impl Reedline {
 
             #[cfg(feature = "external_printer")]
             if let Some(channel) = &self.external_printer_channel {
-                // get messages from printer as crlf separated "lines"
                 if !channel.receiver().is_empty() {
                     self.painter.print_external_message(
                         channel,
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -4,7 +4,7 @@ use itertools::Itertools;
 use nu_ansi_term::{Color, Style};
 
 #[cfg(feature = "external_printer")]
-use crate::ExternalPrinterChannel;
+use crate::ExternalPrinter;
 use crate::{enums::ReedlineRawEvent, CursorConfig};
 #[cfg(feature = "bashisms")]
 use crate::{
@@ -150,7 +150,7 @@ pub struct Reedline {
     kitty_protocol: KittyProtocolGuard,
 
     #[cfg(feature = "external_printer")]
-    external_printer_channel: Option<ExternalPrinterChannel>,
+    external_printer: Option<ExternalPrinter>,
 }
 
 struct BufferEditor {
@@ -221,7 +221,7 @@ impl Reedline {
             bracketed_paste: BracketedPasteGuard::default(),
             kitty_protocol: KittyProtocolGuard::default(),
             #[cfg(feature = "external_printer")]
-            external_printer_channel: None,
+            external_printer: None,
         }
     }
 
@@ -675,13 +675,14 @@ impl Reedline {
             let mut paste_enter_state = false;
 
             #[cfg(feature = "external_printer")]
-            if let Some(channel) = &self.external_printer_channel {
-                if !channel.receiver().is_empty() {
-                    self.painter.print_external_message(
-                        channel,
-                        self.editor.line_buffer(),
-                        prompt,
-                    )?;
+            if let Some(printer) = &self.external_printer {
+                let any_messages = self.painter.print_external_messages(
+                    printer.receiver(),
+                    self.editor.line_buffer(),
+                    prompt,
+                )?;
+
+                if any_messages {
                     self.repaint(prompt)?;
                 }
             }
@@ -1720,22 +1721,22 @@ impl Reedline {
         )
     }
 
-    /// Returns a reference to the external printer's channel, or none if one was not set
+    /// Returns a reference to the external printer, or none if one was not set
     ///
     /// ## Required feature:
     /// `external_printer`
     #[cfg(feature = "external_printer")]
-    pub fn external_printer(&self) -> Option<&ExternalPrinterChannel> {
-        self.external_printer_channel.as_ref()
+    pub fn external_printer(&self) -> Option<&ExternalPrinter> {
+        self.external_printer.as_ref()
     }
 
-    /// Adds an external printer
+    /// Add a new external printer, or overwrite the existing one
     ///
     /// ## Required feature:
     /// `external_printer`
     #[cfg(feature = "external_printer")]
-    pub fn with_external_printer(mut self, channel: ExternalPrinterChannel) -> Self {
-        self.external_printer_channel = Some(channel);
+    pub fn with_external_printer(mut self, printer: ExternalPrinter) -> Self {
+        self.external_printer = Some(printer);
         self
     }
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -3,17 +3,13 @@ use std::path::PathBuf;
 use itertools::Itertools;
 use nu_ansi_term::{Color, Style};
 
+#[cfg(feature = "external_printer")]
+use crate::ExternalPrinterChannel;
 use crate::{enums::ReedlineRawEvent, CursorConfig};
 #[cfg(feature = "bashisms")]
 use crate::{
     history::SearchFilter,
     menu_functions::{parse_selection_char, ParseAction},
-};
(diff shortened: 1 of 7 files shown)


IanManske:
On error, is returning the raw `String` or `Vec<u8>` value ok, or should these be wrapped in an `PrintError<T>(pub T)` type?

There are still some issues with printing. For example, the screen is cleared if the prompt is not at the bottom of the screen and more than one line is printed through the external printer. I think there was a previous, similar issue to this that was fixed, but I guess the fix was not carried over to the external printer?

stormasm:
@IanManske would it be possible or do you think it makes sense to have the ability to send the data from the channel to a file or the terminal --- and or have some type of configuration option that says send data to a file or to the terminal...

The reason I am asking this is because I have implemented via a logger the ability to send data to a file using the EnvLogger crate...

https://github.com/nushell/reedline/compare/main...stormasm:reedline:simplelogcomplete

This is the general API call for the logger -> https://docs.rs/simplelog/0.12.1/simplelog/struct.WriteLogger.html

If we had the ability to send data to a file via the ExternalPrinter then I would not need to use this logging facility...

And or maybe we should have both types of functionality in Reedline...

Both a logging mechanism and the ExternalPrinter ---- but that seems probably redundant ?

Thanks for thinking about this as a potential solution 😄

IanManske:
@storm, I think it's fine to have both a logging mechanism and the external printer, since the two are meant for two different purposes. The current job of the external printer is to synchronize multiple writers to the terminal and to manipulate newlines to account for the terminal being in raw mode. So, if we are just logging to a file, then I don't think there's much overlap with the external printer. I.e., the external printer with its mpsc channel is probably overkill/not applicable, since logs can just be directly written to a file without needing to go through a channel.

For more context, in nushell's case, I intend that redirecting background job output to a file will be handled through normal redirection (`o>`, `e>`, etc.). So, reedline won't be involved at all in this case, since there's no output bound for the terminal.

Anyways, thanks for bringing this up!

> One of the reasons I like sending data to a file instead of the terminal is because it allows me to review what is going on later on as well as have a permanent state of what happened.
> I also find sending data to the terminal as "messy" and confusing with so much stuff flying across the screen it gets kind of hard to understand what is going on.

Yep, makes sense :smile:

stormasm:
@IanManske cool !  thanks for the thorough write up...   I learned a lot yesterday in our meeting discussing the external printer and again with this conversation....   I will probably have more questions --- this is important work and will certainly add more depth to Reedline and Nushell in general so its great you are taking on this project and diving into the details of how everything works ---- and we are the beneficiaries of the education on the subject 👍

stormasm (commented):
Thats cool that 
crossbeam-channel` was merged into `std::sync::mpsc`
and that you made the update...

src/engine.rs
```diff
@@ -679,22 +675,24 @@ impl Reedline {
...
                 }
             }
 
             let mut latest_resize = None;
-            loop {
+
+            // There could be multiple events queued up!
+            // pasting text, resizes, blocking this thread (e.g. during debugging)
+            // We should be able to handle all of them as quickly as possible without causing unnecessary output steps.
+            while event::poll(Duration::from_millis(POLL_WAIT))? {
```
sholderbach, line 695: There was some subtlety around when we `crossterm::event::poll` and `crossterm::event::read` to avoid idling in a hotter than necessary loop https://github.com/nushell/reedline/pull/651

So I am not sure if this `do ... while` to `while` transformation is correct.

src/external_printer.rs
```diff
@@ -1,72 +1,70 @@
...
-    pub fn sender(&self) -> Sender<T> {
+
+    /// Returns a new [`SyncSender`] which can be used in other threads to queue messages to print
+    pub fn sender(&self) -> SyncSender<Vec<u8>> {
         self.sender.clone()
     }
-    /// Receiver to get messages if any
-    pub fn receiver(&self) -> &Receiver<T> {
+
+    pub(crate) fn receiver(&self) -> &Receiver<Vec<u8>> {
```
sholderbach, line 39: Sounds like a good call to me to lock this up

src/external_printer.rs
```diff
@@ -1,72 +1,70 @@
...
-pub struct ExternalPrinter<T>
-where
-    T: Display,
-{
-    sender: Sender<T>,
-    receiver: Receiver<T>,
+#[derive(Debug)]
+pub struct ExternalPrinter {
+    sender: SyncSender<Vec<u8>>,
+    receiver: Receiver<Vec<u8>>,
```
sholderbach, line 13: Also agree on removing the generic here.

sholderbach (commented):
Haven't done adverserial manual testing around the drawing but I like the direction this is going.

src/engine.rs
IanManske: Unfortunately, the current loop structure will block on the first `read` and later external messages will not get printed until the user types which is confusing. I wonder if there's a better solution though...

sholderbach: Damn you asynchronous concurrency....

Not blocking by polling certainly makes reading that channel possible.
Could we shove both kinds of events into one channel and block on that? (that would be the no async just threads everywhere path)

With how we dealt with `crossterm` and stdout so far I am certainly not screaming: `cargo add tokio`

IanManske: Using the channel + threads should work, but I think means there will have to be 3 threads in total:
- One to forward messages from `event::read`
- One to forward messages from the external printer
- The main thread, to pull messages off the channel

It is possible to remove the external printer thread if we wrap the external printer sender in a new type. This way, we can allow the external thread to send messages directly to the merged channel (`Reciever<ReedlineMessage>`), instead of having to move data from the external channel `Reciever<Vec<u8>>` to the merged channel. E.g.,:
```
enum ReedlineMessage {
    Event(crossterm::Event),
    ExternalMessage(Vec<u8>),
}

struct ExternalPrinterSender(SyncSender<ReedlineMessage>)

impl ExternalPrinterSender {
    pub fn send(&self, data: Vec<u8>) -> SendError<Vec<u8> {
        self.0.send(ReedlineMessage::ExternalMessage(data)).map_err(...);
    }
}
```
The potential problem with this is that `nu-system`, or wherever job controls ends up, would need to take `reedline` as a dependency.

sholderbach: Crossterm mentions a restriction on the threads their blocking functions can be run from and that it is incompatible with their async oriented `EventStream`

https://docs.rs/crossterm/latest/crossterm/event/index.html

Maybe we need to bite the `Future` bullet at some point? (providing an executor would suck for library users, but we may want to unlock things like async prompt or completion updates.)

IanManske: > Crossterm mentions a restriction on the threads their blocking functions can be run from

This should be fine, one thread can both poll and read.

> Maybe we need to bite the Future bullet at some point?

In the future (ha), it might be necessary. So far, it looks like we can probably work around it?

stormasm:
>> The potential problem with this is that nu-system, or wherever job controls ends up, would need to take reedline as a dependency.

That would not work well --- ideally we don't want to have any reedline dependencies in nushell except for where they are currently located which is in nu-cli which is the only place we want it...

and nu-lsp...  
And having it in nu-lsp is for the 

```rust
use reedline::Completer
```

Reason being we want to be able to allow our development community to be able to swap out nu-cli and replace it with their own cli / line editor...

Or run nushell inside a [*nana*](https://github.com/nushell/nana) type application where no cli or reedline is needed...

src/engine.rs
IanManske: Two other possible solutions:
1. Use a higher poll wait time like it was suggested in #757. We could use the higher wait time for the first poll only and then switch back to the current wait time after the first read.
2. Have only two threads:
   1. the main thread: has the old `poll` + `read` loop and prints the prompt, etc.
   2. the external printer thread: takes messages off the channel receiver and prints these messages.
   
   Then, we synchronize the two threads using a `Mutex` or something similar to ensure that only one of them prints to the terminal at a time.

sholderbach: Not sure how 1) would play out both in terms of external latency and syscall churn. (would need to see it)

2) sounds more involved. Question would be if the `crossterm` cursor or terminal size polling operations may collide with streaming output from another thread. But yeah with the appropriate locking this may be workable.

IanManske:
Ok, the polling problem is a little more involved than I thought. I'm not sure if either of the thread solutions will work (well):

- For the three thread solution: this will interleave `crossterm` events and external messages into one channel. The potential problem with this is that an external message source can send a bunch of messages and swamp the channel (e.g., background job suddenly outputs a lot of text). This could prevent/delay us from processing user input which is not the best, since the user input could contain ctrlc's or shell commands to kill (background) jobs. 
- For the two thread solution: the external printer thread would need to reprint the prompt, since the main thread could be blocked on `event::read`. But this would require the external printer thread to have access to the `&dyn Prompt` and `&mut self` `EngineState`. Not sure if there's a way to do this well (if at all).

So, this leaves a few options:

1. We use higher poll wait time, and users with the `external-printer` feature would see a little decrease in battery life.
2. We bite the async bullet.
3. Maybe there's another possible solution (e.g., we somehow make one of the thread solutions workable).

PR #751, opened 2024-02-12
Fix logic error in style_range leading to overflow on subtraction

Tastaturtaste:
Fixes https://github.com/nushell/reedline/issues/750#issue-2127609999


fdncred:
Tested with @Tastaturtaste's help and it looks great to me.

PR #758, opened 2024-02-26
Fix rendering prompt after ExecuteHostCommand

### LINKED ISSUE #755: After `ExecuteHostCommand` is handled, new prompt is rendered
**Platform** Windows 11 WSL2 Ubuntu 20.04,
**Terminal software** wsltty 3.5.1

## Steps to reproduce

```nu
❯ def do_nothing [] {}
❯ let x = {name: do_nothing, modifier: control, keycode: char_g, mode: [vi_normal, vi_insert], event: {send: executehostcommand, cmd: 'do_nothing'}}
❯ $env.config.keybindings = ($env.config.keybindings | append $x)
❯ # type C-g here, then new prompt is rendered
❯
```

kit494way:
Fix #755

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -210,7 +210,7 @@ impl Reedline {
             history_excluded_item: None,
             history_cursor_on_excluded: false,
             input_mode: InputMode::Regular,
-            execute_host_command: false,
+            executing_host_command: false,
             painter,
             transient_prompt: None,
             edit_mode,
@@ -671,8 +671,8 @@ impl Reedline {
     /// Helper implementing the logic for [`Reedline::read_line()`] to be wrapped
     /// in a `raw_mode` context.
     fn read_line_helper(&mut self, prompt: &dyn Prompt) -> Result<Signal> {
-        if self.execute_host_command {
-            self.execute_host_command = false;
+        if self.executing_host_command {
+            self.executing_host_command = false;
         } else {
             self.painter.initialize_prompt_position()?;
             self.hide_hints = false;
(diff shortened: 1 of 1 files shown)


fdncred:
Do you think this PR will fix this nushell issue? https://github.com/nushell/nushell/issues/11981

kit494way:
Yes, I think this PR will fix the issue.

src/engine.rs
```diff
@@ -109,6 +109,9 @@ pub struct Reedline {
     history_cursor_on_excluded: bool,
     input_mode: InputMode,
 
+    // ExecuteHostCommand
+    execute_host_command: bool,
+
```
sholderbach, line 114: Small nit: This naming and comment goes a bit lost among all the other configuration focussed members of `Reedline`

It describes a temporary state of `Reedline` so maybe `executing_host_command`/`after_host_command` is more explicit. I don't have a strong opinion where it should be placed in the overall struct, but there we could also emphasize its role.

```suggestion
    // Yielded to the host program after a `ReedlineEvent::ExecuteHostCommand`, thus redraw in-place
    executing_host_command: bool,

```

sholderbach (approved):
Logic looks good.
From a maintainability perspective a small comment but we should test this with Nushell

kit494way:
Thanks, I accepted the suggestion and renamed.

fdncred:
yay! i was hoping we'd get this in before the next release. i'm wondering what impacts it will have on people's keybindings, meaning, if people are relying on a newline, it could be breaking something. let's find out!

Issue #759, opened 2024-02-27
Implicit copy to system clipboard doesn't obey law of least surprise and poses potential security hazard


Tastaturtaste:
Certain operations, such as `alt + d`, cut content and put the content in the cut buffer. With the `system_clipboard` feature enabled, the cut buffer is replaced by the system clipboard. This can lead to surprising behaviour changes for the user and pose a security hazard in case a user cuts sensitive information into the system clipboard without knowing about it.

For more context see https://github.com/nushell/nushell/issues/11907.

## Steps to reproduce
1.  Type a top secret password, eg. '1234'
2. Move the cursor back to the start of the line
3. Press alt+d, invoking `CutWordRight`
4. Switch to another program entirely, such as a text editor, and paste

## Possible Solution
I am currently looking into this and think it would be best if reedline *always* has a local clipboard (which it calls the cut buffer) and additionally with the `system_clipboard` feature enabled a system clipboard. Most commands would always use the local clipboard, only some explicitly named commands interact with the system clipboard. I would also like to avoid changing which clipboard gets used by a command depending on the feature flag used to compile. Commands should either use the cut buffer, which is always available, or the system clipboard. 

Commands that copy to the system clipboard would be `CopySelection` (default Ctrl+Shift+C) and `CutSelection` (default Ctrl+Shift+X). If permitted with regard to backward compatibility of reedline, I would also like to rename them to `CopySelectionToSystem` and `CutSelectionToSystem` to be more explicit about what they do. Changing the way the clipboard works is a semantic breaking change either way.

In addition to `PasteCutBufferBefore` (default Ctrl+Shift+V and emacs Ctrl+Y), I would introduce a new Command `PasteSystemClipboard`, which does what `PasteCutBufferBefore` currently does with the system clipboard enabled. I would then change `PasteCutBufferBefore` to always paste from the local clipboard, matching its name. The default key binding for `Ctrl+Shift+V` would then be changed to invoke the new `PasteSystemClipboard`.

## Open Questions

1. Should the `EditCommand` enum variants concerning the `system_clipboard` feature (`CutSelectionToSystem`, ...) be conditionally included only? This would statically ensure that those commands can only come in when there is actually a system clipboard available. In this case, the enum should also be marked `non_exhaustiv`, otherwise this would break exhaustive matches for crates when somewhere in their dependency graph the `system_clipboard` feature gets newly enabled in the future.
2. Are there commands where it _does_ make sense to change the clipboard used depending on the feature flag?

[...]

PR #761, opened 2024-02-27
Use the OS clipboard only for explicit cut/copy/paste operations

Tastaturtaste:
This PR addresses https://github.com/nushell/reedline/issues/759. 

This PR reverts most commands to using the local cut buffer for cut/copy/paste operations except for `CutSelection` and `CopySelection`. Those two commands now have a parameter to specify which clipboard to use. Current keybindings for those commands (`CutSelection`: `Ctrl+Shift+X`, `CopySelection`: `Ctrl+Shift+V`) still use the system clipboard. Currently, no keybindings are defaulted for these commands and the cut buffer. 

We can't statically compile everything related to the `system_clipboard` feature out, as that would include the parametrized commands mentioned above. Activating the feature would not be purely additive anymore. Therefore those parameters have to stay even when the feature is disabled and the commands have to do _something_ when the feature is not enabled. Currently, there is no facility in place for `EditCommand`s to fail in a recoverable manner or indicate failure to the user. A few possibilities how to handle this situation that come to mind:

1. Panic if the OS clipboard would have to be used. The user of the crate would be responsible to ensure commands using the OS clipboard are only used when the feature is enabled.
2. Silently do nothing when the OS clipboard would have to be used
3. Silently use the local clipboard, even when the command explicitly requests the OS clipboard

Of these options, I currently implement panicking. I think it is the least surprising thing to do for users of `reedline`. Requesting to use the OS clipboard while not having `system_clipboard` enabled is surely a logic bug. But I am open to discussion if someone sees it another way.

The changes can be used with [this nushell branch](https://github.com/Tastaturtaste/nushell), where I use this reedline branch to test things out.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -1,6 +1,8 @@
 use super::{edit_stack::EditStack, Clipboard, ClipboardMode, LineBuffer};
+#[cfg(feature = "system_clipboard")]
+use crate::core_editor::get_system_clipboard;
 use crate::enums::{EditType, UndoBehavior};
-use crate::{core_editor::get_local_clipboard, core_editor::get_system_clipboard, EditCommand};
+use crate::{core_editor::get_local_clipboard, EditCommand};
 use std::ops::DerefMut;
 
 /// Stateful editor executing changes to the underlying [`LineBuffer`]
@@ -534,6 +536,9 @@ impl Editor {
     }
 
     fn cut_selection_to_system(&mut self) {
+        #[cfg(not(feature = "system_clipboard"))]
+        panic!("The OS clipboard is not enabled. Enable its use with the reedline/system_clipboard feature!");
+        #[cfg(feature = "system_clipboard")]
         if let Some((start, end)) = self.get_selection() {
             let cut_slice = &self.line_buffer.get_buffer()[start..end];
             self.system_clipboard.set(cut_slice, ClipboardMode::Normal);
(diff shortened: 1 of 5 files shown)


src/core_editor/editor.rs
```diff
@@ -526,7 +535,19 @@ impl Editor {
         self.line_buffer.move_to_end();
     }
 
-    fn cut_selection(&mut self) {
+    fn cut_selection_to_system(&mut self) {
+        #[cfg(not(feature = "system_clipboard"))]
+        panic!("The OS clipboard is not enabled. Enable its use with the reedline/system_clipboard feature!");
+        #[cfg(feature = "system_clipboard")]
+        if let Some((start, end)) = self.get_selection() {
```
sholderbach, line 542: From the "don't panic" Nushell side this is a dangerous panic footgun.

`EditCommand::Paste { system_clipboard: true }` provides a possibly more aesthetic API over separate variants. But if you were to compile without `system_clipboard` and pass the same event this could panic the shell. 
https://github.com/nushell/reedline/blob/998a2f79092a4c9058e1394a1495063ded8f42ad/src/core_editor/editor.rs#L132-L137
The match compiles either way. This would require extra care while parsing complex user configs.

Either we completely restrict it at compile time via separate variants or turn it into something handlable.

Tastaturtaste: I assume turning it into something handlable would involve returning a Result up the call stack so nushell can do something like inform the user? In general I think that might be a good idea, it also opens up for other failable commands should the need arise. Though I don't know if that is the right approach for handling this specific issue. What is your opinion on it? How should it be handled?

sholderbach (approved):
Thanks for championing this! Let's get this moving.

Tastaturtaste:
Ok, that was fast :D

If you want to give it a try: https://github.com/Tastaturtaste/nushell/tree/bug/11907-system-clipboard-not-for-everything

fdncred:
Thanks again @Tastaturtaste.

PR #770, opened 2024-03-12
Fix (properly) the logic around prompt re-use & Host Command handling

### LINKED ISSUE #755: After `ExecuteHostCommand` is handled, new prompt is rendered

### LINKED ISSUE #771: 0.30.0 (Nu 0.91.0) does not respect cursor position after host cmd exec in a keybinding
### Describe the bug

In zsh I'm used to have a number of keybindings that run commands so I don't need to type them as they are so common, especially some `git` commands.

In nushell I previously (in 0.89.0) was able to do something like this by simply executing a command and the prompt would continue where it was, but with 0.91.0 the output of my command is cleared and prompt is reset to where it was before. ☹

### How to reproduce

1. Setup a keybinding with:
   ```nu
    {
      name: git_status
      mode: [emacs vi_insert vi_normal]
      modifier: Alt keycode: Char_g
      event: {
        send: ExecuteHostCommand
        cmd: "git status"
      }
    }
   ```
2. Open a nushell, go anywhere, use the `Alt-g` keybind to run `git status`

### Expected behavior

I expected nu to continue working as it was before, allowing me to run commands by keybindings and see its output

### Screenshots

https://github.com/nushell/nushell/assets/9730330/ce3b9ffe-7fe4-446f-a309-8ef1f588fdcd

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.91.0                                                            |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.76.0 (07dca489a 2024-02-04) (built from a source tarball) |
| cargo_version      | cargo 1.76.0                                                      |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash, which, zip                                |
| installed_plugins  |                                                                   |

### Additional context

_No response_

bew:
First PR for nushell/reedline \o/

Revert/Rewrites #758

Fixes #755 (correctly this time 😬)

<details>
<summary>Where did the newline actually come from in that issue?</summary>

The _phantom_ newline that OP reported was due to this block in painter initialization:
(👉 It would detect some text before the cursor and always make a new prompt on the next line..)
https://github.com/nushell/reedline/blob/0698712701418a7212ebf75d5724d72eccc4b43f/src/painting/painter.rs#L107-L120

</details>

Fixes #771 (my issue)
Now I have the same experience than on my zsh config :smiley: 

Supports everything I mentioned in #771
Supports replacing the prompt (even from multi-line old prompt)

## Demo

https://github.com/nushell/reedline/assets/9730330/ab52e593-cc1d-4762-8df8-9e5774538531

### ~~MISSING~~ (DONE)
- [x] Unit tests for the painter initialization logic
- [x] Refactor painter initialization logic (to remove that unwrap 👀)

---
What do you think of this implementation?

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -672,7 +672,8 @@ impl Reedline {
     /// Helper implementing the logic for [`Reedline::read_line()`] to be wrapped
     /// in a `raw_mode` context.
     fn read_line_helper(&mut self, prompt: &dyn Prompt) -> Result<Signal> {
-        self.painter.initialize_prompt_position(self.suspended_state.as_ref())?;
+        self.painter
+            .initialize_prompt_position(self.suspended_state.as_ref())?;
         if self.suspended_state.is_some() {
             // Last editor was suspended to run a ExecuteHostCommand event,
             // we are resuming operation now.
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -27,7 +27,7 @@ use {
             FileBackedHistory, History, HistoryCursor, HistoryItem, HistoryItemId,
             HistoryNavigationQuery, HistorySessionId, SearchDirection, SearchQuery,
         },
-        painting::{Painter, PromptLines},
+        painting::{Painter, PainterSuspendedState, PromptLines},
         prompt::{PromptEditMode, PromptHistorySearchStatus},
         result::{ReedlineError, ReedlineErrorVariants},
         terminal_extensions::{bracketed_paste::BracketedPasteGuard, kitty::KittyProtocolGuard},
@@ -109,8 +109,9 @@ pub struct Reedline {
     history_cursor_on_excluded: bool,
     input_mode: InputMode,
 
-    // Yielded to the host program after a `ReedlineEvent::ExecuteHostCommand`, thus redraw in-place
-    executing_host_command: bool,
+    // State of the painter after a `ReedlineEvent::ExecuteHostCommand` was requested, used after
+    // execution to decide if we can re-use the previous prompt or paint a new one.
+    suspended_state: Option<PainterSuspendedState>,
 
     // Validator
     validator: Option<Box<dyn Validator>>,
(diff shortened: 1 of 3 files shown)


bew:
There is a bad case with of multi-line + cursor on one of the first lines + executing command `echo; foo-that-outputs`..
Like with a 3 lines prompt, cursor on first line, that command would go down 1 line but would overwrite the 3rd line.

A solution would be to use the command `commandline set-cursor --end; echo; foo-that-outputs`.
But I'm thinking we could also expose a new flag like `commandline set-cursor --below-prompt` to place the cursor below all the lines of the prompt (and it's nice & easy to read 🙃)

bew:
Hello!
I finally went back to finish this PR, it should be ready for review now 🙏

fdncred:
I've played a little bit with this without nushell and it appears to be outputting less ansi escape codes, which is good. 

Separately, on my own and unrelated to this PR, I'm trying to figure out how/why reedline prints so many CRLFs in nushell and also how/why the prompt is redrawn with every character typed.

fdncred:
> there's a bad case with a multi-line prompt + cursor on one of the first lines + executing command echo; foo-that-outputs..

@bew does this 'bad case' still exist?

bew:
> > there's a bad case with a multi-line prompt + cursor on one of the first lines + executing command echo; foo-that-outputs..
> 
> @bew does this 'bad case' still exist?

The 'bad case' I mentioned is only a limitation with the example, due to using 'echo' in the keybind.
There is nothing I can do in this PR afaik.

To property solve this limitation with using `echo` to move cursor below the prompt I suggested the creation of a `commandline set-cursor` option that would do that, but I think this should be done in a separate PR.

PR #699, opened 2024-01-04
Move left when exiting insert mode

andreistan26:
As mentioned in #694 using the `vi` editor mode does not match the behavior of the `vim/vi` CLI editors.

Issue #772, opened 2024-03-16
Prefix history search broken in vi-mode


nixypanda:
**Platform** macOS
**Terminal software** kitty

Describe the problem you are observing.
prefix history search is not working in vi-mode (normal) when hitting `k` key.

## Steps to reproduce
1. Script used to run reedline.
```rust
use reedline::{
    default_vi_insert_keybindings, default_vi_normal_keybindings, DefaultPrompt, Reedline, Signal,
    Vi,
};

fn main() {
    let mut line_editor = Reedline::create().with_edit_mode(Box::new(Vi::new(
        default_vi_insert_keybindings(),
        default_vi_normal_keybindings(),
    )));

    let prompt = DefaultPrompt::default();

    loop {
        let sig = line_editor.read_line(&prompt);
        match sig {
            Ok(Signal::Success(buffer)) => {
                println!("We processed: {}", buffer);
            }
            Ok(Signal::CtrlD) | Ok(Signal::CtrlC) => {
                println!("\nAborted!");
                break;
            }
            x => {
                println!("Event: {:?}", x);
            }
        }
    }
}
```
2. Write a bunch of commands. eg. (ls rock, ls paper, sl sic, ls rofl)
3. `ls<esc><k>`

## Expected behaviour
Cycles through the commands that start with `ls`

## Actual Behaviour (bug)
Cycles through all the history commands

## Screenshots/Screencaptures

<none>

## Additional Info
Reverting https://github.com/nushell/reedline/pull/699 fixes the issue.

## Thoughts
As someone who briefly contributed to this project a while back and as someone who uses vi-mode (all the time) I really never in my life questioned why, when I hit escape nushell prompt does not go back one place.

I mean why do we want to be exactly like vim? I really think we should focus on doing vi-mode that makes sense for the terminal.
Like if we do this going back stuff and we fix the history navigation. It looks like the prefix will lose it last char in case we want to do history search based on where the cursor position is, and than this case will also require some special handling (afaik).

So, I am really questioning why even we need something like that.

fdncred:
@andreistan26 your PR seems to have broken this functionality. Any thoughts on how to make both your and this prefix searching work together?

andreistan26:
Looking into it

andreistan26:
Thinking about it for a bit, @nixypanda is right, for a CLI row editor going back when switching to normal mode is annoying. I would revert that PR. If you think it should stay as it is I can think about a solution.

fdncred:
[...]

PR #773, opened 2024-03-18
Revert "Move left when exiting insert mode"

fdncred:
Reverts nushell/reedline#699

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -136,17 +136,8 @@ impl EditMode for Vi {
                 }
                 (_, KeyModifiers::NONE, KeyCode::Esc) => {
                     self.cache.clear();
-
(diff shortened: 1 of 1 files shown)


132ikl:
Personally sad to see this reverted, I have muscle memory from using vim/similar constantly. Reverting until it can be handled better/less buggy seems to be the right move, but wanted to voice my support for this behavior generally

fdncred:
I believe we reverted this because the history navigation by prefix failed if the cursor was in the wrong position when in vi mode.

132ikl:
Yes, and I think it's fine to revert the PR since the implementation seems to have been buggy. There was some discussion in #772 about this behavior being "annoying" or not appropriate for a readline, which I disagree with. I figured it made more sense to continue the discussion here rather than in #772 though.

PR #777, opened 2024-03-26
Fix case-consistency searching sqlite history

sholderbach:
After merging downstream will close https://github.com/nushell/nushell/issues/10131

For the `FileBackedHistory` those operations have always been case
sensitive, do the same for `SqliteBackedHistory`. The insensitivity of
`like` in sqlite causes https://github.com/nushell/nushell/issues/10131

Replace this with exact matching via sqlite's [`instr()`](https://www.sqlite.org/lang_corefunc.html#instr), which should also fix a need for escaping of `%` (or `*` with `glob`)

### DIFF
--- a/src/history/base.rs
+++ b/src/history/base.rs
@@ -350,6 +350,24 @@ mod test {
     }
 
     #[test]
+    fn search_prefix_is_case_sensitive() -> Result<()> {
+        // Basic prefix search should preserve case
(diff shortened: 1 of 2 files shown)


fdncred:
Cool! Seems good to be more consistent. Never heard of `instr()` before this.

PR #778, opened 2024-03-27
Make filtering by `cwd` case sensitive

sholderbach:
Taken from #777

## Feedback wanted on:

Making `cwd` matching case sensitive. On windows paths are not necessarily case sensitive when asking for them. would this be a problem. otherwise would incresase correctness on case sensitive file systems.

- **Make cwd prefix case sensitive as well**
- **Use `instr` for cwd as well**


kronberger-droid:
Picking this up since it went quiet.
Pushed a test onto the branch that pins the new behaviour, which fails against the old `LIKE` query since an unescaped `_` in the prefix matched any character.
That wildcard leak is the stronger reason to land this, the case change is a bonus.

PR #779, opened 2024-03-28
add bashism `!term` to prefix search for last command beginning with `term`

fdncred:
This was created by request for @ayax79 but I missed this functionality in nushell too. Hopefully this works the way we all want it to.

The idea is to do `!term` and it do a search backwards through the history looking for the most recent command that begins with `term` and then supply that on the command line.

Currently, I don't believe it understands history isolation, so, _I think_ this is searching the entire history. I could be wrong though.

I threw this together fast because my laptop is about to be reimaged by my company and I can't mess with it anymore right now.


fdncred:
Let's give this a try!

Issue #774, opened 2024-03-19
`columnar_menu` panics on non-ASCII characters in directory


sarubo:
**Platform** Windows 10, Arch Linux
**Terminal software** Windows Terminal, Alacritty 

`nu > version | transpose key value | to md --pretty`

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.91.1                                   |
| branch             | main                                     |
| commit_hash        | 6e37ad0275cb23c5a8a933ba6a46a45d70f74910 |
| build_os           | windows-x86_64                           |
| build_target       | x86_64-pc-windows-gnu                    |
| rust_version       | rustc 1.74.1 (a28077b28 2023-12-04)      |
| rust_channel       | 1.74.1-x86_64-pc-windows-gnu             |
| cargo_version      | cargo 1.74.1 (ecb9851af 2023-10-18)      |
| build_time         | 2024-03-19 18:57:34 +09:00               |
| build_rust_channel | debug                                    |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash, which            |
| installed_plugins  |                                          |

With `RUST_BACKTRACE=1` it will look like this:

```
sarub@LAPTOP-WIN-SRB:~/projects/rust/nushell/target/debug
nu > cd `foo/`Error:   × Main thread panicked.
  ├─▶ at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1\library\core\src\str\mod.rs:660:13
  ╰─▶ byte index 6 is not a char boundary; it is inside 'あ' (bytes 4..7) of `foo\あ\`
         0: 0x7ff62799c550 - rust_begin_unwind
                      at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library\std\src/panicking.rs:597
         1: 0x7ff6279f77f7 - core::panicking::panic_fmt::hd44cd0ce6cc3cd33
                      at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library\core\src/panicking.rs:72
         2: 0x7ff627a006e7 - core::str::slice_error_fail_rt::h9eebe5afeb115af2
         3: 0x7ff627a001c3 - core::str::slice_error_fail::h99c8592d623ec1b6
                      at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library\core\src\str/mod.rs:87
         4: 0x7ff625db595b - core::str::<impl str>::split_at::he586fec1974bbb3e
[...]

PR #730, opened 2024-01-28
Typed text style

maxomatic458:
closes #724

Adds ``selected_match_style`` and ``match_style`` to ``MenuTextStyle``.

IdeMenu
grafik

ColumnarMenu
grafik

PR #782, opened 2024-04-14
Fix panic in menus

### LINKED ISSUE #769: Index out of bounds
**Platform:** Linux, NixOS
**Terminal software:** Ghostty

I get this when I do `ffmpeg -i RGBCube.gif -crf`, assuming ffmpeg is installed.

```
thread 'main' panicked at /build/rustc-1.75.0-src/library/core/src/str/mod.rs:666:13:
byte index 4 is out of bounds of `dds`
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::str::slice_error_fail_rt
   3: core::str::slice_error_fail
   4: reedline::menu::columnar_menu::ColumnarMenu::create_string
   5: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   6: <alloc::string::String as core::iter::traits::collect::FromIterator<alloc::string::String>>::from_iter
   7: <reedline::menu::columnar_menu::ColumnarMenu as reedline::menu::Menu>::menu_string
   8: reedline::painting::painter::Painter::print_menu
   9: reedline::painting::painter::Painter::repaint_buffer
  10: reedline::engine::Reedline::repaint
  11: reedline::engine::Reedline::read_line
  12: nu_cli::repl::evaluate_repl
  13: nu::run::run_repl
  14: nu::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Here is a recording, but it doesn't show the trace as it appears for a split second: https://asciinema.org/a/oO8bzFkNpKFd3aFVq7Bfatr2d

The .cast file has it, though.

Here is my completer config:

```nu
$env.config.completions = {
  algorithm:      prefix
  case_sensitive: false
  partial:        true
  quick:          true
  external: {
    enable:      true
    max_results: 100
    completer:   {|tokens: list<string>|
      let expanded = scope aliases | where name == $tokens.0 | get --ignore-errors expansion.0

      mut expanded_tokens = if $expanded != null and $tokens.0 != "cd" {
        $expanded | split row " " | append ($tokens | skip 1)
      } else {
        $tokens
      }

      $expanded_tokens.0 = ($expanded_tokens.0 | str trim --left --char "^")

      fish --command $"complete '--do-complete=($expanded_tokens | str join ' ')'"
      | $"value(char tab)description(char newline)" + $in
      | from tsv --flexible --no-infer
    }
  }
}
```

Everything is from nixpkgs unstable.

maxomatic458:
closes #769

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -299,13 +299,21 @@ impl ColumnarMenu {
         use_ansi_coloring: bool,
     ) -> String {
         if use_ansi_coloring {
-            let match_len = self.working_details.shortest_base_string.len();
+            let match_len = self
+                .working_details
+                .shortest_base_string
+                .trim_end()
+                .chars()
+                .count();
 
             // Split string so the match text can be styled
-            let (match_str, remaining_str) = if match_len <= suggestion.value.len() {
-                suggestion.value.split_at(match_len)
+            let (match_str, remaining_str) = if match_len <= suggestion.value.chars().count() {
+                (
+                    suggestion.value.chars().take(match_len).collect::<String>(),
+                    suggestion.value.chars().skip(match_len).collect::<String>(),
+                )
             } else {
-                ("", suggestion.value.as_str())
+                ("".to_string(), suggestion.value.to_string())
             };
 
             let suggestion_style_prefix = suggestion
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -302,7 +302,11 @@ impl ColumnarMenu {
             let match_len = self.working_details.shortest_base_string.len();
 
             // Split string so the match text can be styled
-            let (match_str, remaining_str) = suggestion.value.split_at(match_len);
+            let (match_str, remaining_str) = if match_len <= suggestion.value.len() {
+                suggestion.value.split_at(match_len)
+            } else {
+                ("", suggestion.value.as_str())
+            };
 
             let suggestion_style_prefix = suggestion
                 .style
--- a/examples/ide_completions.rs
+++ b/examples/ide_completions.rs
@@ -33,7 +33,7 @@ fn main() -> io::Result<()> {
     // Max width of the completion box, including the border
     let max_completion_width: u16 = 50;
     // Max height of the completion box, including the border
-    let max_completion_height: u16 = u16::MAX;
+    let max_completion_height = u16::MAX;
     // Padding inside of the completion box (on the left and right side)
     let padding: u16 = 0;
     // Whether to draw the default border around the completion box
--- a/examples/ide_completions.rs
+++ b/examples/ide_completions.rs
@@ -33,7 +33,7 @@ fn main() -> io::Result<()> {
     // Max width of the completion box, including the border
     let max_completion_width: u16 = 50;
     // Max height of the completion box, including the border
-    let max_completion_height = u16::MAX;
+    let max_completion_height: u16 = u16::MAX;
     // Padding inside of the completion box (on the left and right side)
     let padding: u16 = 0;
     // Whether to draw the default border around the completion box
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -822,7 +822,7 @@ impl Menu for IdeMenu {
     }
 
     fn menu_required_lines(&self, _terminal_columns: u16) -> u16 {
-        self.get_rows().min(self.default_details.max_completion_height)
+        self.get_rows()
     }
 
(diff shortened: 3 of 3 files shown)


maxomatic458:
now also fixes #774

src/menu/columnar_menu.rs
```diff
@@ -299,10 +299,22 @@ impl ColumnarMenu {
         use_ansi_coloring: bool,
     ) -> String {
         if use_ansi_coloring {
-            let match_len = self.working_details.shortest_base_string.len();
+            let match_len = self
+                .working_details
+                .shortest_base_string
+                .trim_end()
```
sholderbach, line 305: Why do we need to trim the string and thus change the length here?

src/menu/columnar_menu.rs
```diff
@@ -299,10 +299,22 @@ impl ColumnarMenu {
...
+                .chars()
+                .count();
 
             // Split string so the match text can be styled
-            let (match_str, remaining_str) = suggestion.value.split_at(match_len);
+            let (match_str, remaining_str) = if match_len <= suggestion.value.chars().count() {
```
sholderbach, line 310: As the `self.working_details.shortest_base_string` succeeded to be generated in `self.update_values()` (the range coming from `Completer::complete_with_base_ranges` did not violate char boundaries which would panic the `SliceIndex`) 
https://github.com/nushell/reedline/blob/46f410bfdce0b64c3d94fdc0271886f30d64b151/src/menu/columnar_menu.rs#L497-L504
 having to do char based indexing, indicates to me that `shortest_base_string` seems to be not a strict prefix of `suggestion.value`. 
 Don't know if that is really the intended semantics to then split here based on length.

src/menu/columnar_menu.rs
maxomatic458: ``shortest_base_string`` can include trailing whitespace from the buffer, because the completer will ignore whitespace, when making a completion. "test " -> will complete the same as "test" but will have a different span. so we need to trim it, so 4 chars match and not 5.

src/menu/columnar_menu.rs
maxomatic458: the ranges from ``Completer::complete_with_base_ranges`` (and pretty much all buffer related stuff) are byte based not character based, to support unicode we need to do char indexing here.

> shortest_base_string seems to be not a strict prefix of suggestion.value

thats actually somewhat true if you use other completion methods (instead of the default starts_with)
if thats the case ``if match_len <= suggestion.value.chars().count()`` will most likely fail and not style the matching string at all.

sholderbach: If the byte indxes are wrong it should already fail on the index in L502, if I remember the slice index semantics of `str` correctly.

So this whole section doing the match highlighting either needs to be aware of which matching mode was used, get the correct coordinates to highlight or perform some alignment to check which characters in the query match to the picked Suggestion.

sholderbach: > Im pretty sure the indexing in L502 wont fail because the range is directly coming from the completer (doesnt matter which completion mode was used)

The `Completer` implementer can deliver invalid (i.e. not byte-encoding aware or out-of bounds) ranges in its span (or completely override `complete_with_base_ranges`).
But this would panic L502 as it tries to index the buffer `&str`. [Indexing outside the encoded chars/bounds of a `str` is a panic](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=308c9a8363567e8f769f979a338de2c2)
So the culprit for what you a trying to fix is not a malformed span from the `Completer` (which should be fixed at that source first and then hardened here in a way that still informs implementers).

> if you use other completion methods (instead of the default starts_with)
> if thats the case `if match_len <= suggestion.value.chars().count()` will most likely fail and not style the matching string at all.

We shouldn't accept a highlighting system that only stops to highlight mismatching prompted text and suggestion under the circumstance that something else is fishy.

The highlighting should mark the text that is equal/equivalent between the user query and the suggestions.
And the assumption that the user query is a prefix of all suggestions is not true. 

(I don't have the time to do a full archaeology on #730 or before if the `shortest_base_string` is the right proxy to reflect the matching text itself)

maxomatic458: I think ill close this PR for now and make another one where i will implement your suggestions. (so add proper support for different completion methods and figure out a safer way to get the base ranges)

RGBCube:
closed as completed, is this fixed? if so, we should close the tracking issues

fdncred:
@RGBCube I closed it based on this comment https://github.com/nushell/reedline/pull/782#discussion_r1605486354

PR #784, opened 2024-04-22
Remove debug print

sholderbach:


fdncred:
oops, i thought i caught all those. guess i missed one, thanks.

PR #796, opened 2024-05-31
Add PWD to the `Reedline` state

YizhePKU:
This PR adds the current working directory as part of the `Reedline` state, instead of using `std::env::current_dir()` to retrieve it from the environment. Applications should call `Reedline::with_cwd()` to update the current working directory when it changes.

This is necessary to support https://github.com/nushell/nushell/pull/12922, which causes `std::env::current_dir()` to deviate from Nushell's `$env.PWD`.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -142,9 +142,9 @@ pub struct Reedline {
     // Use ansi coloring or not
     use_ansi_coloring: bool,
 
-    // Current working directory as defined by the application, which might not
-    // be the actual working directory of the process
-    cwd: String,
+    // Current working directory as defined by the application. If set, it will
+    // override the actual working directory of the process.
+    cwd: Option<String>,
 
     // Engine Menus
     menus: Vec<ReedlineMenu>,
@@ -228,7 +228,7 @@ impl Reedline {
             hide_hints: false,
             validator,
             use_ansi_coloring: true,
-            cwd: "".to_string(),
+            cwd: None,
             menus: Vec::new(),
             buffer_editor: None,
             cursor_shapes: None,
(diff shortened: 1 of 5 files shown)


devyn:
It would probably be best that it's either required or defaults to the actual current dir. Reedline is intended to be useful for more than just nushell

src/engine.rs
```diff
@@ -224,6 +228,7 @@ impl Reedline {
             hide_hints: false,
             validator,
             use_ansi_coloring: true,
+            cwd: "".to_string(),
```
devyn, line 231: Just to make it clear - I don't think an empty string should be the default. I think making it an `Option` and then grabbing process working directory if it's `None` is probably the best way to go

devyn (changes requested):

YizhePKU:
> Just to make it clear - I don't think an empty string should be the default. I think making it an Option and then grabbing process working directory if it's None is probably the best way to go.

So you're saying that if `cwd` is set to None, we fallback to the old behavior (calling `std::env::current_dir()`), so that we don't break existing code that doesn't require special PWD logic?

That sounds good. I'll work on it.

devyn:
Yeah, exactly. Thanks!

fdncred:
@YizhePKU Do you have any more time to work on this in regards to @devyn's comments?

YizhePKU:
I think this will do? I'll update the other PR if this gets merged.

fdncred:
Thanks @YizhePKU! We appreciate you coming back around to this PR. @devyn, when you get a chance can you please look at this. I'd like to get it landed if possible.

devyn (approved):

devyn:
Looks correct to me. Thanks!

jakeswenson:
Does anyone know how this relates to nushell/nushell/issues/5799? Not sure if this directly resolves it or if there is more work to be done?

PR #798, opened 2024-06-09
Add match_indices field to Suggestion

ysthakur:
## Description

This PR adds a `match_indices: Vec<usize>` field to the `Suggestion` struct representing which indices of the Suggestion's `value` field were matched.

## Motivation

Currently, in Nushell, if you have fuzzy matching on, the matches are still highlighted with the assumption that you're using prefix matching:
image

This PR doesn't fix that, but adding `match_indices` to `Suggestion` does make it so that completers using fuzzy matching can include the indices of the matches in their returned suggestions. Here's the `fuzzy_completions` example included in this PR, which uses a columnar menu (ignore how bad the fuzzy completion algorithm is, I didn't bother using skim or Nucleo):

[asciicast](https://asciinema.org/a/KeHlkgwdFSTrTHUIOycJ2TVzb?t=5)

And modified to use an IDE menu with fake descriptions:

[asciicast](https://asciinema.org/a/bOLaIbyLM8C6hasSqPGcE50Kh)

## Tests

I only added a test for the `style_suggestion` helper that I added for highlighting the matched portions of a suggestion. Other than that, I just manually tried the menu out a bit. If there are any edge cases anyone can think of to test (either through unit tests or manually), I'd appreciate that.

### DIFF
--- a/examples/fuzzy_completions.rs
+++ b/examples/fuzzy_completions.rs
@@ -1,0 +1,135 @@
+// Modifies the completions example to demonstrate highlighting of fuzzy completions
+// cargo run --example fuzzy_completions
+
+use reedline::{
+    default_emacs_keybindings, ColumnarMenu, Completer, DefaultPrompt, EditCommand, Emacs, KeyCode,
+    KeyModifiers, Keybindings, MenuBuilder, Reedline, ReedlineEvent, ReedlineMenu, Signal, Span,
+    Suggestion,
+};
+use std::io;
+
+struct HomegrownFuzzyCompleter(Vec<String>);
+
+impl Completer for HomegrownFuzzyCompleter {
+    fn complete(&mut self, line: &str, pos: usize) -> Vec<reedline::Suggestion> {
+        // Grandma's fuzzy matching recipe. She swears it's better than that crates.io-bought stuff
+        self.0
+            .iter()
+            .filter_map(|command_str| {
+                let command = command_str.chars().collect::<Vec<_>>();
+                let mut start = 0;
+                let mut match_indices = Vec::new();
+                for l in line.chars() {
+                    if start == command.len() {
+                        break;
+                    }
+                    let mut i = start;
+                    while i < command.len() && l != command[i] {
+                        i += 1;
+                    }
+                    if i < command.len() {
+                        match_indices.push(i);
+                        start = i + 1;
+                    }
+                }
+                if match_indices.is_empty() || match_indices.len() * 2 < pos {
+                    None
+                } else {
+                    Some(Suggestion {
+                        value: command_str.to_string(),
+                        description: None,
+                        style: None,
+                        extra: None,
+                        span: Span::new(pos - line.len(), pos),
+                        append_whitespace: false,
+                        match_indices: Some(match_indices),
+                    })
+                }
+            })
+            .collect()
+    }
+}
+
+fn add_menu_keybindings(keybindings: &mut Keybindings) {
+    keybindings.add_binding(
+        KeyModifiers::NONE,
+        KeyCode::Tab,
+        ReedlineEvent::UntilFound(vec![
+            ReedlineEvent::Menu("completion_menu".to_string()),
+            ReedlineEvent::MenuNext,
+        ]),
+    );
+    keybindings.add_binding(
+        KeyModifiers::ALT,
+        KeyCode::Enter,
+        ReedlineEvent::Edit(vec![EditCommand::InsertNewline]),
+    );
+}
+
+fn main() -> io::Result<()> {
+    // Number of columns
+    let columns: u16 = 4;
+    // Column width
+    let col_width: Option<usize> = None;
+    // Column padding
+    let col_padding: usize = 2;
+
+    let commands = vec![
+        "test".into(),
+        "clear".into(),
+        "exit".into(),
+        "history 1".into(),
+        "history 2".into(),
+        "logout".into(),
+        "login".into(),
+        "hello world".into(),
+        "hello world reedline".into(),
+        "hello world something".into(),
+        "hello world another".into(),
+        "hello world 1".into(),
+        "hello world 2".into(),
+        "hello another very large option for hello word that will force one column".into(),
+        "this is the reedline crate".into(),
+        "abaaabas".into(),
+        "abaaacas".into(),
+        "ababac".into(),
+        "abacaxyc".into(),
+        "abadarabc".into(),
+    ];
+
+    let completer = Box::new(HomegrownFuzzyCompleter(commands));
+
+    // Use the interactive menu to select options from the completer
+    let columnar_menu = ColumnarMenu::default()
+        .with_name("completion_menu")
+        .with_columns(columns)
+        .with_column_width(col_width)
+        .with_column_padding(col_padding);
+
+    let completion_menu = Box::new(columnar_menu);
+
+    let mut keybindings = default_emacs_keybindings();
+    add_menu_keybindings(&mut keybindings);
+
+    let edit_mode = Box::new(Emacs::new(keybindings));
+
+    let mut line_editor = Reedline::create()
+        .with_completer(completer)
+        .with_menu(ReedlineMenu::EngineCompleter(completion_menu))
+        .with_edit_mode(edit_mode);
+
+    let prompt = DefaultPrompt::default();
+
+    loop {
+        let sig = line_editor.read_line(&prompt)?;
+        match sig {
+            Signal::Success(buffer) => {
+                println!("We processed: {buffer}");
+            }
+            Signal::CtrlD | Signal::CtrlC => {
+                println!("\nAborted!");
+                break Ok(());
+            }
+        }
+    }
+}
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -90,4 +90,7 @@ pub struct Suggestion {
     /// Whether to append a space after selecting this suggestion.
     /// This helps to avoid that a completer repeats the complete suggestion.
     pub append_whitespace: bool,
+    /// Indices of the characters in the suggestion that matched the typed text.
+    /// Useful if using fuzzy matching.
+    pub match_indices: Option<Vec<usize>>,
 }
--- a/src/completion/default.rs
+++ b/src/completion/default.rs
@@ -110,6 +110,7 @@ impl Completer for DefaultCompleter {
                                         extra: None,
                                         span,
                                         append_whitespace: false,
+                                        match_indices: None,
                                     }
                                 })
                                 .filter(|t| t.value.len() > (t.span.end - t.span.start))
@@ -384,6 +385,7 @@ mod tests {
                     extra: None,
                     span: Span { start: 0, end: 3 },
                     append_whitespace: false,
+                    match_indices: None,
                 },
                 Suggestion {
                     value: "ｎｕｍｂｅｒ".into(),
@@ -392,6 +394,7 @@ mod tests {
                     extra: None,
                     span: Span { start: 0, end: 3 },
                     append_whitespace: false,
+                    match_indices: None,
                 },
                 Suggestion {
                     value: "ｎｕｓｈｅｌｌ".into(),
@@ -400,6 +403,7 @@ mod tests {
                     extra: None,
                     span: Span { start: 0, end: 3 },
                     append_whitespace: false,
+                    match_indices: None,
                 },
             ]
         );
@@ -428,6 +432,7 @@ mod tests {
                     extra: None,
                     span: Span { start: 8, end: 9 },
                     append_whitespace: false,
+                    match_indices: None,
                 },
                 Suggestion {
                     value: "this is the reedline crate".into(),
@@ -436,6 +441,7 @@ mod tests {
                     extra: None,
                     span: Span { start: 8, end: 9 },
                     append_whitespace: false,
+                    match_indices: None,
                 },
                 Suggestion {
                     value: "this is the reedline crate".into(),
@@ -444,6 +450,7 @@ mod tests {
                     extra: None,
                     span: Span { start: 0, end: 9 },
                     append_whitespace: false,
+                    match_indices: None,
                 },
             ]
         );
--- a/src/completion/history.rs
+++ b/src/completion/history.rs
@@ -65,6 +65,7 @@ impl<'menu> HistoryCompleter<'menu> {
             extra: None,
             span,
             append_whitespace: false,
+            match_indices: None,
         }
     }
 }
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -1,12 +1,13 @@
 use super::{Menu, MenuBuilder, MenuEvent, MenuSettings};
 use crate::{
     core_editor::Editor,
-    menu_functions::{can_partially_complete, completer_input, replace_in_buffer},
+    menu_functions::{
+        can_partially_complete, completer_input, replace_in_buffer, style_suggestion,
+    },
     painting::Painter,
     Completer, Suggestion,
 };
 use nu_ansi_term::ansi::RESET;
-use unicode_width::UnicodeWidthStr;
 
 /// Default values used as reference for the menu. These values are set during
 /// the initial declaration of the menu and are always kept as reference for the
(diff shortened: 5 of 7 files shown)


sholderbach:
I think @maxomatic458 would be the ideal reviewer here, based on recent work in the completion logic and the discussion we had around the pitfalls with this highlighting.

maxomatic458:
looks good so far.

I think instead of defaulting to this
https://github.com/ysthakur/reedline/blob/44e18841807ef914d302f57fc9cb9b2c151c099b/src/menu/ide_menu.rs#L516-L523

it might be better to just ignore the styling (so basically fully leave it up to the completer).

Then we could get rid of this trait function here (since we have the ``match_indices`` now):
https://github.com/ysthakur/reedline/blob/44e18841807ef914d302f57fc9cb9b2c151c099b/src/completion/base.rs#L39-L51

If you are ok with that i can make a PR for that.

ysthakur:
@maxomatic458 Interesting idea, I was going to say highlighting shouldn't be the completer's responsibility, but if it simplifies things, that'd be great. It's definitely worth a PR. @sholderbach Thoughts?

qfel:
> I was going to say highlighting shouldn't be the completer's responsibility, but if it simplifies things, that'd be great.

How would that interact with other highlighting functionality in reedline? Eg. it takes care of highlighting the selected item, if the completer dealt with highlighting individual characters how would that mix?

> it might be better to just ignore the styling (so basically fully leave it up to the completer).

Do you mean decoupling value from how it's displaying, eg. adding something like `display_value` to `Suggestion`, while `value` is still what gets put on the command line on selection?


Slightly tangential, but there's plenty of places that initialize `Suggestion` structs and have to deal with setting lots of fields to None (though mostly in nushell proper it seems) - maybe it's time to implement `Default`?

maxomatic458:
> Do you mean decoupling value from how it's displaying, eg. adding something like display_value to Suggestion, while value is still what gets put on the command line on selection?

I think this might actually be better solution. So instead of adding more stuff to the Suggestion we just have one ``display_value``

that can be modified by a seperate styling layer (that can be attached to a completer).
This would also make styling way more customizable.

But this might require a good amount of breaking changes, so maybe thats something for a future PR?

ysthakur:
> How would that interact with other highlighting functionality in reedline? Eg. it takes care of highlighting the selected item, if the completer dealt with highlighting individual characters how would that mix?

Ah, you're right @qfel, highlighting in the completer wouldn't work.

> I think this might actually be better solution. So instead of adding more stuff to the Suggestion we just have one `display_value`

@maxomatic458 As @qfel pointed out, we'd need two display values, one in case the suggestion is selected and one in case it isn't, right? That feels like too much to me.

> that can be modified by a seperate styling layer (that can be attached to a completer).
This would also make styling way more customizable.

@maxomatic458 This sounds interesting, could you explain more? (having a hard time figuring out what a separate styling layer would look like)

maxomatic458:
@ysthakur 

maybe this could be implemented as a trait
something like

```rs
pub trait SuggestionStyler {
    fn style_selected(&self, suggestion: Suggestion, ...) -> String {
        ...
    }
    
    fn style_unselected(&self, suggestion: Suggestion, ...) -> String {
        ...
    }
}
```
then we could pass a SuggestionStyler struct down to the menu and inside the `create_string` function of the menu we can style the suggestions

the functions of the `SuggestionStyler` trait could have paremeters like typed text, cursor pos

so we would move the styling out of the menu and out of the completer and we can implement it using this trait

ysthakur:
@maxomatic458 Sorry, still not getting it. Do you mean that the `style` field would be removed from `Suggestion`, the `match_indices` field from this PR would still be added, and then a `SuggestionStyler` would perform styling? And this styler could output by mutating a `display_value` field on the suggestion?

maxomatic458:
@ysthakur 
that approach would not need any more fields on the Suggestion itself
all styling would be handled inside this trait

that would potentially allow for more complex styling, without having to make any changes to the menu

so something like this here
```rs

enum MatchMode {
    Fuzzy,
    Prefix,
}

struct UnderlineStyler {
    color: ...,
    mode: MatchMode
}

impl SuggestionStyler for UnderlineStyler {
    fn style_selected(&mut self, suggestion: Suggestion, cursor_pos: Option<(u16, u16)>, text_typed_by_user: String, terminal_dims: Option<(u16, u16)>, ...) -> Suggestion {
        // Do the styling here
        ...
        return suggestion
    }
    
    fn style_unselected(...) -> Suggestion {
         ...
    }
}

```

then we could add this ``Styler`` as a field to the menus

and then call the trait functions in menu when we are generating the strings for the suggestions


Thats just an idea how it could be implemented if we want to make the suggestion-styling that extensible.
But i also understand that there probably isnt really a need to make it that extensible atm. But that would just avoid having to add more fields to the Suggestions struct in the future.

qfel:
IIUC that would either mean we keep `match_indices` so that `Styler` can use the information to style part of the suggestion - in which case that seems fully orthogonal to this PR and better discuss outside (so we can make the smaller incremental change here).

Or it would mean inside the styler we have to recompute the matching indices, which sounds a bit unpleasant. Unless we allow to attach arbitrary data with suggestions - then completers could generate suggestions with completer-specific hints, and provide stylers that consume those hints to producer stylized output. Though that feels like a fairly big change?

maxomatic458:
> Or it would mean inside the styler we have to recompute the matching indices, which sounds a bit unpleasant

That was my idea (if you are concerned about the recomputing, i believe in both cases we would need to recompute the indices whenever the user types something). We would attach a Styler to a completer, and that would handle the styling of the suggestions. (so the Suggestions struct would not have to be modified again, when we want to introduce a new way of styling suggestions)

ysthakur:
> @ysthakur
that approach would not need any more fields on the Suggestion itself
all styling would be handled inside this trait

@maxomatic458 Oh, I see now. I like the idea of avoiding adding fields to `Suggestion` to control styling, but as qfel said, recomputing the matching indices sounds unpleasant. Also, both the completer and the styler will need to be aware of how matching works.

That said, I agree with you that performance might not be a huge problem. Only a few suggestions are displayed on the screen at any point, even if there are hundreds of suggestions in total. If we only recompute the match indices for each suggestion when it shows up on the screen, the user may not see a noticeable delay.

I'm not an expert here, though. Anyone have any idea what impact this would have on performance?

> so the Suggestions struct would not have to be modified again, when we want to introduce a new way of styling suggestions.

@maxomatic458 Did you have any possible future ways of styling suggestions in mind? Do you mean we might want to highlight suggestions in three different colors (rather than the current two)?

maxomatic458:
@ysthakur

> Also, both the completer and the styler will need to be aware of how matching works.

Ah now i see, yea in that case we would need to do the same calculation twice.


> Did you have any possible future ways of styling suggestions in mind? Do you mean we might want to highlight suggestions in three different colors (rather than the current two)?

yes multicolored stuff would be possible.
Nushell currently has different colored file suggestions per filetype (i think you need to enable it in the config)
grafik

So this if we provide file system information to the styler stuff like this would be possible.
We could even mutate the text of the suggestion (without changing the completion value if we add a `display_value` field)
so maybe we could add info like file size, or edit date or stuff like that to the suggestion.

like:

    File 1 (1 GB) yesterday
    File 2 (2 GB) 10 days ago

qfel:
> I agree with you that performance might not be a huge problem.

That aside, potentially having to run the same matching code in two different places in the Completer is an unpleasant API. Not a huge issue I suppose, but makes it a little harder to get things going (and might have some weird implications for non-deterministic/pure matching, maybe if somebody decided to do some server-side completions).

> terminal_dims: Option<(u16, u16)>

I noticed one of the examples above had this - what would it be used for?

Hm let me expand a bit on the styler idea that I think could work well (if we really want to go there):

1. We split `Suggestion` into two separate types, so to not mix presentation with completion (plus for some fields it would make no sense to change them in Styler, so why have it push them trough):

```rust
pub struct Suggestion<T> {
    /// String replacement that will be introduced to the the buffer
    pub value: String,
    /// Replacement span in the buffer
    pub span: Span,
    /// Whether to append a space after selecting this suggestion.
    /// This helps to avoid that a completer repeats the complete suggestion.
    pub append_whitespace: bool,
    /// Completer-associated context, passed back to Styler.
    pub context: T,
}

pub struct DisplaySuggestion {
    /// The value to display, often the same as coming from Suggestion but will have ANSI escapes.
    pub value: String,
    /// Optional description for the replacement, can contain ANSI escapes.
    pub description: Option<String>,
    /// Optional vector of strings in the suggestion. These can be used to
    /// represent examples coming from a suggestion. Can contain ANSI escapes.
    pub extra: Option<Vec<String>>,
}
```
2.  We introduce a `Styler` trait, to convert `Suggestion` to `DisplaySuggestion`:

```rust
// Probably needs more thought about the naming.
pub trait Styler<T> {
  fn display(&self, suggestion: Suggestion<T>, is_selected: bool, some_type_of_user_config: _, ...) -> DisplaySuggestion;
  // Also batch_display with some default implementation. May be useful for some completers to parallelize.
}
```
3. Update Completer to allow for generic Suggestions and Styler.

```rust
pub trait Completer {
  type Context;
  fn complete(&mut self, line: &str, pos: usize) -> Vec<Suggestion<Self::Context>>;

  // We *probably* want to instead have it be a separate field on ReedlineMenu::WithCompleter,
  // so that any compatible Styler can be used. But that's maybe something to think on later.
  //
  // This could return one of predefined simple static stylers, or something custom for the completer.
  fn styler<'a>(&'a self) -> &'a dyn Styler<Self::Context>;
}
```

4. Provide some basic Stylers, eg. one for highlighting the entire item, one highlighting a prefix (for completers with eg. u32 Context). New Stylers could use the base ones for generating regular `DisplaySuggestion` and then updating just a few fields as needed. 

Regarding the `Context` associated type: it don't necessarily think it's important for performance (though that really depends on a particular completer), but aside from making think a bit (I think) cleaner, it also can serve as a way to declare which Stylers are compatible with which Completers - which in turn would let use decouple the two more, rather than have Completer define a single associated Styler.

maxomatic458:
> > terminal_dims: Option<(u16, u16)>

> I noticed one of the examples above had this - what would it be used for?

Yea not really sure what that could be used for either, just came into my mind when thinking of things that might be useful to style a suggestion.

I think it would also be useful to store the string the completer used to suggest that suggestion in the Suggestion struct.

ysthakur:
> I think it would also be useful to store the string the completer used to suggest that suggestion in the Suggestion struct.

@maxomatic458 Would this be used by the `Styler` trait to have more information for styling?

> Not a huge issue I suppose, but makes it a little harder to get things going (and might have some weird implications for non-deterministic/pure matching, maybe if somebody decided to do some server-side completions).

That's a good point, @qfel, running matching twice wouldn't be good for non-deterministic/time-consuming suggestions.

That said, `match_indices` won't be enough if we ever decide we want more than 2 styles in a single suggestion, e.g. green for the first half, red for the second half, underlines for the matched characters. I personally cannot think of why this would happen, but it is possible we need advanced styling in the future.

@maxomatic458's `display_value` idea solves that, but I was thinking that instead of a single field, we could instead have two fields, `display_value_unselected` and `display_value_selected`. Instead of providing a style, each completer could style its suggestions completely on its own, using helper functions provided by Reedline. The advanced styling that would have gone into the `Styler` trait would be done by each completer instead, though most would just Reedline's helpers for that. No `match_indices` field would be needed in each suggestion itself. Thoughts?

maxomatic458:
> Would this be used by the Styler trait to have more information for styling?

@ysthakur that could be also be used outside of styling. Currently we have the helper function "complete_with_base_ranges", that basically does the same thing already, but instead of collecting the actual strings, we just get the ranges in the buffer (i think).

So an extra field that holds the "completion base value" could be used to store that instead.
Currently thats required to align the IDE menu suggestions with the text already typed in the terminal.
https://github.com/nushell/reedline/blob/ad56a23d05cb1701c397bc2c3010c8048bb920f5/src/menu/ide_menu.rs#L80-L87

webbdays:
updating create_string method is enough i think.
split suggestion.value into three parts and add highlight to matched part and return a formatted string.

like this


let suggestion_style_prefix = suggestion.clone()
        .style
        .unwrap_or(Style::new().bold().fg(Color::Cyan))
        .prefix();
        let extras = suggestion.extra.as_ref().unwrap();
        let match_str = extras.get(0).unwrap();
        let splits = Self::split_string_by_pattern(&suggestion.value, &match_str).unwrap();

        format!(
            "{}{}{}{}{}{}{}{}{}{:>empty$}{}{}",
            suggestion_style_prefix,
            splits.0,
            Style::default().underline().prefix(),
            match_str,
            RESET,
            suggestion_style_prefix,
            splits.2,
            RESET,
            Color::Yellow.normal().prefix(),
            "",
            RESET,
            "\r\n",
            empty = empty_space,
            )

for example.

ysthakur:
> Hi, does this also solves the issue of mismatch in highlight. the create_string is assuming the match string start at the front.

@webbdays Yes, we're trying to handle cases like `obr` matching `foobar` with a fuzzy matching algorithm. Doesn't solve anything yet, though, this is still being worked on :(

> updating create_string method is enough i think.
split suggestion.value into three parts and add highlight to matched part and return a formatted string.

I'm not sure I understand. I don't think this would handle a case like the one I mentioned above, where there are multiple substrings that need to be highlighted (above, `"ob"` and `"r"`).

webbdays:
I am talking of one match
lib in printinglib

webbdays:
It highlights only pri as length of lib is 3

ysthakur:
@maxomatic458 @qfel I think going forward with a `match_indices` field is the simplest option. And @qfel's [suggestion](https://github.com/nushell/reedline/pull/798#issuecomment-2237877464) of using a `display_value` field should allow users to customize the text shown and its styling.

The reason I left this PR up for so long is because I wasn't sure we could make the `display_value` idea work. The route I took now is having both an optional `display_value` and a `match_indices` field, with `display_value` possibly being an already-styled string with ANSI escapes included. Then we apply the match style to just the matched characters within the `display_value` string, taking care to respect the existing styling.

This is complicated by the fact that crates for properly parsing styles from strings with ANSI escapes would increase the size of the Reedline crate quite a bit. But there's a very simple crate, [`parse-ansi`](https://crates.io/crates/parse-ansi/0.1.6), that just contains a regex for matching ANSI escape codes. That's good enough for our purposes, even if it's not the best. The crate is 3 KiB.

With that, you can create a suggestion with `display_value: "<RED-FG>abc<GREEN-FG>def<RESET>"` (pretend those are ANSI escapes) and `match_indices: [0, 5]` (highlight "a" and "f"). If the match style is meant to give the text a blue foreground, the styled result will be `<RED-FG><BLUE-FG>a<RESET><RED-FG>bc<GREEN-FG>de<BLUE-FG>f<RESET><RESET>`. Note the unnecessary ANSI resets. These are because the code doesn't understand the actual styles.

When it sees a chunk of the string like this:
```
<ORIG STYLE><unmatched text><matched text><unmatched text>
```
it will turn that into
```
<ORIG STYLE><unmatched text><MATCH STYLE><matched text><RESET><ORIG STYLE><unmatched text>
```
(where `ORIG STYLE` is a bunch of ANSI escapes included in the original styled string and `MATCH STYLE` is the ANSI escapes for applying the match style)

Because we aren't properly extracting `Style`s from the string, this also means that if `display_value` has styles that compose, like `<RED-FG>foo<UNDERLINE>bar` (`bar` should both be red and underlined), the match styling will break that. Instead, users will have to do `<RED-FG>foo<RED-FG><UNDERLINE>bar`. But I do think that's fine. Those advanced users who need multiple colors for a single suggestion can probably handle adding resets themselves.

Since the `display_value` thing is orthogonal to this feature, I haven't implemented that here.

divanshu-go:
@ysthakur 
will user need to input match_indices: [0, 5]
or will it receive it in the result match_indices: [0, 5]

divanshu-go:
@ysthakur Please check https://github.com/ysthakur/reedline/pull/2
as it fixes the cargo clippy warning
```bash
warning: current MSRV (Minimum Supported Rust Version) is `1.63.0` but this item is stable since `1.80.0`
   |
11 | static ANSI_REGEX: LazyLock<Regex> = LazyLock::new(|| {
   |                                      ^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv
   = note: `#[warn(clippy::incompatible_msrv)]` on by default
   ```

divanshu-go:
@ysthakur The regex dependency remains in Cargo.lock because it's a transitive dependency of rstest (our test framework)

ysthakur:
This PR now does the ANSI sequence parsing itself, rather than using regex. It should be ready to merge now (built and ran Nushell with these changes and things seem fine).

kubouch (approved):
I don't have anything against merging it

132ikl (approved):

PR #799, opened 2024-06-10
Make menus process events before updating working details

ysthakur:
This PR makes the columnar and description menus update their lists of suggestions after processing events (in `update_working_details`), rather than before. Currently, these menus can be rendered with outdated working details.

The completions example in Reedline itself demonstrates this. Here's how it looks currently (I modified the default completer to add a dummy description just for this recording):
[asciicast](https://asciinema.org/a/pd84reD58dR487a7oQ8VVV6iR?t=2)

Here's how the columnar menu looks after this PR:
[asciicast](https://asciinema.org/a/rwrWi6lHqTG2pUP5cGcQ8Bx6I)

Description menu (before):
[asciicast](https://asciinema.org/a/qeh6wysIoVRwdCjh7iI8JKE6S?t=6)

Description menu (after):
[asciicast](https://asciinema.org/a/RLwikRIqaXv1moJ7igbF2nbA6?t=4)

I assume there was a reason for originally updating working details before processing events, given that both the columnar and description menu have comments explicitly saying that that's what they're doing. However, reading the code, I didn't see any reason to believe that processing the events first would mess with the working details, I didn't run into any problems myself, and the new IDE menu made by maxomatic458 (which processes events before updating working details) seems to work just fine. If anyone knows of any possible problems with processing events first, do let me know, but otherwise, I think this PR is safe to merge.

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -515,8 +515,42 @@ impl Menu for ColumnarMenu {
         painter: &Painter,
     ) {
         if let Some(event) = self.event.take() {
-            // The working value for the menu are updated first before executing any of the
-            // menu events
+            match event {
+                MenuEvent::Activate(updated) => {
+                    self.active = true;
+                    self.reset_position();
+
+                    self.input = if self.settings.only_buffer_difference {
+                        Some(editor.get_buffer().to_string())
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks

PR #800, opened 2024-06-19
Feature: vi visual mode

adamschmalhofer:
This extends the vi mode with a visual mode. Can be tested in the basic example: v to enter in normal mode. Regular movements (e.g. b,w) work and d or C to delete and reenter normal or invert mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,13 +222,13 @@ Reedline has now all the basic features to become the primary line editor for [n
 - Undo support.
 - Clipboard integration
 - Line completeness validation for seamless entry of multiline command sequences.
+- Visual selection
 
 ### Areas for future improvements
 
 - [ ] Support for Unicode beyond simple left-to-right scripts
 - [ ] Easier keybinding configuration
 - [ ] Support for more advanced vi commands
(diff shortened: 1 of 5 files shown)


adamschmalhofer:
> we require `cargo fmt --all` to be run as well.


fixed.

> also, how can i test this. I'm not familiar with the vi keybindings.

$ cargo run --example demo -- --vi

type some text to test on

Press "ESC" to switch into the vi normal mode and then press "v" to enter visual mode.

Now using the same keys as in vi normal mode moves the cursor while also selecting the text. These are "h" and "l" (lower case L) for moving a char left and right, or "b" and "w" for moving a word left and right, "B" and "W" for the big words, as well as "0" (zero) and "$" for the start and end of the line.

Once you have selected your text "d" deletes the selected text or "c" to delete the selected text and enter new text.

fdncred:
ah, ok. thanks for taking the time to explain it to me. that helps. we really need to make the colors configurable but that's not for this pr. It just looks a bit janky (not your fault).
        
image

fdncred:
Thanks

PR #802, opened 2024-06-23
Refractor: group colors

adaschma:
In https://github.com/nushell/reedline/pull/800#issuecomment-2182943637 @fdncred said:

> we really need to make the colors configurable but that's not for this pr. It just looks a bit janky (not your fault).
> 
> image

This pull request groups the statically defined colors into a single record. This is the first step if we would want to support configurable colors at a later time. In addition it changes the fg color of selected text to black so there is a contrast to the bg color. This improves the readability and it looks nicer.

image

The colors of the menu and the highlighter have not been moved to the theme record. It is at least theoretically possible to have different menus and highlighters with different colors and I conservatively didn't want to ruin that.

### DIFF
--- a/src/prompt/base.rs
+++ b/src/prompt/base.rs
@@ -1,5 +1,4 @@
 use {
-    crossterm::style::Color,
     serde::{Deserialize, Serialize},
     std::{
         borrow::Cow,
--- a/src/prompt/base.rs
+++ b/src/prompt/base.rs
@@ -8,12 +8,6 @@ use {
     strum_macros::EnumIter,
 };
 
-/// The default color for the prompt, indicator, and right prompt
-pub static DEFAULT_PROMPT_COLOR: Color = Color::Green;
-pub static DEFAULT_PROMPT_MULTILINE_COLOR: nu_ansi_term::Color = nu_ansi_term::Color::LightBlue;
-pub static DEFAULT_INDICATOR_COLOR: Color = Color::Cyan;
-pub static DEFAULT_PROMPT_RIGHT_COLOR: Color = Color::AnsiValue(5);
-
 /// The current success/failure of the history search
 pub enum PromptHistorySearchStatus {
     /// Success for the search
@@ -97,22 +91,6 @@ pub trait Prompt: Send {
         &self,
         history_search: PromptHistorySearch,
     ) -> Cow<str>;
-    /// Get the default prompt color
-    fn get_prompt_color(&self) -> Color {
-        DEFAULT_PROMPT_COLOR
-    }
-    /// Get the default multiline prompt color
-    fn get_prompt_multiline_color(&self) -> nu_ansi_term::Color {
-        DEFAULT_PROMPT_MULTILINE_COLOR
-    }
-    /// Get the default indicator color
-    fn get_indicator_color(&self) -> Color {
-        DEFAULT_INDICATOR_COLOR
-    }
-    /// Get the default right prompt color
-    fn get_prompt_right_color(&self) -> Color {
-        DEFAULT_PROMPT_RIGHT_COLOR
-    }
 
     /// Whether to render right prompt on the last line
     fn right_prompt_on_last_line(&self) -> bool {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -180,7 +180,6 @@ impl Drop for Reedline {
     }
 }
 
-
 pub struct Theme {
     pub visual_selection: Style,
     pub use_ansi_coloring: bool,
@@ -189,13 +188,14 @@ pub struct Theme {
     pub prompt_multiline: nu_ansi_term::Color,
     pub indicator: Color,
     pub prompt_right: Color,
-
 }
 
 impl Default for Theme {
     fn default() -> Self {
         Self {
-            visual_selection: Style::new().fg(nu_ansi_term::Color::Black).on(nu_ansi_term::Color::LightGray),
+            visual_selection: Style::new()
+                .fg(nu_ansi_term::Color::Black)
+                .on(nu_ansi_term::Color::LightGray),
             use_ansi_coloring: true,
             prompt: Color::Green,
             prompt_multiline: nu_ansi_term::Color::LightBlue,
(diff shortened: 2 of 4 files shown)


src/prompt/base.rs
```diff
@@ -97,22 +90,6 @@ pub trait Prompt: Send {
         &self,
         history_search: PromptHistorySearch,
     ) -> Cow<str>;
-    /// Get the default prompt color
-    fn get_prompt_color(&self) -> Color {
```
fdncred, line 101: I'm just wondering if there's a way to do this without creating a breaking change. Removing these 4 functions will break everyone who's using reedline and making custom prompts, right?

fdncred:
I think this is cool and an interesting way to abstract the colors into a theme, but I worry about existing users as mentioned above. Also the "theme" name should probably be less generic.

src/prompt/base.rs
adamschmalhofer: I don't see a way to keep these functions.

These 4 functions have default implementations and are the current way to change the default colors of reedline.  Looking at [the reverse dependancies of reedline](https://crates.io/crates/reedline/reverse_dependencies) I picked a few to check if they have there own implementations ans so this would be a breaking change:

That don't: 

- [nu-cli](https://github.com/nushell/nushell/blob/main/crates/nu-cli/src/prompt.rs)
- [reedline-repl-rs](https://github.com/arturh85/reedline-repl-rs/blob/main/src/prompt.rs)
- [aichat](https://github.com/sigoden/aichat/blob/main/src/repl/prompt.rs)
- [Cecile](https://github.com/your-moon/Cecile/blob/master/src/repl/mod.rs)
- [dust_lang](https://git.jeffa.io/jeff/dust/src/branch/main/src/main.rs)

That do:

- [diatom-cli](https://github.com/diatom-lang/diatom/blob/master/diatom-cli/src/cli/prompt.rs)

Since changing the colors are already used by existing users, at the very least I would need to expose Theme record for diatom. If having a breaking change isn't a show stopper, I will add that.

adamschmalhofer:
I thought Theme was in line with Prompt. What would you prefer? ReedlineTheme or ColorTheme? I don't have strong opinions about the name and am open to change it.

fdncred:
Maybe ReedlineTheme. I'm just thinking that everyone calls their theme things "Theme". When we're in nushell, I don't want to be confused about which theme we're talking about.

src/prompt/base.rs
fdncred: of course, I'm most concerned about nushell with this change.

adamschmalhofer: Glade this isn't a blocker. Added builder methods so diathom-cli (and other users) can still change the colors.

adamschmalhofer:
> Looks like there are some conflicts to resolve.

Fixed.

fdncred:
Thanks. Are you up for adding a PR to nushell with config settings in order to give this a thorough test?

fdncred:
ping @adamschmalhofer

fdncred:
ok, i don't want to land this and then be stuck in a place where someone else has to implement something in nushell without fully understanding these changes. so we can wait on landing this until you're ready to land its counterpart in nushell.

adamschmalhofer:
> ok, i don't want to land this and then be stuck in a place where someone else has to implement something in nushell without fully understanding these changes. so we can wait on landing this until you're ready to land its counterpart in nushell. Thanks for pitching in!

Well, if changes would be need to be done in nushell for this pull request, I would do them. Most projects that I know of (including myself) prefer small incremental changes rather than a large "dump". This reduces the problem of merge conflicts and helps the code reviews stay focused. 

Does a pull request need a visible user facing feature in nushell to be considered? Or is the problem that it isn't properly tested, because nushell doesn't change the colors? In that case would a pull request for [diatom-cli](https://github.com/diatom-lang/diatom/blob/master/diatom-cli/src/cli/prompt.rs) be sufficient? I'm just a bit confused and would like to know for this and future pull requests.

fdncred:
Thanks for responding. I'm hyped about this PR. Here is my thinking/reasoning:
1. we want this change in nushell because it's helpful to people. i.e. we don't like how it's working today in nushell.
2. it's good to have code and tests in reedline but for a feature like this, we really need a test case that we can dogfood for a while. that's why i'm pushing for a pr in nushell as a counterpart to this pr. 
3. we 100% agree that we like bite-sized-prs too.
4. does it "need" to be a visible facing feature in nushell just to be considered? hmm, thinking through this, maybe/probably? there are some prs that we allow and just land, but with things like ui features, we like to dogfood for a while. i assume your code works and i've tested it in reedline but i don't feel like that's a real use case. we feel that it needs to be in someone's daily driver so we can be sure it delivers on its promise. there are always so many edge cases to consider.

does all that help understand my position?

adamschmalhofer:
Thanks for taking the time to elaborate. I think to part where we disagree is where we cut the line on what counts as ui features. I wouldn't count these changes as that, as it really just groups the code differently which is an important step on supporting themes, but IMHO not a ui change itself. I would also argue that even a limited testing by merging is better than no testing until it is a larger change. We don't need to agree, I can accept that answer as is. Often there isn't really an objective place to draw a line, but it has to be done nonetheless.

How would you like to continue? Should I ping you here when each additional step is completed or would you prefer to only get an update when the reedline part is done? Or even only once the reedline and nushell parts are completed?

fdncred:
Maybe start with a todo list and ping me on each item? I'm up for whatever is the easiest usually.

fdncred:
@adaschma are you still in on this? i was testing this PR and don't see any problems right now.

adaschma:
Hi @fdncred I haven't worked on this since. I had started some work started on the vi-mode, but even that I stopped working on about two months ago. I had some urgent life todos and I haven't been in the right head space for programming on reedline, since. I have started some other work, that I plan on finishing next week before giving reedline my full attention again. I still think it can be merged as is.

fdncred:
Sounds like a fine plan. Hope all is well on your urgent life todos. Glad to have you around.

adaschma:
It is time to admit, that I won't be continuing this is the foreseeable future. I hope closing it now will make it more likely that I start working on reedline, again.

PR #807, opened 2024-07-18
Fix vi mode change not pastable

adamschmalhofer:
While extending the tests for vi mode, I noticed that c$ (replace text until end of line with newly entered text) and cc (replace the entire line with newly entered text) don't actually cut the removed text. This fixed that. Verified with vim 9.1 that it should be cut a.k.a. pastable. 

Test:

setup:

1. $ cargo run --example demo -- --vi
2. enter a text (optionally use <Alt> + <Enter> to enter multiple line) 
3. press <esc> to enter vi normal mode and use the arrow keys to navigate somewhere in the line

actual test:

both "cc <esc> p" and "c$ <esc> p" should now leave the text unchanged by cutting out text and pasting it back again (, but moving the cursor to the end).


src/edit_mode/vi/parser.rs
```diff
@@ -510,6 +510,36 @@ mod tests {
...
+    #[case(&['c', 'e'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::CutWordRight]), ReedlineEvent::Repaint]))]
+    #[case(&['c', 'b'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::CutWordLeft]), ReedlineEvent::Repaint]))]
+    #[case(&['c', 'B'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::CutBigWordLeft]), ReedlineEvent::Repaint]))]
+    #[case(&['d', 'h'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::Backspace])]))]
+    #[case(&['d', 'l'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::Delete])]))]
+    #[case(&['2', 'd', 'd'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::CutCurrentLine]), ReedlineEvent::Edit(vec![EditCommand::CutCurrentLine])]))]
+    // #[case(&['d', 'j'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::CutCurrentLine]), ReedlineEvent::Edit(vec![EditCommand::CutCurrentLine])]))]
+    // #[case(&['d', 'k'], ReedlineEvent::Multiple(vec![ReedlineEvent::Up, ReedlineEvent::Edit(vec![EditCommand::CutCurrentLine]), ReedlineEvent::Edit(vec![EditCommand::CutCurrentLine])]))]
```
fdncred, line 524: What's the point of these commented out cases? Do they not work? Are they not supposed to work? Are we coming back to them later? etc.

adamschmalhofer: They don't work. I would rather wait on adding them because (1) using this combination in vim is uncommon [citation needed] and (2) multi line edits in nushell is not very frequent, (3) there might be strange edge cases with the menu and (4) adding it will add onto the hacks and (5) we should cleanup before that as well as (6) other lower hanging fruit in the vi mode.

adamschmalhofer: It's ready to land.

fdncred: ok, we typically freeze PR merges when we're this close to a release (tuesday). so, we'll probably have to wait on this one.

fdncred:
Thanks

PR #813, opened 2024-08-05
feat: impl ClearScreen with crossterm's Clear

### LINKED ISSUE #811: clear_screen to fit in Warp's manner
Warp always has an upper block for the last command:

image

Currently, Reedline's `clear_screen` will print empty lines to move the current insert point to the most upper position, which is hidden by the block quoted.

I don't know how `mysql` can "clear screen" (ctrl+l) properly as shown above, but this can be a compatibility request.

tisonkun:
This closes #811.

cc @fdncred @sholderbach 

I can use this with:

```rust
    let mut keybindings = reedline::default_emacs_keybindings();
    keybindings.add_binding(
        reedline::KeyModifiers::CONTROL,
        reedline::KeyCode::Char('l'),
        reedline::ReedlineEvent::ClearDisplay,
    );

    let mut state = Reedline::create()
        .with_validator(Box::new(...))
        .with_highlighter(Box::new(...))
        .with_edit_mode(Box::new(Emacs::new(keybindings)));
```

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -666,6 +666,13 @@ impl Reedline {
         self.painter.paint_line(msg)
     }
 
+    /// <placeholder docs>
+    pub fn clear_display(&mut self) -> Result<()> {
+        self.painter.clear_display()?;
+
+        Ok(())
+    }
+
     /// Clear the screen by printing enough whitespace to start the prompt or
     /// other output back at the first line of the terminal.
     pub fn clear_screen(&mut self) -> Result<()> {
@@ -847,6 +854,10 @@ impl Reedline {
             ReedlineEvent::CtrlC => {
                 self.input_mode = InputMode::Regular;
                 Ok(EventStatus::Exits(Signal::CtrlC))
+            }
+            ReedlineEvent::ClearDisplay => {
+                self.painter.clear_display()?;
+                Ok(EventStatus::Handled)
             }
             ReedlineEvent::ClearScreen => {
                 self.painter.clear_screen()?;
@@ -1071,6 +1082,11 @@ impl Reedline {
                 self.run_edit_commands(&[EditCommand::Clear]);
                 self.editor.reset_undo_stack();
                 Ok(EventStatus::Exits(Signal::CtrlC))
+            }
+            ReedlineEvent::ClearDisplay => {
+                self.deactivate_menus();
+                self.painter.clear_display()?;
+                Ok(EventStatus::Handled)
             }
             ReedlineEvent::ClearScreen => {
                 self.deactivate_menus();
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -666,13 +666,6 @@ impl Reedline {
         self.painter.paint_line(msg)
     }
 
-    /// <placeholder docs>
-    pub fn clear_display(&mut self) -> Result<()> {
-        self.painter.clear_display()?;
-
-        Ok(())
-    }
-
     /// Clear the screen by printing enough whitespace to start the prompt or
     /// other output back at the first line of the terminal.
     pub fn clear_screen(&mut self) -> Result<()> {
@@ -854,10 +847,6 @@ impl Reedline {
             ReedlineEvent::CtrlC => {
                 self.input_mode = InputMode::Regular;
                 Ok(EventStatus::Exits(Signal::CtrlC))
-            }
-            ReedlineEvent::ClearDisplay => {
-                self.painter.clear_display()?;
-                Ok(EventStatus::Handled)
             }
             ReedlineEvent::ClearScreen => {
                 self.painter.clear_screen()?;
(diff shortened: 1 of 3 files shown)


src/painting/painter.rs
```diff
@@ -507,6 +507,16 @@ impl Painter {
         self.stdout.flush()
     }
 
+    /// <placeholder docs>
+    pub(crate) fn clear_display(&mut self) -> Result<()> {
+        self.stdout.queue(cursor::Hide)?;
+        self.stdout.queue(Clear(ClearType::All))?;
```
tisonkun, line 513: I try all the type and only `All` does what I mean.

src/painting/painter.rs
```diff
@@ -507,6 +507,16 @@ impl Painter {
...
+        self.stdout.queue(MoveTo(0, 0))?;
+        self.stdout.queue(cursor::Show)?;
+        self.stdout.flush()?;
+        self.initialize_prompt_position(None)
+    }
+
     /// Clear the screen by printing enough whitespace to start the prompt or
     /// other output back at the first line of the terminal.
     pub(crate) fn clear_screen(&mut self) -> Result<()> {
```
tisonkun, line 522: The lines calculation looks correct @sholderbach. But Warp has a sticky upper block that would hide the first few lines. I don't know how Warp's people implement their terminal and it looks `terminal::size()` returns the whole window including those hidden by the block.

src/painting/painter.rs
fdncred: Have you tried with MoveTo(0,0) first and then try the other ClearType's?

tisonkun: Tried. Still almost the same.

`FromCursorDown` and others can clear the buffer for history also. I'd like to get a new screen only, and `All` keeps the buffer history that you can get back by scrolling up.

fdncred: 👍🏻

src/engine.rs
```diff
@@ -666,6 +666,13 @@ impl Reedline {
         self.painter.paint_line(msg)
     }
 
+    /// <placeholder docs>
```
tisonkun, line 669: I'll update these docs once the direction reaches a consensus (that we will implement the feature in this way).

The current patch can fit my requirement, but I can lack some of the design principles or other considerations within this project's practice. So please let me know if there is anything should be considered. And if it's good to proceed, please tell me explicitly also.

fdncred (commented):
My current objections.
1. place holder docs, mentioned already of course
2. i don't really like having clear display and clear screen so i would hope at some point (maybe now) we'd be able to combine them because it's really just confusing why we'd have both
3. are there ci tests that we could create to validate functionality?

if push comes to shove, i could live with cleardisplay and clearscreen in the short term.

tisonkun:
@fdncred Thanks for your feedback!

> i don't really like having clear display and clear screen so i would hope at some point (maybe now) we'd be able to combine them because it's really just confusing why we'd have both

Yeah. After a consideration we may directly change the `clear_screen` function to the current `clear_display` impl. @sholderbach implemented `clear_screen` directly in this way and I wonder if the impl in this PR fits the requirement in #28.

> are there ci tests that we could create to validate functionality?

If we change the impl of `clear_screen`, we don't need new tests (covered by regressions). If we have a separated `clear_display` function, I can add some.

... while my opinion is that we just do some trivial tests (ensure the keybindings cause the event triggered), instead of really check the terminal's manner (and there is no existing tests to check that).

tisonkun:
Push a new commit to showcase the reimpl alternative https://github.com/nushell/reedline/pull/813/commits/9f33d5ef64b6ff84483dea00aab6760478b7c5db

tisonkun:
@fdncred @sholderbach Any thoughts here?

sholderbach (approved):
Can only verify on Windows Terminal right now but looks good in preserving any old stuff, and the `ClearScrollback` exists for the full on purge clearing

fdncred (commented):
This is probably ok on MacOS but ctrl+l clears some of the back buffer.

Repro with this PR
in nushell do ls as many times as it takes to scroll a couple off the top of the screen
cargo run --example demo
ctrl+l
part of the back buffer is there, some is gone.

it appears to clear the screen and put the cursor on the top line but when i scroll back, i get something like this when i scroll back after ctrl+L where i thought i would have all 5 or 6 ls outputs i only have 2 and a line.
image

Like I said, I can probably live with this behavior, and I doubt it's created by the code change here. I looked at how it worked before these changes and it's a bit different. I wouldn't call either before or after this PR perfect. (On Windows with WT it looks near perfect to me)

tisonkun:
@sholderbach @fdncred Thanks for your review!

@fdncred is it possible we ship with this patch now? And I wonder if the manner you report is kind of a bug or improvement point in [crossterm](https://github.com/crossterm-rs/crossterm). That basically we use their ClearScreen but it doesn't work as expected in some term, while they declear it should be "crossterm".

fdncred:
I'm fine with landing. FYI - I tried the latest crossterm and it works similarly on macos but eats more scroll back. Since Stefan already approved. Let's go!

tisonkun:
Thank you! The last question is if there is an estimate of the next release I can get this patch in my project ..

Or where I can subscribe the release news.

fdncred:
August 20th

tisonkun:
Cool! Looking forward to upgrade :D

PR #821, opened 2024-08-24
Add config for keeping menu always open

hkalbasi:
fix #820

PR #823, opened 2024-09-04
Fix panic in long item completion in ide menu

HKalbasi:
Ide menu uses `...` for items that are very long, and if the prompt already contains some string inside of `...` it will panic. It can be reproduced by completing the largest item in the `ide_completions`, then deleting one or two characters, and hitting `Tab` again.

The panic was found in #821

### DIFF
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -512,7 +512,11 @@ impl IdeMenu {
         };
 
         if use_ansi_coloring {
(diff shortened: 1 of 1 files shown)


fdncred:
This is how I was able to repro the bug and test this fix.
1. cargo run --example ide_completions
2. h<tab>
3. the longest entry should now be selected with ... at the end
4. hit enter
5. hit backspace
6. hit tab <crash before this pr>

Thanks for fixing this. Let's go!

PR #826, opened 2024-09-12
emacs mode: change ctrl-k to `cut to line end`

WindSoilder:
When making relative pr in nushell: https://github.com/nushell/nushell/pull/13801

I found it would be good to change the keybinding here too.

This pr is going to change ctrl-k to `cut to line end`, rather than `cut to end`.


fdncred:
Thanks Wind!

PR #828, opened 2024-09-17
Added partial completions to the MenuNext event

uek-1:
This PR is meant to address a nushell specific issue (https://github.com/nushell/nushell/issues/11251) about the tab button not partially completing suggestions if the menu is already open. I added partial completions to the MenuNext event after checking if they were enabled.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -967,8 +967,9 @@ impl Reedline {
             }
             ReedlineEvent::MenuNext => {
                 if let Some(menu) = self.menus.iter_mut().find(|menu| menu.is_active()) {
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks!

blindFS:
I believe this is the cause of nushell #14152, How about we add a new event for that, and move it out from menu next, @uek-1 ?

uek-1:
This seems to be a very similar same issue to #833 . The problem seems to be caused by IdeMenu's implementation of `can_partially_complete` calling `update_values` if quick completions are off, which resets the selection index to 0 (the first element). I'll get to work on moving this out into a new menu event!

PR #831, opened 2024-09-24
Add test for kitty protocol modifier key in vi mode

replcat:
From the discussion in https://github.com/nushell/nushell/pull/13906, this is a copy-paste of the test above, demonstrating [kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol) modifier support via crossterm in vi mode.


sholderbach (approved):
Thanks for investigating this and adding the test!

PR #832, opened 2024-09-24
fix: treat `CTRL + J` as `Enter`

ErichDonGubler:
This resolves the issue I noted having in <https://github.com/nushell/nushell/issues/6427#issuecomment-1821895162>, where hitting `Enter` in MacOS and Linux outside the main `Reedline::read_line` call in raw mode results in a <kbd>CTRL + J</kbd> event that yields no actual input once `read_line` is called again.

I'm not positive that this doesn't introduce regressions with, e.g., pasted input. I'm not sure what the best way to test this would be.

I'm also not sure if it resolves that issue in its entirety, since the question of whether its scope should include `\n` is unresolved.

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -121,6 +121,7 @@ impl EditMode for Emacs {
             Event::Key(KeyEvent {
                 code, modifiers, ..
             }) => match (modifiers, code) {
+                (KeyModifiers::CONTROL, KeyCode::Char('j')) => ReedlineEvent::Enter,
                 (modifier, KeyCode::Char(c)) => {
                     // Note. The modifier can also be a combination of modifiers, for
                     // example:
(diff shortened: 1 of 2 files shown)


sholderbach:
I don't think hardcoding the binding like this is the right move, even if I can see it being part of the default binding. If you use the kitty keyboard extensions it may be possible to bind the true `Ctrl-j` then you probably want the ability to rewire it.

ErichDonGubler:
I use Wezterm, so maybe there's some configuration I need to investigate there. I'm not currently aware of a way to rebind keys like that in Wezterm, but admittedly I haven't looked.

ErichDonGubler:
Thinking about this, IIRC editors like Neovim are often interested in exposing <kbd>CTRL + J</kbd> as a separate binding from `Enter`. I don't know if those sorts of applications are interested in using `reedline`, but I suppose it _does_ mean that maybe applications should be on the hook for dealing with the distinction.

The application I want to "fix" this in is Nushell; perhaps I should contribute there instead? It is [my observation elsewhere](https://github.com/nushell/nushell/issues/6427#issuecomment-1821895162) that Nushell understands that it's receiving a <kbd>CTRL + J</kbd>.

ErichDonGubler:
Ah, looks like I can resolve this with Nushell configuration: https://github.com/nushell/nushell/issues/6427#issuecomment-2395060695

🫡 Guess I'm done here, sorry for the noise!

Issue #833, opened 2024-09-25
Can't tab-complete to highlight autcomplete entries for certain files which share prefixes


132ikl:
**Platform** Arch Linux
**Terminal software** alacritty
**Nushell** [71d604](https://github.com/nushell/nushell/commit/71d604067aefa10148ed62063034bde0ba4dfeaf)

It seems like under some circumstances I'm not able to press tab to highlight autocomplete for files. I think it's related to the files having the same prefix, but I'm not sure. I haven't fully bisected the issue, but I wasn't able to reproduce the issue on nushell [28a746](https://github.com/nushell/nushell/commit/28a7461057456555314e65e9b10cb5b996d4f72d) and was able to reproduce the issue on 71d604. There was a reedline update between those two commits which includes 2a57906, 4096b93, and 660a507. The latter seems the most related to the issue, so I suspect it's the culprint (@uek-1)

## Steps to reproduce
In nushell,
1. Make a new directory and `cd` into it
2. `touch testa testb testc`
3. `ls test<TAB>`, then continue to press tab. `testc` will not be highlighted, however you can still select it with right arrow.

132ikl:
Resolved by #834

PR #834, opened 2024-09-26
made can_partially_complete() only complete when suggesting more than the input

uek-1:
Addresses #833 . `menu.can_partially_complete()` would reset the current selection 0 and return true even if the suggested partial completion was exactly the same. Changed `can_partially_complete` in `menu_functions` to return false and not complete if the suggested partial completion is the exact same as the input.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -967,17 +967,17 @@ impl Reedline {
             }
             ReedlineEvent::MenuNext => {
                 if let Some(menu) = self.menus.iter_mut().find(|menu| menu.is_active()) {
-                    if self.partial_completions {
-                        menu.can_partially_complete(
-                            self.quick_completions,
-                            &mut self.editor,
-                            self.completer.as_mut(),
-                            self.history.as_ref(),
-                        );
-                    };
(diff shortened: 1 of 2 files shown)


fdncred:
I'm trying to test this but can't get it to work. Here's what I did.
```
cargo run --example demo
a<tab><tab><tab>
```
I can't get any of the tabs to progress. I figure I'm doing something wrong. Please let me know.

uek-1:
I think this might be an issue with the demo itself. Tabs don't seem to progress as of d8b240ade899f95b22d49a4cec995f5e19bafca8 either.  Looking at the code, I see that MenuNext is actually not bound to tab in the demo (binding it does fix issue).  https://github.com/nushell/reedline/blob/660a5074b2bd432b47e2e0bb7cca11b4509ff3df/examples/demo.rs#L244-L250

fdncred:
Can you submit a PR for that fix to the demo.rs app too please? This makes total sense.

132ikl:
Is there anything blocking this PR?

fdncred:
nope

PR #835, opened 2024-09-27
Fixed demo completions

uek-1:
Rel: #834 Added MenuNext as a menu event upon pressing tab to the demo example. Fixes the currently broken completions on the example.


fdncred:
Thanks!

Issue #757, opened 2024-02-24
external_printer does not print until I press some key


tom-code:
**Platform** macOS, linux
**Terminal software** Apple Terminal.app 

**Describe the problem you are observing.**
Lines printed using external_printer feature are printed only when some key is pressed.

## Steps to reproduce
Just run external_printer example.

Problem seems to be caused by change https://github.com/nushell/reedline/pull/651

Following change seems to fix issue for me:
```
diff --git a/src/engine.rs b/src/engine.rs
index e7bcceb..203fd1b 100644
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -695,6 +695,10 @@ impl Reedline {
 
             let mut latest_resize = None;
             loop {
+                #[cfg(feature = "external_printer")]
+                if !event::poll(Duration::from_millis(POLL_WAIT*10))? {
+                    break;
+                }
                 match event::read()? {
                     Event::Resize(x, y) => {
                         latest_resize = Some((x, y));
```
this undoes solution of https://github.com/nushell/reedline/pull/651 , but I tried to limit it to only configuration when external_printer feature is enabled and used POLL_WAIT*10 to wait 100ms instead of 10 to not waste too much cpu.

PR #840, opened 2024-10-02
Fix external printer don't print until key press

andylokandy:
Closes https://github.com/nushell/reedline/issues/757


fdncred:
Thanks

PR #844, opened 2024-10-11
Implemented change in and delete in functionality

ayax79:
This implements the change in (ci) and delete in functionality:
- ci" - Change in quotes
- di" - Delete in quotes
- ci[ - change in brackets

supports [ , ( , ", ', `, {

### DIFF
--- a/src/edit_mode/vi/command.rs
+++ b/src/edit_mode/vi/command.rs
@@ -9,7 +9,15 @@ where
     match input.peek() {
         Some('d') => {
             let _ = input.next();
-            Some(Command::Delete)
+            if let Some('i') = input.peek() {
+                let _ = input.next();
+                match input.next() {
+                    Some(c) => Some(Command::DeleteInside(*c)),
+                    None => Some(Command::Incomplete),
+                }
+            } else {
+                Some(Command::Delete)
+            }
         }
         Some('p') => {
             let _ = input.next();
@@ -33,7 +41,15 @@ where
         }
         Some('c') => {
             let _ = input.next();
-            Some(Command::Change)
+            if let Some('i') = input.peek() {
+                let _ = input.next();
+                match input.next() {
+                    Some(c) => Some(Command::ChangeInside(*c)),
+                    None => Some(Command::Incomplete),
+                }
+            } else {
+                Some(Command::Change)
+            }
         }
         Some('x') => {
             let _ = input.next();
@@ -107,6 +123,8 @@ pub enum Command {
     HistorySearch,
     Switchcase,
     RepeatLastAction,
+    ChangeInside(char),
+    DeleteInside(char),
 }
 
 impl Command {
@@ -150,10 +168,24 @@ impl Command {
             // Whenever a motion is required to finish the command we must be in visual mode
             Self::Delete | Self::Change => vec![ReedlineOption::Edit(EditCommand::CutSelection)],
             Self::Incomplete => vec![ReedlineOption::Incomplete],
-            Command::RepeatLastAction => match &vi_state.previous {
+            Self::RepeatLastAction => match &vi_state.previous {
                 Some(event) => vec![ReedlineOption::Event(event.clone())],
                 None => vec![],
             },
+            Self::ChangeInside(char) => {
+                let right = right_bracket_for(char);
+                vec![
+                    ReedlineOption::Edit(EditCommand::CutLeftBefore(*char)),
+                    ReedlineOption::Edit(EditCommand::CutRightBefore(right)),
+                ]
+            }
+            Self::DeleteInside(char) => {
+                let right = right_bracket_for(char);
+                vec![
+                    ReedlineOption::Edit(EditCommand::CutLeftBefore(*char)),
+                    ReedlineOption::Edit(EditCommand::CutRightBefore(right)),
+                ]
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/edit_mode/vi/command.rs
```diff
@@ -276,3 +308,20 @@ impl Command {
...
+        '{' => '}',
+        '<' => '>',
+        _ => *c,
+    }
+}
+
+fn is_valid_change_in(c: &char) -> bool {
+    matches!(
+        c,
+        '(' | '[' | '{' | ')' | ']' | '}' | '"' | '\'' | '`' | '<'
```
sholderbach, line 325: If you support the closing delimiters here, there needs to be the inverse logic for `left_bracket_for` as well, right?

src/edit_mode/vi/parser.rs
```diff
@@ -101,6 +101,7 @@ impl ParsedViSequence {
             (Some(Command::EnterViInsert), ParseResult::Incomplete)
             | (Some(Command::EnterViAppend), ParseResult::Incomplete)
             | (Some(Command::ChangeToLineEnd), ParseResult::Incomplete)
+            | (Some(Command::ChangeInside(_)), ParseResult::Incomplete)
```
sholderbach, line 104: Not sure how much of a deal breaker it is to land, but it will toggle into insert mode even if the delimiter is not found.

sholderbach (commented):
Thanks, such a nice convenience of vim to have!

There are certainly some mismatches we will still have due to the clunky event translation (like not advancing to the text object if before or multiplier and redo things that may need some tweaking but that probably needs a rework outside the scope of one PR)

src/edit_mode/vi/command.rs
ayax79: Yes, it should support both.. I actually never used the right bracket version. Just tested that it is a thing in neovim.

src/edit_mode/vi/parser.rs
ayax79: I added another check to make sure it was comparing against a valid change inside character before switching to insert mode.

blindFS:
I was waiting for this feature for long. It made me curious about the future plans for nushell repl parser/highlighter, are we going to use treesitter here in reedline in near future? Since without AST info, it's hard to deal with nasty cases where brackets get nested like `(foo(bar "(" ) baz)`

src/edit_mode/vi/command.rs
```diff
@@ -150,10 +176,44 @@ impl Command {
...
+                    ReedlineOption::Edit(EditCommand::CutRightBefore(right)),
+                ]
+            }
+            Self::ChangeInside(right) if is_valid_change_inside_right(right) => {
+                let left = bracket_for(right);
+                vec![
+                    ReedlineOption::Edit(EditCommand::CutLeftBefore(left)),
+                    ReedlineOption::Edit(EditCommand::CutRightBefore(*right)),
+                ]
+            }
```
sholderbach, line 196: I am thinking you could save yourself some duplication here by doing the conversion to the one bracket side when creating the `Change/DeleteInisde(char)` but then there would be also something to watch out for correctness.

sholderbach (approved):
Looking good to get that going, in principle the `i{char}` text object parsing could also move to the motion parsing part. The command logic should preempt to insert mode before it is interpreted as a traditional motion. But the gain from that refactor is not as big yet as we only have `d` and `c` accepting motions at the moment.

src/edit_mode/vi/command.rs
ayax79: I would still need to know if it is a right of left side bracket to get the ordering right.. It felt cleaner to have be a separate match statements. I did collapse the Delete and Change conditions together since they were inherently the same.

ayax79:
Another thing that should be addressed at some point.. Since this is basically supplying two actions, it requires two undo commands to get back to the previous state.

bishopmatthew:
> I was waiting for this feature for long. It made me curious about the future plans for nushell repl parser/highlighter, are we going to use treesitter here in reedline in near future? Since without AST info, it's hard to deal with nasty cases where brackets get nested like `(foo(bar "(" ) baz)`

+1 to that — I didn't realize someone was already working on this and had started looking into how one might implement the full set of vim text object operations...

> The basic pattern is <operation><scope><text object> where:

> Operation is something like d (delete), c (change), y (yank/copy), v (select)
> Scope is either i (inside) or a (around)
> Text object can be things like w (word), p (paragraph), ( (parentheses), " (quotes), } (curly braces)

Anyways — it seems like the way to be able to implement all of this stuff in a consistent way would be to use tree sitter.

- reedline could be more useful to other shells since support for other languages 
- what @blindFS mentioned about handling complex nesting — actually it seems like maybe you've already started on improving https://github.com/nushell/tree-sitter-nu in support of this?

blindFS:
@bishopmatthew tree-sitter-nu's accuracy has been improved recently on its own. Feel free to report any parsing error or WA. As for the integration to reedline, I think it's not planned yet, and a general set of api to process ast/cst would be a better option on second thought. Using treesitter may not be the only option since it needs extra overhead compared to the native parser.

Actually there's a hacking way (in case you want the functionality right now) for limited textobj related ops with current reedline implementation: mapping to user defined command with internal ast and commandline for the dirty job.

Issue #521, opened 2022-12-02
nushell consumes CPU when idle


WGH-:
### Describe the bug

nushell in interactive mode consumes some CPU time

### How to reproduce

1. Launch `nu` in a terminal (interactive mode)
2. Observe that `nu` process consumes some amount of CPU time, about 1%

### Expected behavior

0% CPU consumed, sleeping in `epoll_wait` or something

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.72.1                                   |
| branch             | main                                     |
| commit_hash        | fcdc4747311dd9075bdfd24bbc846144fda924f8 |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.65.0 (gentoo)                    |
| cargo_version      | cargo 1.65.0                             |
| pkg_version        | 0.72.1                                   |
| build_time         | 2022-12-02 16:08:55 +03:00               |
| build_rust_channel | debug                                    |
| features           | database, default, trash, which, zip     |
| installed_plugins  |                                          |

### Additional context

`strace`:
```
<...>
epoll_wait(5, [], 3, 0)                 = 0
epoll_wait(5, [], 3, 0)                 = 0
epoll_wait(5, [], 3, 99)                = 0
epoll_wait(5, [], 3, 0)                 = 0
epoll_wait(5, [], 3, 99)                = 0
epoll_wait(5, [], 3, 0)                 = 0
epoll_wait(5, [], 3, 0)                 = 0
epoll_wait(5, [], 3, 99)                = 0
epoll_wait(5, [], 3, 0)                 = 0
epoll_wait(5, [], 3, 99)                = 0
epoll_wait(5, [], 3, 99)                = 0
epoll_wait(5, [], 3, 99)                = 0
epoll_wait(5, [], 3, 99)                = 0
epoll_wait(5, [], 3, 0)                 = 0
<...>
```

Stack trace:
```
(gdb) bt
#0  0x00007f2adee3990a in epoll_wait (epfd=5, events=0x556afffb23d0, maxevents=3, timeout=99)
    at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
nushell/nushell#1  0x0000556afdf07585 in mio::sys::unix::selector::epoll::Selector::select (self=0x556affdf68c0, events=0x556affdf68d0,
    timeout=...) at src/sys/unix/selector/epoll.rs:68
[...]

PR #846, opened 2024-10-17
Reduce typing latency caused by `POLL_WAIT`

### LINKED ISSUE #643: POLL_WAIT hurts user experience

YizhePKU:
This PR makes it so that Reedline no longer wait for `POLL_WAIT` after every keystroke. Instead, the delay is only triggered if text pasting or resizing is detected.

It also re-introduces a fix from PR #651. That PR was reverted by #840 because of issue #757 about external printers. If the feature `external_printer` is not enabled, this PR will just block on input forever for maximum power efficiency. Users with `external_printer` enabled will also see a small improvment in power efficiency, because this PR makes it possible to use a longer polling inteval without affacting typing latency.

Fixes #643, [#10484](https://github.com/nushell/nushell/issues/10484), and #521, and hopeful doesn't break #757.

### DIFF
--- a/src/edit_mode/emacs.rs
+++ b/src/edit_mode/emacs.rs
@@ -194,7 +194,7 @@ mod test {
     #[test]
     fn ctrl_l_leads_to_clear_screen_event() {
         let mut emacs = Emacs::default();
-        let ctrl_l = ReedlineRawEvent::convert_from(Event::Key(KeyEvent::new(
+        let ctrl_l = ReedlineRawEvent::try_from(Event::Key(KeyEvent::new(
             KeyCode::Char('l'),
             KeyModifiers::CONTROL,
         )))
@@ -214,7 +214,7 @@ mod test {
         );
 
         let mut emacs = Emacs::new(keybindings);
-        let ctrl_l = ReedlineRawEvent::convert_from(Event::Key(KeyEvent::new(
+        let ctrl_l = ReedlineRawEvent::try_from(Event::Key(KeyEvent::new(
             KeyCode::Char('l'),
             KeyModifiers::CONTROL,
         )))
@@ -227,7 +227,7 @@ mod test {
     #[test]
     fn inserting_character_works() {
         let mut emacs = Emacs::default();
-        let l = ReedlineRawEvent::convert_from(Event::Key(KeyEvent::new(
+        let l = ReedlineRawEvent::try_from(Event::Key(KeyEvent::new(
             KeyCode::Char('l'),
             KeyModifiers::NONE,
         )))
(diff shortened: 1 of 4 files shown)


YizhePKU:
PR draft #737 discussed reworking the external printer API. I wonder how is that going?

sholderbach:
Nushell doesn't make use of the external printer, so politically it takes precedence in that decision in my view. We should prioritize the battery saving behavior and figure out a solution for a responsive external printer in due course. (as the stalling of #737 shows, probably non trivial refactor needed.)

YizhePKU:
> We should prioritize the battery saving behavior and figure out a solution for a responsive external printer in due course.

This PR aims to be maximally power efficient when the feature `external_printer` is not enabled, so I think that's covered.

PR #849, opened 2024-11-12
Change in word for vi key mappings

ayax79:
Implements change in word (ciw) keymapping.


fdncred:
Seems pretty harmless. Just some refactor and adding some w stuff. I don't use vi mode but I'd vote to land it.

ayax79:
Closing out as I will not be able to fix bugs that @sholderbach brought up in this implementation.

Issue #792, opened 2024-05-20
Default keybind for "newline"


NotTheDr01ds:
Nushell pipelines are best split over multiple lines, and the `nufmt` [specification](https://github.com/nushell/nufmt/blob/main/docs/specification.md) recommends "wall of pipes" formatting.  E.g.:

```Nushell
ls
| where size > 0kb
| insert name_length {$in.name | str length}
```

However, that's extremely tedious to create "out of the box" with Nushell/Reedline at the moment, since a <kbd>Enter</kbd> only creates a newline in this case if there's a **trailing** pipe character on the existing line.  E.g.:

* `ls |` + <kbd>Enter</kbd> will create a newline which can be followed by the next expression, but
* `ls` + <kbd>Enter</kbd> executes the command (as it should - I'm not advocating for that to change ;-).

In order to create the "wall of pipes", you have to start with the first form `ls |`, hit <kbd>Enter</kbd> , then continue your `| pipeline`, and then (if you're pedantic) remove the trailing pipe from the previous line.  You could also do something like start the pipeline with a `(` open-paran to keep <kbd>Enter</kbd> from executing the pipeline until complete.

While it's relatively easy for a user to create a keybinding in their config to map <kbd>Alt</kbd>+<kbd>Enter</kbd> (assuming macOS <kbd>⌘</kbd>?), given the importance of newline in Nushell CLI-usage, I'd recommend this be a built-in keybinding.

Note:  This is built-in to the Fish shell.

fdncred:
Ya, I've always thought the pipes belong on the right side for the same reasons. I have this in my keybindings because the problem with alt-enter is that it's not free in all terminals.
```
  {
    name: insert_newline
    modifier: alt
    keycode: char_i
    mode: [emacs vi_normal vi_insert]
    event: { edit: insertnewline }
  }
  {
    name: insert_newline
    modifier: alt
    keycode: enter
    mode: [emacs vi_normal vi_insert]
    event: { edit: insertnewline }
  }
```
But I wouldn't have a problem with adding these to the default_config.nu and in the rust code.

sholderbach:
Closed by #854

PR #854, opened 2024-11-19
Add Newline keybindings

NotTheDr01ds:
Implements #792 and also should be the proper implementation of https://github.com/nushell/nushell/pull/13606

---

Adds <kbd>Alt</kbd>+<kbd>Enter</kbd> and <kbd>Shift</kbd>+<kbd>Enter</kbd> as keybindings to insert a newline in Emacs and Vi-insert modes.

In the Nushell PR, @sholderbach said:

> so it might be that for some `Alt-Enter` won't work while for others `Shift-Enter` or `Ctrl-Enter` is unreliable.

I figure there's no reason we can't just bind *both* so that multiple platforms and terminals can handle this gracefully.  I know <kbd>Alt</kbd>+<kbd>Enter</kbd> is commonly captured by Windows Terminal, and <kbd>Shift</kbd>+<kbd>Enter</kbd> doesn't seem to be recognized on Linux (also via `input listen`).

If the Terminal captures one of the keybindings, then hopefully the other is available.  And if not, the user can always add their own in their Nushell config.

### DIFF
--- a/src/edit_mode/keybindings.rs
+++ b/src/edit_mode/keybindings.rs
@@ -234,6 +234,7 @@ pub fn add_common_edit_bindings(kb: &mut Keybindings) {
     );
     kb.add_binding(KM::ALT, KC::Enter, edit_bind(EC::InsertNewline));
     kb.add_binding(KM::SHIFT, KC::Enter, edit_bind(EC::InsertNewline));
+    kb.add_binding(KM::CONTROL, KC::Char('j'), edit_bind(EC::InsertNewline));
 }
 
 pub fn add_common_selection_bindings(kb: &mut Keybindings) {
--- a/src/edit_mode/keybindings.rs
+++ b/src/edit_mode/keybindings.rs
@@ -232,6 +232,8 @@ pub fn add_common_edit_bindings(kb: &mut Keybindings) {
         KC::Char('v'),
         edit_bind(EC::PasteSystem),
     );
+    kb.add_binding(KM::ALT, KC::Enter, edit_bind(EC::InsertNewline));
+    kb.add_binding(KM::SHIFT, KC::Enter, edit_bind(EC::InsertNewline));
 }
 
 pub fn add_common_selection_bindings(kb: &mut Keybindings) {


NotTheDr01ds:
Alt+Enter is not a system wide binding.  I use it in Nushell in Windows Terminal personally. It's bound by default in Windows Terminal, but it's also easy to unbind it.

fdncred:
Ya, I think you're right about the system level aspect but Alt-Enter does the same thing in WT, cmd.exe, alacritty.exe, wezterm.exe, yori.exe, tabby.exe. There may be some terminal app on Windows that doesn't use it, but I haven't found it yet.

NotTheDr01ds:
> There may be some terminal app on Windows that doesn't use it, but I haven't found it yet.

Interesting - I guess us old-time Fish shell users have just gotten used to unbinding it, since Fish *only* provides <kbd>Alt</kbd>+<kbd>Enter</kbd>.

> I'm not opposed to multiple keybindings but if we know Alt-Enter never works on Windows, we might want to only apply it in non-Windows environments. Maybe?

The question is - Why disallow it?  Binding it in Reedline/Nushell is never going to cause any problems<sup>1</sup>.  That's why I think we should just go ahead and provide multiple options.

<sup>1</sup> AFAIK, keybindings are always handled in the following order (hand-waving over a few, like hardware):

1. If the OS or window-manager has the keybinding, it will act on it and (typically) not pass it on further.  If it is not handled, then it gets passed to the ...
2. Terminal, which if it has the keybinding, will act on it and (typically) not pass ito n further.  If it is not handled, then it gets passed to the ...
3. Application/shell/line-editor

So, since Reedline will only get the <kbd>Alt</kbd>+<kbd>Enter</kbd> event if it is unbound at the higher-levels, it shouldn't matter if we just leave the keybinding in the code for all platforms, right?

NotTheDr01ds:
Do we want a third-fallback like <kbd>Ctrl</kbd>+<kbd>J</kbd> (Emacs)? Someone mentioned using that, and again, there doesn't seem to be any harm in binding it.  Users who want to bind any of these to another function are free to do so in Nushell.

fdncred:
I'm good with double binding it. It's not worth all this conversation. We can always change it if it gets in the way.

fdncred:
Thanks

sholderbach:
`Ctrl-J` should probably be `Enter` as folks have been complainig that their tools that send `\n` (that is `Ctrl-J` in ansi) don't trigger a run as the other shells do that also bind it. (The Enter key itself gets represented by `\r`/`Ctrl-M`)

See #832 (wasn't merged by me because of the hardcoding but should be fine as a soft binding like you did)

fdncred:
oops, should we revert?

Issue #684, opened 2023-12-19
Redraw prompt on resize


fdncred:
**Platform** WSLg
**Terminal software** ghostty and iterm2

The prompt is disappearing when terminal window is resized in Ghostty
![disappearing-prompt](https://github.com/nushell/reedline/assets/343840/8c05de10-13e6-426e-867d-4115b388ea70)

This is the explanation I received from the author of the Ghostty terminal @mitchellh (thank you for the answer).

----

When Ghostty sees a "prompt start" semantic prompt message from shell integration, it also assumes that the shell integration is capable of redrawing the prompt on resize events. On future resize events, Ghostty will clear the currently active prompt line if the cursor is at a prompt and assume the shell will redraw it. This should be the same behavior as Kitty.

The reason for this is that it enables complex and multi-line prompts to cleanly redraw when being resized. For example, look at the message with my prompt and iTerm2 (which doesn't have this feature). My shell is fish. The code to do this is in the src/shell-integration folder in Ghostty.

Notice iTerm2 the prompt gets completely screwed up, but on Ghostty the prompt and text remains clean throughout multiple text reflows. This is due to this redraw feature. 

iterm2

https://github.com/nushell/reedline/assets/343840/ee6443aa-e6ca-4f39-881c-c0352a928e33

Ghostty

https://github.com/nushell/reedline/assets/343840/82e75336-9734-48e7-ae26-2b4eac23a5ba

My guess is nushell is sending the "prompt start" semantic prompt OSC sequence, but is NOT doing the prompt redraw behavior. I'm not sure how Kitty handles this with nushell... (if at all)

----

We've been seeing weird prompt-drawing/redrawing issues for a while. I, for one, had no idea that terminals would erase the prompt while waiting for a resize. We need to look into this because that seems to be the case. We need to detect the resize and repaint the prompt. I'm not sure if it's a good idea to try and detect the terminal (kitty, ghostty, iterm2, etc) and only redraw for a few terminals or if we should just keep-it-simple and always redraw the prompt on resize. I'm also not sure if we should wait for all the resize messages to stop coming in before we repaint the prompt, or what heuristic we should use exactly.

Also not sure if this is better or worse since https://github.com/nushell/reedline/pull/675. I haven't tried an earlier version. /cc @danielsomerfield 

[...]

PR #857, opened 2024-11-25
Redraw on resize

milomg:
I'm not quite sure if this is the right approach, but this seems to partially fix the issue in ghostty reported in #684. Unfortunately, this still briefly shows a blank prompt while resizing, but eventually settles to a redrawn prompt


fdncred:
oh, wow, that seems like a small change. very interesting. thanks!

fdncred:
I've tested it and it seems to work and it's better than before. Let's run with it. Thanks!

Issue #820, opened 2024-08-23
Keeping completion menu always open


hkalbasi:
Currently it is possible to trigger opening the completion menu by a specific key binding like tab, but my usage is very completion heavy and I want to keep the completion menu always open for more discoverability and ease of use. Is it possible to configure reedline so that it keeps the completion menu always open, or reopen it after each key or each space? If not, would you accept a PR adding it?

fdncred:
I don't think it's possible right now. I'm always up for cool PRs as long as they're implemented well. I know that's vague but we're a little strict on what we accept here as to keep the maintenance easier. I'd say go for it, and we might accept it.

kronberger-droid:
dublicate of #356 is being worked on in #821

Issue #789, opened 2024-05-15
Modal Menus


CabalCrow:
Currently you need to press a keybind to activate any of the menus (help, completion, etc.). Being able to see the menu suggestions all the time would be a great - similar to how you would use autocompletion in IDE.
Modal menus would mean that instead of having to press a keybind to activate them, they are always active. Instead of pressing a keybind to activate the menu itself, the keybinds can be used to switch between different menu - hence modal.

The downside of this approach is that menus take a big part of the screen & this would hide the output from previously ran commands. This could make it hard to work with smaller terminals. Because of this, modal menus should be an opt-in option.

## References
zsh has a plugin that does this for the completion menu in zsh: https://github.com/marlonrichert/zsh-autocomplete
It also has an extra feature for fuzzy searching through completion or history items as seperate menus (ctrl-s/r in zsh-autocomplete).

kronberger-droid:
dublicate of #356 and is worked on in #821

PR #608, opened 2023-07-20
Add inapplicable condition on menu activation and navigation

Abdillah:
Allow `until` keybinding with `send: menu*` (activation and navigation) stacked with other action by defining some `Inapplicable` conditions.

Fixes #603.
Fixes #604.

PR #859, opened 2024-12-03
Modal Modes

CabalCrow:
This is directly related to: #789
It also could be a potential solution to: #820 

The main thing I did was disable the deactivation of menus if you have the modal_menu option set to true, and instead you need
to use a ReedlineEvent to force the deactivation of the menus in order to submit your message on the reedline.
This works almost in the ide_completions_with_modal.rs example (based on ide_completions, with just additional keybinds and option for modal_menus set) - it still has some visual issues, that I would iron out.

The reason why I'm making this draft PR is because there is an issue I'm not sure how to handle. While I can get this working in the examples, building nushell produces a different behaviour. With just the main branch of reedline, the examples allow you to stay in the completion menus which pressing backspace - however in nushell backspace throws you out of menu completion. Because of this nushell and the reedline examples have different behaviour when it comes to the reedline completion.

Is there a way to get the reedline working the same way in nushell?

### DIFF
--- a/examples/ide_completions_with_modal.rs
+++ b/examples/ide_completions_with_modal.rs
@@ -28,7 +28,7 @@ fn add_menu_keybindings(keybindings: &mut Keybindings) {
     keybindings.add_binding(
         KeyModifiers::ALT,
         KeyCode::Char('f'),
-        ReedlineEvent::ForceDeactivate,
+        ReedlineEvent::Multiple(vec![ReedlineEvent::ForceDeactivate, ReedlineEvent::Enter]),
     );
 }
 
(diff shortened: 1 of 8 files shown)


fdncred:
I haven't studied all this code but I wonder if it could be because nushell uses NuMenuCompleter which may be different than what reedline uses internally.

https://github.com/nushell/nushell/blob/c63bb81c3e6af38e4b528aac7eb6d491f0132758/crates/nu-cli/src/reedline_config.rs#L521-L536

kronberger-droid:
This overlaps with #821 and #608 and hasn't moved for long.
I would close this to tidy up a little, feel free to open another one if you want to pick up work with current main as base.

PR #867, opened 2024-12-31
 Improve Visual Mode Selection and Command Consistency

### LINKED ISSUE #865: Vi mode: 's' key not working in visual mode
**Platform** Linux x86_64
**Terminal software** Terminal independent. I've tried Alaritty and Ghostty

Describe the problem you are observing:

## Steps to reproduce
1. Standard nulshell, with `$env.config.edit_mode = vi`
2. Press following keys starting from empty command line under insert mode, type `foo123bar`, then press keys:  `ESC, 6, h, v, l, l, s`. This enters normal mode, move cursor on "1" and enter visual mode to select "123". The **bug behavior" is when pressing `s` key it only deletes character "3" but it should delete "123" altogether.

#### Recording

out

[asciicast](https://asciinema.org/a/gk9NPfqfohUmvgC3YN1fWD1vB)

## Claude Sonnet's analysis, FYI

The problem is that while the 's' key is properly handled in normal mode through the Command::SubstituteCharWithInsert command, there's no special handling for it in visual mode.

Here's what I found:

In vi/command.rs, the 's' key is parsed into Command::SubstituteCharWithInsert in normal mode
In vi/mod.rs, visual mode is entered when 'v' is pressed in normal mode
However, when in visual mode (ViMode::Visual), the code doesn't have special handling for the 's' command to delete the selection and enter insert mode
The fix would need to:

Add visual mode handling for the 's' key in the command parser
Make it behave similar to 'c' in visual mode (change/delete selection and enter insert mode)

deephbz:
# Fix Vi Mode Visual Selection Behavior

This PR addresses several inconsistencies in Reedline's Vi mode implementation, particularly around visual mode and cut operations, making the behavior more aligned with standard Vim.

### 1. Visual Mode Selection Clearing
- Added proper selection clearing when pressing Esc
- Fixed by adding explicit cursor movements to reset selection state
- Ensures visual feedback matches mode state

### 2. Cut Operations Consistency
- Fixed `x`, `s`, `d`, and `c` commands to behave consistently with Vim standards
- Made `x` properly cut selection in visual mode instead of just one character
- Fixed `c` to enter insert mode after cutting selection
- Corrected selection range calculation to be inclusive (include the last character)

### 3. Mode Transitions
- Properly restore normal mode after cut operations in visual mode
- Added mode transitions for `DeleteChar` and `DeleteToEnd` commands
- Fixed mode switching timing to happen before command execution

## Technical Details
- Modified `get_selection()` to include the last character in selection range
- Updated command parsing to handle mode transitions correctly
- Added proper mode restoration for all cut-related commands

## Testing
The changes have been tested with the following scenarios:
- Visual mode selection and Esc behavior
- Cut operations in both normal and visual modes
- Mode transitions after various commands
- Selection range calculations

Fixes #865 .
I've done `cargo fmt --all`, `cargo clippy`, and `cargo test` for each of these three commits.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -568,12 +568,14 @@ impl Editor {
             if self.insertion_point() > selection_anchor {
                 (
                     selection_anchor,
-                    (self.insertion_point() + 1).min(buffer_len),
+                    self.line_buffer.grapheme_right_index().min(buffer_len),
                 )
             } else {
                 (
                     self.insertion_point(),
-                    (selection_anchor + 1).min(buffer_len),
+                    self.line_buffer
+                        .grapheme_right_index_from_pos(selection_anchor)
+                        .min(buffer_len),
                 )
             }
         })
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -564,10 +564,17 @@ impl Editor {
     /// The range is guaranteed to be ascending.
     pub fn get_selection(&self) -> Option<(usize, usize)> {
         self.selection_anchor.map(|selection_anchor| {
+            let buffer_len = self.line_buffer.len();
             if self.insertion_point() > selection_anchor {
-                (selection_anchor, self.insertion_point())
+                (
+                    selection_anchor,
+                    (self.insertion_point() + 1).min(buffer_len),
+                )
             } else {
-                (self.insertion_point(), selection_anchor)
+                (
+                    self.insertion_point(),
+                    (selection_anchor + 1).min(buffer_len),
+                )
             }
         })
     }
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -655,6 +655,10 @@ impl Editor {
         self.delete_selection();
         insert_clipboard_content_before(&mut self.line_buffer, self.cut_buffer.deref_mut());
     }
+
+    pub(crate) fn reset_selection(&mut self) {
+        self.selection_anchor = None;
+    }
 }
 
 fn insert_clipboard_content_before(line_buffer: &mut LineBuffer, clipboard: &mut dyn Clipboard) {
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -90,7 +90,7 @@ impl EditMode for Vi {
                             ReedlineEvent::None
                         } else if res.is_complete(self.mode) {
                             let event = res.to_reedline_event(self);
-                            if let Some(mode) = res.changes_mode() {
+                            if let Some(mode) = res.changes_mode(self.mode) {
                                 self.mode = mode;
                             }
                             self.cache.clear();
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -143,12 +143,7 @@ impl EditMode for Vi {
                     self.cache.clear();
                     self.mode = ViMode::Normal;
                     ReedlineEvent::Multiple(vec![
-                        // Move left then right to clear the selection.
-                        // Order matters here: this makes sure cursor does not move at end of line
-                        ReedlineEvent::Edit(vec![
-                            EditCommand::MoveLeft { select: false },
-                            EditCommand::MoveRight { select: false },
-                        ]),
+                        ReedlineEvent::ResetSelection,
                         ReedlineEvent::Esc,
                         ReedlineEvent::Repaint,
                     ])
@@ -201,10 +196,7 @@ mod test {
         assert_eq!(
             result,
             ReedlineEvent::Multiple(vec![
-                ReedlineEvent::Edit(vec![
-                    EditCommand::MoveLeft { select: false },
-                    EditCommand::MoveRight { select: false }
-                ]),
+                ReedlineEvent::ResetSelection,
                 ReedlineEvent::Esc,
                 ReedlineEvent::Repaint
             ])
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -142,7 +142,16 @@ impl EditMode for Vi {
                 (_, KeyModifiers::NONE, KeyCode::Esc) => {
                     self.cache.clear();
                     self.mode = ViMode::Normal;
-                    ReedlineEvent::Multiple(vec![ReedlineEvent::Esc, ReedlineEvent::Repaint])
+                    ReedlineEvent::Multiple(vec![
+                        // Move left then right to clear the selection.
+                        // Order matters here: this makes sure cursor does not move at end of line
+                        ReedlineEvent::Edit(vec![
+                            EditCommand::MoveLeft { select: false },
+                            EditCommand::MoveRight { select: false },
+                        ]),
+                        ReedlineEvent::Esc,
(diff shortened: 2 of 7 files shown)


src/edit_mode/vi/mod.rs
```diff
@@ -143,7 +142,16 @@ impl EditMode for Vi {
...
                     self.cache.clear();
                     self.mode = ViMode::Normal;
-                    ReedlineEvent::Multiple(vec![ReedlineEvent::Esc, ReedlineEvent::Repaint])
+                    ReedlineEvent::Multiple(vec![
+                        // Move left then right to clear the selection.
+                        // Order matters here: this makes sure cursor does not move at end of line
+                        ReedlineEvent::Edit(vec![
+                            EditCommand::MoveLeft { select: false },
+                            EditCommand::MoveRight { select: false },
+                        ]),
```
sholderbach, line 151: As we are controlling things anyways, wouldn't a new `ReedlineEvent::ResetSelection` be easier and much clearer?

src/core_editor/editor.rs
```diff
@@ -564,10 +564,17 @@ impl Editor {
     /// The range is guaranteed to be ascending.
     pub fn get_selection(&self) -> Option<(usize, usize)> {
         self.selection_anchor.map(|selection_anchor| {
+            let buffer_len = self.line_buffer.len();
             if self.insertion_point() > selection_anchor {
-                (selection_anchor, self.insertion_point())
+                (
+                    selection_anchor,
+                    (self.insertion_point() + 1).min(buffer_len),
```
sholderbach, line 571: byte index + 1 will run into the issue that this may advance into part of a multibyte UTF-8 character, producing nonsense and a panic downstream.

src/edit_mode/vi/parser.rs
```diff
@@ -108,13 +108,19 @@ impl ParsedViSequence {
             | (Some(Command::RewriteCurrentLine), ParseResult::Incomplete)
             | (Some(Command::SubstituteCharWithInsert), ParseResult::Incomplete)
             | (Some(Command::HistorySearch), ParseResult::Incomplete)
-            | (Some(Command::Change), ParseResult::Valid(_)) => Some(ViMode::Insert),
+            | (Some(Command::Change), ParseResult::Valid(_))
+            | (Some(Command::Change), ParseResult::Incomplete) => Some(ViMode::Insert),
```
sholderbach, line 112: Not sure if the mode change should occur for the `Change`/`Incomplete` combination already? At least in normal mode we should remain there. Different story in Visual mode, so maybe that is what you are trying to achieve. 

We just may want to defend against it becoming an issue in normal mode.

Same for the delete part.

sholderbach (commented):
Thanks for tackling the VI mode, very much lacking and deserving attention!

src/core_editor/editor.rs
deephbz: Do you think the `min(buffer_len)` can help here?
If not would you kindly suggest alternative way to approach the wrong current selection range issue?

src/edit_mode/vi/mod.rs
deephbz: Sounds great I agree an atomic `ReedlineEvent::ResetSelection` is much better. Implemented in the 4th commit just pushed (original PR has 3 commits).

src/edit_mode/vi/parser.rs
deephbz: Agree the mode change depends on **current mode**. Fixed this in the 5th commit of this PR (if you'd like to squash it works together with the 1st commit).

deephbz:
Minor note: Github incorrectly introduced another "participant" because I messed up my `.gitconfig` "name" setting. It's fixed now

src/core_editor/editor.rs
johanneshorner: Instead of `(selection_anchor, self.insertion_point()` you could use `(selection_anchor, self.line_buffer.grapheme_right_index())`.
For the `else` branch i am not sure. You could change the `grapheme_right_index` method to allow passing a 'cursor position' and then call it with `selection_anchor`. Along the lines of:
```
    pub fn grapheme_right_index_from_pos(&self, pos: usize) -> usize {
        self.lines[pos..]
            .grapheme_indices(true)
            .nth(1)
            .map(|(i, _)| pos + i)
            .unwrap_or_else(|| self.lines.len())
    }
```

sholderbach: Yup, something along those lines sounds right. We advance the cursor always a full grapheme (e.g. so you don't get stuck between an accent and a vowel or the components of an emoji if they are composite and rendered together)

Here it is also worth checking if changing the semantics of the selection for Vi visual mode doesn't introduce a regression for our selection stuff in the emacs mode.

src/edit_mode/vi/parser.rs
sholderbach: Looks good, at first I was wondering about the special delete versions you had in the previous versoins but they are unreachable in visual mode due to all of them starting with `d`... Oh do we have to handle an `x` here?

deephbz: I believe `x` should be handled, to be consistent with most Vim editors (including the classic `vi`): Under visual mode user is able to select a range and delete with either `x` or `d`.

src/core_editor/editor.rs
deephbz: Goot catch! Thank you Johannes for the helpful example. I've made the change as you suggest, tested it and added unittests for the new `grapheme_right_index_from_pos` method.

It looks we've agreed on all points of this PR. Please help review and feel free to throw qs / edit / merge.

sholderbach:
Trying out the visual selection while the `v`-characterwise mode seems to behave as I would expect for the deletion itself, the highlight lags one character behind in my testing of a478473

1. insert `blurb`
2. move to `l`
3. press `l`
grafik

132ikl:
hey @deephbz would you be able to finish this one up? it looks like we should be good if the styling issue mentioned by @sholderbach is resolved?

deephbz:
> Trying out the visual selection while the `v`-characterwise mode seems to behave as I would expect for the deletion itself, the highlight through the background lags one character behind in my testing of [a478473](https://github.com/nushell/reedline/commit/a4784733f9bb6835b990952cf2ee70594b44d93d) while the foreground changes.
> 
> 1. insert `blurb`
> 2. move to `l`
> 3. press `l`
>    grafik
> 
> Are we running into an unfortunate situation regarding the styling? (setting background for the selection and inverting it again for the cursor?)

I cannot reproduce this. Below is my recording:
out

Even if there's small rendering/styling issue, I think we could address that in future PRs given current PR functions well in a self-contained way and it's blocking others

deephbz:
> hey @deephbz would you be able to finish this one up? it looks like we should be good if the styling issue mentioned by @sholderbach is resolved?

See my above reply

132ikl:
hmm, I also can't reproduce it. I also tried using the demo example with `--vi` like Stefan did and I don't seem to experience the issue. @sholderbach could you see if you can still reproduce it?

sholderbach:
Possibly there is an interaction with the way Windows Terminal renders the ANSI background coloring in my case. The char under cursor is still somewhat recognizable due to the missing syntax highlighting there.

In the interest of moving things along let's merge this and revisit the cursor styling in due time.

PR #868, opened 2024-12-31
Implement vi-mode "Yank" (copy):

deephbz:
Implement vi-mode "Yank" (copy):
    - Vi Command;
    - Editor Command;
    - Editor methods;
    - Mode changes back to Normal after yank

Note only last commit in this PR should be reviewed as part of this PR as this is on top of:
- [x] #867 
- [ ]  #874

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -120,6 +120,49 @@ impl Editor {
             EditCommand::CutSelection => self.cut_selection_to_cut_buffer(),
             EditCommand::CopySelection => self.copy_selection_to_cut_buffer(),
             EditCommand::Paste => self.paste_cut_buffer(),
+            EditCommand::CopyFromStart => self.copy_from_start(),
+            EditCommand::CopyFromLineStart => self.copy_from_line_start(),
+            EditCommand::CopyToEnd => self.copy_from_end(),
+            EditCommand::CopyToLineEnd => self.copy_to_line_end(),
+            EditCommand::CopyWordLeft => self.copy_word_left(),
+            EditCommand::CopyBigWordLeft => self.copy_big_word_left(),
+            EditCommand::CopyWordRight => self.copy_word_right(),
+            EditCommand::CopyBigWordRight => self.copy_big_word_right(),
+            EditCommand::CopyWordRightToNext => self.copy_word_right_to_next(),
+            EditCommand::CopyBigWordRightToNext => self.copy_big_word_right_to_next(),
+            EditCommand::CopyRightUntil(c) => self.copy_right_until_char(*c, false, true),
+            EditCommand::CopyRightBefore(c) => self.copy_right_until_char(*c, true, true),
+            EditCommand::CopyLeftUntil(c) => self.copy_left_until_char(*c, false, true),
+            EditCommand::CopyLeftBefore(c) => self.copy_left_until_char(*c, true, true),
+            EditCommand::CopyCurrentLine => {
+                let range = self.line_buffer.current_line_range();
+                let copy_slice = &self.line_buffer.get_buffer()[range];
+                if !copy_slice.is_empty() {
+                    self.cut_buffer.set(copy_slice, ClipboardMode::Lines);
+                }
+            }
+            EditCommand::CopyLeft => {
+                let insertion_offset = self.line_buffer.insertion_point();
+                if insertion_offset > 0 {
+                    let left_index = self.line_buffer.grapheme_left_index();
+                    let copy_range = left_index..insertion_offset;
+                    self.cut_buffer.set(
+                        &self.line_buffer.get_buffer()[copy_range],
+                        ClipboardMode::Normal,
+                    );
+                }
+            }
+            EditCommand::CopyRight => {
+                let insertion_offset = self.line_buffer.insertion_point();
+                let right_index = self.line_buffer.grapheme_right_index();
+                if right_index > insertion_offset {
+                    let copy_range = insertion_offset..right_index;
+                    self.cut_buffer.set(
+                        &self.line_buffer.get_buffer()[copy_range],
+                        ClipboardMode::Normal,
+                    );
+                }
+            }
             #[cfg(feature = "system_clipboard")]
             EditCommand::CutSelectionSystem => self.cut_selection_to_system(),
             #[cfg(feature = "system_clipboard")]
@@ -127,6 +170,7 @@ impl Editor {
             #[cfg(feature = "system_clipboard")]
             EditCommand::PasteSystem => self.paste_from_system(),
             EditCommand::CutInside { left, right } => self.cut_inside(*left, *right),
+            EditCommand::YankInside { left, right } => self.yank_inside(*left, *right),
         }
         if !matches!(command.edit_type(), EditType::MoveCursor { select: true }) {
             self.selection_anchor = None;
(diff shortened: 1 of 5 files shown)


sholderbach (commented):
Thanks for giving this a shot and putting in the work! One thing that makes me hesitant about this, that the copying we do does not work with multipliers at all as each `EditCommand` will update the clipboard. Also operations you define in terms of two `EditCommand`s suffer the same unfortunate fate.

One solution to that would be to do extra state management around clipboard but keep the old edit commands (suggested in #631 ) or otherwise rethink how we implement the vi mode interface to use an internal selection build up from the motion before performing the action. The latter would require way fewer additional `EditCommand` to be maintained, like the ones you propose here. Related issues where that came up was the impl of #849 where there are cases where the chaining of EditCommands did not achieve the expected result.

deephbz:
> Thanks for giving this a shot and putting in the work! One thing that makes me hesitant about this, that the copying we do does not work with multipliers at all as each `EditCommand` will update the clipboard. Also operations you define in terms of two `EditCommand`s suffer the same unfortunate fate.
> 
> One solution to that would be to do extra state management around clipboard but keep the old edit commands (suggested in #631 ) or otherwise rethink how we implement the vi mode interface to use an internal selection build up from the motion before performing the action. The latter would require way fewer additional `EditCommand` to be maintained, like the ones you propose here. Related issues where that came up was the impl of #849 where there are cases where the chaining of EditCommands did not achieve the expected result.

What you say are valid concerns and I do agree in the future a re-think/refactor is needed to make Vi mode elegant and better.  
However, I think it's still of great utility (and also harmless) to merge this PR in order to make Nushell's Vi mode "feature-complete", in a sense that 99% users' 99% use cases are covered. It's really common to yank and paste (like moving CLI options) so missing it is a barrier to onboard new years.

src/edit_mode/vi/command.rs
```diff
@@ -214,6 +246,23 @@ impl Command {
...
+                    ReedlineOption::Edit(EditCommand::CopyRightBefore(right)),
+                ]
+            }
+            Self::YankInside(right) if is_valid_change_inside_right(right) => {
+                let left = bracket_for(right);
+                vec![
+                    ReedlineOption::Edit(EditCommand::CopyLeftBefore(left)),
+                    ReedlineOption::Edit(EditCommand::CopyRightBefore(*right)),
+                ]
+            }
```
sholderbach, line 262: As those are guranteed to be broken for all cases where you are not on the beginning of the text object (putting the left half in the clipboard and then overwriting with the right half) I would omit them until we have a better solution. Yes to shipping but let's not ship broken stuff.

src/edit_mode/vi/parser.rs
```diff
@@ -108,13 +108,23 @@ impl ParsedViSequence {
...
+            (Some(Command::Delete), ParseResult::Incomplete)
+            | (Some(Command::DeleteChar), ParseResult::Incomplete)
+            | (Some(Command::DeleteToEnd), ParseResult::Incomplete)
+            | (Some(Command::Delete), ParseResult::Valid(_))
+            | (Some(Command::DeleteChar), ParseResult::Valid(_))
+            | (Some(Command::DeleteToEnd), ParseResult::Valid(_))
+            | (Some(Command::Yank), ParseResult::Valid(_))
+            | (Some(Command::Yank), ParseResult::Incomplete)
+            | (Some(Command::YankInside(_)), ParseResult::Valid(_))
+            | (Some(Command::YankInside(_)), ParseResult::Incomplete) => Some(ViMode::Normal),
```
sholderbach, line 127: Let's not forget to rebase/merge onto the updated main branch when landing #867

src/core_editor/editor.rs
```diff
@@ -648,6 +698,141 @@ impl Editor {
...
+
+    pub(crate) fn copy_from_start(&mut self) {
+        let insertion_offset = self.line_buffer.insertion_point();
+        if insertion_offset > 0 {
+            self.cut_buffer.set(
+                &self.line_buffer.get_buffer()[..insertion_offset],
+                ClipboardMode::Normal,
+            );
+        }
+    }
```
sholderbach, line 710: I haven't gone through all of those to check if their indexing is correct. That's an annoying part to deal with this level of duplication for the different versions referring to the same text-objects/motions.

src/enums.rs
```diff
@@ -267,6 +267,57 @@ pub enum EditCommand {
     /// Paste content from local buffer at the current cursor position
     Paste,
 
+    /// Copy from the start of the buffer to the insertion point
+    CopyFromStart,
+
+    /// Copy from the start of the current line to the insertion point
+    CopyFromLineStart,
```
sholderbach, line 274: For the users trying to configure their emacs mode having all those extra variants certainly adds some potential for confusion, and when we remove them for a simpler VI mode more breaking changes where we hope people don't try to depend on them.

src/edit_mode/vi/command.rs
deephbz: This was indeed my mistake initially. Thanks for helping review! This specific issue should be addressed in #874

deephbz:
> Thanks for giving this a shot and putting in the work! One thing that makes me hesitant about this, that the copying we do does not work with multipliers at all as each `EditCommand` will update the clipboard. Also operations you define in terms of two `EditCommand`s suffer the same unfortunate fate.
> 
> One solution to that would be to do extra state management around clipboard but keep the old edit commands (suggested in #631 ) or otherwise rethink how we implement the vi mode interface to use an internal selection build up from the motion before performing the action. The latter would require way fewer additional `EditCommand` to be maintained, like the ones you propose here. Related issues where that came up was the impl of #849 where there are cases where the chaining of EditCommands did not achieve the expected result.

Indeed I see a common pattern Move/Cut/Copy + selection range (word, big word, line end, etc.). 3+ actions plus ~10 selections give us numerous `EditCommand`s.

132ikl:
is anything blocking this? I think we're all in agreement that the vi mode interface should be reworked eventually but this PR seems like a helpful stop-gap for now

deephbz:
> is anything blocking this? I think we're all in agreement that the vi mode interface should be reworked eventually but this PR seems like a helpful stop-gap for now

I don't think there's any block except for the dependent PRs https://github.com/nushell/reedline/pull/867 , https://github.com/nushell/reedline/pull/874. 
I've personally used this branch for months on a daily basis without any issue and the Vi mode works perfectly. I'm open to owners modify my above PRs if that help get things merged.

sholderbach:
Should be good for a rebase/merge following the merge of #867.

Is the refactor proposed/documented in #874 already part of this branch or am I misreading this?

Otherwise I am good with landing this (@132ikl if you want to look into something feel free to chime in otherwise I would give the green light after resolving the branch)

deephbz:
> Should be good for a rebase/merge following the merge of #867.
> 
> Is the refactor proposed/documented in #874 already part of this branch or am I misreading this?
> 
> Otherwise I am good with landing this (@132ikl if you want to look into something feel free to chime in otherwise I would give the green light after resolving the branch)

I've rebased this to be on top of #874. Current PR includes all of #874 with one additional commit for Vi-Yank. Once 874 is merged, we can merge this PR with a single-clean commit for yank.

132ikl:
sounds good, let me know when you update the branch

deephbz:
> sounds good, let me or @sholderbach know when you update the branch

Done ready for review :D

132ikl:
thank you!!

PR #869, opened 2025-01-02
Remove questionable `unsafe impl Send`

sholderbach:
`HistoryCompleter` is not `Send` due to it containing a reference to a `&dyn History` which was only `trait History: Send`.
Here the property of a type T needing to implement `Sync` for its `&T` to be `Send` comes into play. Thus this `unsafe impl Send` is unsound.

The change of making `trait History: Send + Sync` makes `HistoryCompleter: Send` automatically and should be compatible for all sound implementations.


kronberger-droid:
@sholderbach sqlite backed history does not compile.
Without a plan to mitigate this no way we can merge this.

Closed in favor of: #1092 would love for you to take a look.

PR #877, opened 2025-01-31
Add Vi `o` command to swap anchor and cursor

thomasschafer:
This PR adds a keybinding to `o` in Vi visual mode, which swaps the cursor and the anchor - see [here](https://vimdoc.sourceforge.net/htmldoc/visual.html#visual-change) for Vim docs explaining this behaviour.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -1127,6 +1127,31 @@ mod test {
         editor.run_edit_command(&EditCommand::Undo);
         assert_eq!(editor.get_buffer(), "This \r\n is a test");
     }
+
+    #[test]
+    fn test_swap_cursor_and_anchor() {
+        let mut editor = editor_with("This is some test content");
+        editor.line_buffer.set_insertion_point(0);
+        editor.update_selection_anchor(true);
+
+        for _ in 0..3 {
+            editor.run_edit_command(&EditCommand::MoveRight { select: true });
+        }
+        assert_eq!(editor.selection_anchor, Some(0));
+        assert_eq!(editor.insertion_point(), 3);
+        assert_eq!(editor.get_selection(), Some((0, 4)));
+
+        editor.run_edit_command(&EditCommand::SwapCursorAndAnchor);
+        assert_eq!(editor.selection_anchor, Some(3));
+        assert_eq!(editor.insertion_point(), 0);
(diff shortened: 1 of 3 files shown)


src/core_editor/editor.rs
```diff
@@ -564,11 +573,12 @@ impl Editor {
...
     pub fn get_selection(&self) -> Option<(usize, usize)> {
         self.selection_anchor.map(|selection_anchor| {
-            if self.insertion_point() > selection_anchor {
+            let (start, end) = if self.insertion_point() > selection_anchor {
                 (selection_anchor, self.insertion_point())
             } else {
                 (self.insertion_point(), selection_anchor)
-            }
+            };
+            (start, end + 1)
```
sholderbach, line 581: This `+ 1` is bad as we operate on graphemes (so potentially multiple potentially multi-byte UTF-8 encoded code points) instead of only teletype era ASCII with 1 byte per atomic character. I think the change to `get_selection` is not needed anymore as this was covered in #867

sholderbach (commented):
I never stop learning new things about vim :) Didn't know about this Visual mode feature.

The feature part looks good to me. So the only thing that needs rework is the part extending the selection by a "character", which should now be covered by the changes in #867. Then we should be good to land.

Thanks for the tests as well!

src/core_editor/editor.rs
thomasschafer: Ah good point - I've just removed those changes

sholderbach: Thanks!

src/edit_mode/vi/command.rs
```diff
@@ -113,6 +113,10 @@ where
             let _ = input.next();
             Some(Command::RepeatLastAction)
         }
+        Some('o') => {
+            let _ = input.next();
+            Some(Command::SwapCursorAndAnchor)
+        }
```
sholderbach, line 119: I think in theory this dispatch from the parser is in sufficient as it would compete with the normal mode `o` (open insert mode in the line below) and we don't discriminate between the modes here.

But as we don't have that `o` implemented yet (as far as i can tell) this should be fine.

sholderbach (approved):
Looks good and works for me!

Also thanks for adding tests here.

PR #878, opened 2025-02-01
Clear selection when exiting visual mode

thomasschafer:
After exiting visual mode with Escape, the selection is currently not cleared until another action is performed such as moving the cursor left or right. This PR fixes that, clearing the selection as soon as Escape is pressed to match the behaviour of Vim.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -202,12 +202,6 @@ impl Editor {
             None
         };
     }
-
-    /// Sets the selection anchor to None, clearing the selection
-    pub fn clear_selection_anchor(&mut self) {
-        self.selection_anchor = None;
-    }
-
     fn move_to_position(&mut self, position: usize, select: bool) {
         self.update_selection_anchor(select);
         self.line_buffer.set_insertion_point(position)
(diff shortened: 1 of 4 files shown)


132ikl:
mentioned this in the Discord, but wanted to add it here also that #867 also implements this, so assuming we get that merged in a timely fashion we can close this PR in favor of #867. otherwise, we can merge this before next Nushell release.

thomasschafer:
> Ah making that part of the `Esc` handler itself is elegant. Should also transfer to the emacs style bindings.
> 
> If you want to revisit that part of #867 feel free to update the changes from it, otherwise this is closed by it.

I've just updated this PR in light of the changes in #867, let me know if this is what you were thinking!

sholderbach (commented):
Sounds good to me. @132ikl or @deephbz do you have a use case where an explicit event for that would be helpful?

132ikl:
hmm, i can't think of anything. it seems like it _could_ be useful (maybe if you were doing a weird macro-y keybinding?), but honestly we could probably just revert this if anyone really wanted this available

PR #881, opened 2025-02-08
Moved partial completion menu next into new menu event

uek-1:
Addresses nushell/nushell#14152 (more context can also be found in #828 ). Creates a separate event, MenuNextComplete, that not only selects the next element but also attempts to partially complete the selection. Reverts MenuNext to simply advancing the next element and quick completing if possible. 

The nushell PR to use this new event can be found at nushell/nushell#15063

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -918,6 +918,7 @@ impl Reedline {
             | ReedlineEvent::OpenEditor
             | ReedlineEvent::Menu(_)
             | ReedlineEvent::MenuNext
+            | ReedlineEvent::MenuNextComplete
             | ReedlineEvent::MenuPrevious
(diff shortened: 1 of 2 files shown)


132ikl:
is there a use case for sending MenuNext without completing? not sure if we should have a new event for this, or just change the behavior of MenuNext. but i'm not super familiar with the completion stuff. (cc @blindFS, @ysthakur)

blindFS:
@132ikl I'd like this because when the only valid options are say `foo` and `foooo`, a typical bash behavior of `f#tab` would complete the repl text to `foo` without moving towards `foooo`

kronberger-droid:
superseded by #967

PR #882, opened 2025-02-11
feat: immediately execute command

mrdgo:
# Description

Pass a boolean value `immediately_execute` to each invocation of `get_line`.
This way, the nu command `commandline edit` can accept a flag `--accept` and not only edit the currently, but also immediately execute the result.

# API

The API severely changes, as now this flag has to be passed to `get_line` and `get_line_helper`.
This could be changed to a channel, but then the initialization of the module would get more complicated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ let mut line_editor = Reedline::create();
 let prompt = DefaultPrompt::default();
 
 loop {
-    let sig = line_editor.read_line(&prompt);
+    let sig = line_editor.read_line(&prompt, false);
     match sig {
         Ok(Signal::Success(buffer)) => {
             println!("We processed: {}", buffer);
--- a/examples/cwd_aware_hinter.rs
+++ b/examples/cwd_aware_hinter.rs
@@ -71,7 +71,7 @@ fn main() -> io::Result<()> {
         } else {
             std::env::set_current_dir(&home_dir).unwrap();
         }
-        let sig = line_editor.read_line(&prompt)?;
+        let sig = line_editor.read_line(&prompt, false)?;
         match sig {
             Signal::Success(buffer) => {
                 println!("We processed: {buffer}");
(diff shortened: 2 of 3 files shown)


src/engine.rs
```diff
@@ -734,30 +738,32 @@ impl Reedline {
...
+
+                // If we believe there's text pasting or resizing going on, batch
+                // more events at the cost of a slight delay.
+                if events.len() > EVENTS_THRESHOLD
+                    || events.iter().any(|e| matches!(e, Event::Resize(_, _)))
+                {
+                    while !completed(&events) && event::poll(POLL_WAIT)? {
+                        events.push(crossterm::event::read()?);
+                    }
+                }
```
132ikl, line 766: could you explain the purpose of these changes? it seems like only the `if immediately_execute` below all this should be necessary to accomplish the change, but not sure what you had in mind.

mrdgo: Thanks for the review! Now that I see how github displays the change, I am confused as well. All of this is indentation, except for the `if !immediately_execute`. I think that github confuses the two instances of `events.push(crossterm::event::read()?);`.

mrdgo: The point is that I don't want reedline to read from stdin when it is in `immediately_accept` mode. So I skip all these conditional/looped input reads and further down add the `Submit` to the queue.

mrdgo:
I'm so sorry for missing some of the calls. Now, clippy should pass

VuiMuich:
Is there anything that could be helped to get this merged?

mrdgo:
Ping maintainers?

VuiMuich:
Well, when I am new to a repo I try to be careful in my interaction as I don't want to annoy the maintainers.
If I am not mistaken you are a first contributor to this project yourself, so I understand why you might be hesitant to drop a ping as well.

Btw: part reason, why I would like to see this getting into the next release is that I recently learned about the `brush-shell` project, so there is two very interesting shells depending on this.

PR #887, opened 2025-03-13
fix: Fuzzy completion with Unicode chars

ysthakur:
This PR fixes a panic that happens when trying to display fuzzy-matched suggestions. Previously, #886 fixed this for prefix matching. Closes https://github.com/nushell/nushell/issues/12680 and https://github.com/nushell/nushell/issues/15302.

This is temporary, until we start highlighting fuzzy-matched suggestions properly (underlining only the graphemes that were matched, rather than underlining a prefix of the suggestion).

I chose to operate on graphemes rather than characters because that's what Nucleo operates on and because they seem to correspond better to a block of text shown on the screen. First we get the width of the shortest string, and then for every suggestion, we highlight the suggestion up to the last grapheme that matches or exceeds this width.

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -1,4 +1,4 @@
-use super::{Menu, MenuBuilder, MenuEvent, MenuSettings};
+use super::{menu_functions::split_suggestion, Menu, MenuBuilder, MenuEvent, MenuSettings};
 use crate::{
     core_editor::Editor,
     menu_functions::{can_partially_complete, completer_input, replace_in_buffer},
@@ -38,8 +38,8 @@ struct ColumnDetails {
     pub columns: u16,
     /// Column width
     pub col_width: usize,
-    /// The shortest of the strings, which the suggestions are based on
-    pub shortest_base_string: String,
+    /// The display width of the shortest string, which the suggestions are based on
+    pub shortest_base_string: usize,
 }
 
 /// Menu to present suggestions in a columnar fashion
@@ -299,23 +299,8 @@ impl ColumnarMenu {
         use_ansi_coloring: bool,
     ) -> String {
         if use_ansi_coloring {
-            // strip quotes
-            let is_quote = |c: char| "`'\"".contains(c);
-            let shortest_base = &self.working_details.shortest_base_string;
-            let shortest_base = shortest_base
-                .strip_prefix(is_quote)
-                .unwrap_or(shortest_base);
-            let match_len = shortest_base.len();
-
-            // Split string so the match text can be styled
-            let skip_len = suggestion
-                .value
-                .chars()
-                .take_while(|c| is_quote(*c))
-                .count();
-            let (match_str, remaining_str) = suggestion
-                .value
-                .split_at((match_len + skip_len).min(suggestion.value.len()));
+            let match_width = self.working_details.shortest_base_string;
+            let (match_str, remaining_str) = split_suggestion(&suggestion.value, match_width);
 
             let suggestion_style_prefix = suggestion
                 .style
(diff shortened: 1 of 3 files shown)


src/menu/columnar_menu.rs
```diff
@@ -38,8 +38,8 @@ struct ColumnDetails {
     pub columns: u16,
     /// Column width
     pub col_width: usize,
-    /// The shortest of the strings, which the suggestions are based on
-    pub shortest_base_string: String,
+    /// The display width of the shortest string, which the suggestions are based on
+    pub shortest_base_string: usize,
```
blindFS, line 42: The naming is a little bit confusing. I think we're safe to change it in this private struct?

ysthakur:
Temporarily converting this to a draft because I don't think this checks for suggestions shorter than the text being completed. Reported here: https://github.com/reubeno/brush/issues/406.

ysthakur:
Just threw in a test to make sure that suggestions shorter than the text being matched don't cause a panic. Should be ready to go now.

blindFS:
I'd like to bring another (probably) related issue https://github.com/nushell/nushell/issues/4595 into the sight here. @uek-1 would know it better.

ysthakur:
> I'd like to bring another (probably) related issue [nushell/nushell#4595](https://github.com/nushell/nushell/issues/4595) into the sight here. @uek-1 would know it better.

@blindFS I'm not sure I understand. Isn't handling quotes in partial completions a related but separate matter?

blindFS:
> > I'd like to bring another (probably) related issue [nushell/nushell#4595](https://github.com/nushell/nushell/issues/4595) into the sight here. @uek-1 would know it better.
> 
> 
> 
> @blindFS I'm not sure I understand. Isn't handling quotes in partial completions a related but separate matter?

Yes, related but separate, just in case we need some top-down redesign to address them at the same time.

src/menu/menu_functions.rs
```diff
@@ -354,6 +357,27 @@ pub fn can_partially_complete(values: &[Suggestion], editor: &mut Editor) -> boo
...
+    for (offset, grapheme) in sugg.grapheme_indices(true) {
+        if curr_width >= match_width {
+            match_end = offset;
+            break;
+        }
+        // Strip quotes from the beginning
+        if offset == 0 && (grapheme == "`" || grapheme == "'" || grapheme == "\"") {
+            continue;
+        }
+        curr_width += grapheme.width();
```
sholderbach, line 375: I guess with this code doing operations on Unicode width would not panic.

But I have a question if this misbehaves if the user tries to complete starting from `"` etc. or `"` being a particular suggestion? (Is this relying on a particular behavior of our `Completer` implementations to take out all quotes?)

sholderbach (commented):
Thanks for tackling this complexity.

I don't quite follow why this code has to use the vague oracle of `UnicodeWidth` and can't simply use the grapheme/byte indices etc. which are the relevant unit for where we actually perform the highlighting/string splitting etc.

src/menu/menu_functions.rs
ysthakur: I was planning on allowing adding a separate display value on `Suggestion`s, so that when showing suggestions for files, the displayed values could be unquoted and unescaped. Then, Reedline would no longer have to strip quotes. This current quote stripping is really for Nushell, and I assume it won't hurt other users of Reedline either.

ysthakur:
> I don't quite follow why this code has to use the vague oracle of `UnicodeWidth` and can't simply use the grapheme/byte indices etc. which are the relevant unit for where we actually perform the highlighting/string splitting etc.

I just used width so that when displaying, the underlined sections would be (around) the same width. But yeah, looking at graphemes makes more sense, I've updated it now

Issue #885, opened 2025-03-09
Perf issue: scrolling in the completion menu keeps triggering (re)parsing of the command line over and over


blindFS:
**Platform** macOS
**Terminal software** ghostty

Found this issue during https://github.com/nushell/nushell/pull/15270

Basically the menu next/prev event will trigger the parsing of the cli, which involves dozens of `parse_module_file` calls.

I guess it has something to do with the highlighting. But since the cli text doesn't change by those events (maybe I missed some config here to make it happen), I think it should be optimized.

## Steps to reproduce

1. type in `use std <tab>`
2. hold ctrl-p/ctrl-n down for a while

rustc optimized nu release performs noticeably better.

Typing in cli after `use std ` is also not very responsive due to the same reason, but that is probably harder to fix.

fdncred:
Nice catch. Would be wonderful to figure out how to fix this.

blindFS:
Ideally, highlighting better to be an async function and then we can set a debounce timeout for it.

I'm not sure if it worth the effort, since the only remaining situation that I can notice the lagging (but I only tested it on M4 pro chip) is `use std <typing/left/right>`. And that one seems to me a nu_parser issue.

PR #888, opened 2025-03-15
feat: events that only repaint the menu

blindFS:
For menu events of #885 .

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -801,6 +801,25 @@ impl Reedline {
                         }
                         return Ok(signal);
                     }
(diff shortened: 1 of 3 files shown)


blindFS:
Need more test in different environments

blindFS:
Moving cursor around (left/right mostly) also triggers the repainting, prompt + parsing + highlighting.
I think that could also be optimized.

ysthakur:
@blindFS Is this ready to review?

blindFS:
> @blindFS Is this ready to review?

It's too hacky and of low priority. I'd prefer an async solution for all events anyways.

PR #891, opened 2025-03-19
Add smooth scrolling to menus

### LINKED ISSUE #733: smoother menu scrolling
scrolling through menus currently looks like this

https://github.com/nushell/reedline/assets/104733404/0a60ef2f-c042-49d8-9a68-ea6d9561390f

notice how when scrolling up, the selection is stuck at the bottom.
This could be improved by not making the selection stuck at the bottom when scrolling up.

maxomatic458:
currently when scrolling up, the selected item will always stick to the bottom (see #733)

This PR fixes that.

Currently implemented for
* ide menu
* columnar menu

closes #733

### DIFF
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -812,9 +812,15 @@ impl Menu for IdeMenu {
             self.working_details.space_left = space_left;
             self.working_details.space_right = space_right;
 
-            let available_lines = painter
-                .remaining_lines()
+            let mut available_lines = painter
+                .remaining_lines_real()
                 .min(self.default_details.max_completion_height);
+
+            // Handle the case where a prompt uses the entire screen.
+            // Drawing the menu has priority over the drawing the prompt.
+            if available_lines == 0 {
+                available_lines = painter.remaining_lines().min(self.min_rows());
+            }
 
             let visible_items = available_lines.saturating_sub(border_width);
 
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -811,8 +811,11 @@ impl Menu for IdeMenu {
             self.working_details.space_left = space_left;
             self.working_details.space_right = space_right;
 
-            let available_lines = painter.remaining_lines().min(self.default_details.max_completion_height)
-                .saturating_sub(1); // Not sure why this is 1 less than the `available_lines` from [`Menu::menu_string`] 
+            let available_lines = painter
+                .remaining_lines()
+                .min(self.default_details.max_completion_height)
+                .saturating_sub(1); // Not sure why this is 1 less than the `available_lines` from [`Menu::menu_string`]
+            
             let visible_items = available_lines.saturating_sub(border_width);
 
             self.skip_values = if self.selected < self.skip_values {
@@ -857,7 +860,6 @@ impl Menu for IdeMenu {
                 0
             };
 
-            
             let available_lines = available_lines.min(self.default_details.max_completion_height);
             let skip_values = self.skip_values as usize;
 
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -818,7 +818,7 @@ impl Menu for IdeMenu {
 
             let visible_items = available_lines.saturating_sub(border_width);
 
-            self.skip_values = if self.selected < self.skip_values {
+            self.skip_values = if self.selected <= self.skip_values {
                 // Selection is above the visible area, scroll up to make it visible
                 self.selected
             } else if self.selected >= self.skip_values + visible_items {
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -814,8 +814,8 @@ impl Menu for IdeMenu {
 
             let available_lines = painter
                 .remaining_lines()
-                .min(self.default_details.max_completion_height)
-                .saturating_sub(1);
+                .saturating_sub(1)
+                .min(self.default_details.max_completion_height);
 
             let visible_items = available_lines.saturating_sub(border_width);
 
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -814,7 +814,6 @@ impl Menu for IdeMenu {
 
             let available_lines = painter
                 .remaining_lines()
-                .saturating_sub(1)
                 .min(self.default_details.max_completion_height);
 
             let visible_items = available_lines.saturating_sub(border_width);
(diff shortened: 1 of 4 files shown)


maxomatic458:
Here is a video of the new behavior, see issue above to see the current behavior

https://github.com/user-attachments/assets/f54a7993-9919-425e-97d7-2e5db7126ddf

src/menu/ide_menu.rs
```diff
@@ -807,6 +811,24 @@ impl Menu for IdeMenu {
 
             self.working_details.space_left = space_left;
             self.working_details.space_right = space_right;
+
+            let available_lines = painter
+                .remaining_lines()
+                .min(self.default_details.max_completion_height)
+                .saturating_sub(1);
```
maxomatic458, line 818: for some reason this is always 1 higher than `available_lines` from the `menu_string` function, not sure why that is the case as both should be the lines (not including the prompt).

Thats why im subtracting 1 here.

ysthakur:
I'm really sorry, I'm having a hard time understanding what exactly this PR does from looking at the video, and when I run it myself (VS Code terminal, Windows Terminal + Zellij, Windows Terminal without Zellij), I don't see any change in behavior from the main branch. Is this supposed to keep the selected item at the top when scrolling?

maxomatic458:
@ysthakur see the video linked in the original issue, basically currently when you scroll up, the selected item will be stuck at the bottom.
This PR fixes that.

src/menu/ide_menu.rs
ysthakur: It seems like the prompt is indeed being included in `painter.remaining_lines()`. If I resize my terminal so that the prompt takes up 2 or 3 lines, `painter.remaining_lines()` seems to be 2 or 3 more lines too big.

ysthakur: Here's a recording demonstrating that: https://asciinema.org/a/j8bRJs2t2RrIvUGlIFrXNav20. Notice how the suggestion `foo9` isn't visible when selected

maxomatic458: ah ok i see

looks like``Painter::remaining_lines`` does not consider the actual prompt (only the location of the prompt, e.g if its not at the top of the terminal)

maxomatic458: should be fixed with the latest commit but ill do some more testing tomorrow as i did some changes in the Painter

src/painting/painter.rs
```diff
@@ -123,7 +126,7 @@ impl Painter {
 
     /// Returns the available lines from the prompt down
     pub fn remaining_lines(&self) -> u16 {
-        self.screen_height().saturating_sub(self.prompt_start_row)
+        self.screen_height().saturating_sub(self.prompt_height)
```
ysthakur, line 129: Hmm, should this also subtract the start row, to account for the prompt not being at the top of the terminal?
```suggestion
        self.screen_height().saturating_sub(self.prompt_start_row).saturating_sub(self.prompt_height)
```

maxomatic458:
ok i believe the PR should be ready now.
i reverted the change in ``prompt_lines_with_wrap`` as the `-1` seems to be intentional and i dont want to break stuff anywhere else.

I tested it with multiline prompts and with prompt wrapping (due to terminal size)

The CI is failing because of the new rust lints added with 1.86, it would probably be better to fix them in another PR (although i can also fix them here if you want to)

ysthakur:
@sholderbach Could you take a look at this? I don't know Reedline well enough to say if the change to say if the `remaining_lines` change is fine.

---

> The CI is failing because of the new rust lints added with 1.86, it would probably be better to fix them in another PR (although i can also fix them here if you want to)

Yeah that's fine, no need to do it in this PR.

---

With the change to `Painter::remaining_lines`, this is what the description menu looks like when I open up the help menu and scroll all the way to the bottom of the description for `cd`:
image

This is how it looks currently without your change:
image

So the new behavior actually looks more reasonable to me, since the prompt is visible. We do need to throw in an ANSI reset or something to fix that color, though.

---

The only other place using `remaining_lines()` is `Painter::repaint_buffer()`:
https://github.com/nushell/reedline/blob/cc7960523c43dc50a383d66e745e3536afe760c7/src/painting/painter.rs#L230-L231

Here, it hasn't drawn the prompt yet, so it doesn't want you to subtract the prompt height from the remaining lines.

maxomatic458:
> So the new behavior actually looks more reasonable to me, since the prompt is visible. We do need to throw in an ANSI reset or something to fix that color, though.

Ok so the prompt should now always be visible whenever a menu is open (so any content that is not fitting will get cut of at the end).

> So the new behavior actually looks more reasonable to me, since the prompt is visible. We do need to throw in an ANSI reset or something to fix that color, though.

For me the colors are still the same (maybe check if that still happens with the latest commit)

> Here, it hasn't drawn the prompt yet, so it doesn't want you to subtract the prompt height from the remaining lines.

Fixed that by re-adding the `prompt_height`

ysthakur:
Hmm, I'm still getting those colors on the latest commit, along with the prompt not always being visible:
image

It does seem unlikely to me that someone would be activating a description menu with such a small terminal, but I don't know how people use Reedline (apart from Nushell itself).

Apart from that, I see no problems with this PR. It works well for the columnar and ide menus. I'd just like Stefan to take a look to confirm that this is fine.

src/painting/painter.rs
```diff
@@ -209,7 +217,7 @@ impl Painter {
         }
 
         // Lines and distance parameters
-        let remaining_lines = self.remaining_lines();
+        let remaining_lines = self.remaining_lines() + self.prompt_height;
```
ysthakur, line 220: Edge case: not enough space to even draw the prompt fully, in which case `remaining_lines` should be 0, but will be positive with this change. Might be safer to use `self.screen_height().saturating_sub(self.prompt_start_row)` directly

maxomatic458:
PR currently breaks menu rendering with multiline prompts.

1: Current behavior -> menu has priority over prompt (prompt will be partially rendered if required)
2: This PR -> prompt has priority over menu

realistically people will probably at most use 2-3 line prompts, so option 2 is probably better for most cases
however with Large prompts (where a prompt is larger than the terminal) menus will not correctly render (or not at all)

but i believe i will try change it so we stick with the current behavior.

ysthakur (approved):
Looks good to me with the latest change, making a separate function should avoid issues. I was a little worried about `lines.prompt_lines_with_wrap(screen_width) + 1` because `prompt_lines_with_wrap` doesn't always subtract 1 from its result, but I don't think it'll actually be a problem. Sorry for the long wait.

We really need to improve our behavior for small terminals in the future.

PR #894, opened 2025-03-24
fix: prompt duplication when resizing with non-empty right_prompt

### LINKED ISSUE #864: Right prompt redraw issue in nu 0.101
**Platform** macOS
**Terminal software** ghostty

Noticed that right prompt redraw happens with resize in nu 0.101.0
But there's a side effect as shown in the screenshot

## Steps to reproduce
1. set some right prompt
2. multiline left prompt
3. resize horizontally

## Screenshots/Screencaptures

https://github.com/user-attachments/assets/dde62383-9b35-42d0-90c4-b1f3b5306d76

possibly related to #857

blindFS:
Supposed to fix #864 , this method works well on my mac with ghostty/wezterm/terminal, but I need help to test on other terminal/os.

After:

https://github.com/user-attachments/assets/e1198bed-49db-4003-a388-22e9e2f8290f

The prompt start position goes up with the line-wrapped text, which is the same behavior as zsh.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -86,6 +86,10 @@ fn select_prompt_row(
     PromptRowSelector::MakeNewPrompt { new_row }
 }
 
+struct ResizeResult {
+    width_decreased: bool,
+}
+
 /// Implementation of the output to the terminal
 pub struct Painter {
     // Stdout
@@ -94,7 +98,7 @@ pub struct Painter {
     terminal_size: (u16, u16),
     last_required_lines: u16,
     large_buffer: bool,
-    just_resized: bool,
+    just_resized: Option<ResizeResult>,
     after_cursor_lines: Option<String>,
 }
 
@@ -106,7 +110,7 @@ impl Painter {
             terminal_size: (0, 0),
             last_required_lines: 0,
             large_buffer: false,
-            just_resized: false,
+            just_resized: None,
             after_cursor_lines: None,
         }
     }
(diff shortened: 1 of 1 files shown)


fdncred:
I tested it on macos. I don't think it's working right. I mean the right prompt drawing isn't bad but it just goes up the screen erasing the terminal one line at a time as it goes up.
resize

blindFS:
Oops, what terminal are you using in that?

fdncred:
> Oops, what terminal are you using in that?

That's ghostty on macos

blindFS:
> > Oops, what terminal are you using in that?
> 
> That's ghostty on macos

Hmmm, interesting. That means my duplication experience could be caused by bad configuration, OMG.
Thanks for your effort, I'll take more look at it.

blindFS:
@fdncred , according to my test, the discrepancy is caused by whether the left prompt string is multiline.

Would you please try with `let $env.PROMPT_COMMAND = "foo\nbar"` to see if it behaves differently?

blindFS:
I've found out the minimal steps required to reproduce the bug with ghostty on macOS and the default nu config:

1. multiline left prompt and non-empty right prompt (the default one is fine), `$env.PROMPT_COMMAND = {|| "foo\nbar"}`
2. turn off `osc133`, either `$env.config.shell_integration.osc133 = false`, or open a tmux session in ghostty.

So it's definitely more complicated than I thought.

fdncred:
I was just running the reedline demo like this `cargo run --example demo --features=sqlite,bashisms`

blindFS:
> I was just running the reedline demo like this `cargo run --example demo --features=sqlite,bashisms`

Oh, I mean the steps to reproduce #864. 

There are many factors that can determine how long-line text gets wrapped in a terminal. 

Due to my lack of knowledge about the terminal and OSC stuff, I can't find a proper way to locate the prompt start position given all those complications vary from app to app. So I'm going to close this PR for now. 

However, the method used in current reedline implementation (find cursor first, then rewind back by counting `\n` in the text) certainly behaves worse compared to other shells like zsh/bash. I'll create another issue to elaborate that problem.

PR #895, opened 2025-03-24
Upgrade `unicode-width` to 0.2

KAAtheWiseGit:
`unicode-width` has to bundle unicode codepoint tables, so it's quite large (1.5MB).  They've recently released the version 0.2, which caused a split in Nushell dependencies between 0.1 and 0.2, doubling the size.

The main breaking change for `unicode-width` is that `\n` is now counted as having width 1 and not 0.  But that shouldn't matter for reedline, since the menus which use it (columnar/IDE/list) work on lines.


sholderbach:
Sounds generally good to me, I remember that the changed semantics caused some serious churn for the `tabled` layouting.

PR #841, opened 2024-10-06
Attempt to fix multiline prompt resize issue

blindFS:
Tried to fix #809 in this PR.

The problem lies in the handle_resize function https://github.com/nushell/reedline/blob/4634f71705785a772c894aa45e30db88f3c7a8d5/src/painting/painter.rs#L475-L492 where the new position is set according to cursor, while the cursor's row is larger than the prompt start row in multiline cases.

PR #898, opened 2025-03-26
fix: prompt glitch when resizing with cursor in a multiline command

blindFS:
Fixes a bug overlooked in #841 (forgot that the `before_cursor` string could have newlines in it too).

Before:

https://github.com/user-attachments/assets/cf79d8db-24c8-45f0-9a6d-f0092daaa7eb

After: stays in the same row.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -203,7 +203,8 @@ impl Painter {
         if self.just_resized {
             self.prompt_start_row = self.prompt_start_row.saturating_sub(
                 (lines.prompt_str_left.matches('\n').count()
-                    + lines.prompt_indicator.matches('\n').count()) as u16,
+                    + lines.prompt_indicator.matches('\n').count()
+                    + lines.before_cursor.matches('\n').count()) as u16,
             );
             self.just_resized = false;
         }


fdncred:
Thanks

src/painting/painter.rs
```diff
@@ -215,19 +215,19 @@ impl Painter {
...
+            self.prompt_start_row = self
+                .prompt_start_row
+                .saturating_sub(lines_before_cursor - 1);
             self.just_resized = false;
         }
 
         // Lines and distance parameters
         let remaining_lines = self.remaining_lines();
-        let required_lines = lines.required_lines(screen_width, menu);
+        let required_lines = lines.required_lines(screen_width, false, menu);
```
blindFS, line 230: Line 217, 218, and 230 share quite a few common computations that hold potential for improving responsiveness.

fdncred:
I'm not sure whether it is with this PR or not, but I've seen the resizing Windows Terminal to the right works but resizing to the left eats up one line per resize. Here's a quick gif.
resize-glitch

fdncred:
> Does expanding to the right (by dragging the right boundary) also eat up history?

Yes, and sometimes more than one line at a time. I just had it jump half the width of the screen while expanding from the right side more to the right. Dragging the right boundary towards the left just makes text wrap and doesn't have the same effect.

> Does it have something to do with the right prompt? What if it's empty or $env.config.render_right_prompt_on_last_line = true?

It could. My render_right_prompt_on_last_line is set to false. I use the oh-my.nu script in nu_scripts. This is my prompt configuration in config.nu.
```nushell
# prompt
use modules\prompt\oh-my.nu git_prompt
$env.PROMPT_COMMAND = {|| (git_prompt).left_prompt }
$env.PROMPT_COMMAND_RIGHT = {|| (git_prompt).right_prompt }
$env.PROMPT_INDICATOR = {|| if ($env.LAST_EXIT_CODE == 0) { $"(ansi green_bold)\n❯ (ansi reset)" } else { $"(ansi red_bold)\n❯ (ansi reset)" } }
$env.TRANSIENT_PROMPT_COMMAND = ""
$env.TRANSIENT_PROMPT_COMMAND_RIGHT = {|| (git_prompt).right_prompt }
$env.TRANSIENT_PROMPT_INDICATOR = {|| if ($env.LAST_EXIT_CODE == 0) { $"(ansi light_yellow_bold)❯ (ansi reset)" } else { $"(ansi light_red_bold)❯ (ansi reset)" } }
$env.PROMPT_MULTILINE_INDICATOR = {|| $"(ansi -e {fg: '#CB4B16'})(char -u '276f')(ansi -e {fg: '#CACA02'})(char -u '276f')(ansi -e {fg: '#4E9A06'})(char -u '276f')(ansi reset)(char space)" }
```

PR #899, opened 2025-03-27
fix: dislocation of cursor after previous_history navigation to a multiline entry

blindFS:
Fixes this bug:

When I press ctrl-p/up to bring up this kind of command:

image

The cursor is at the end of the first line. However when I press left/right key to move the cursor, it unexpectedly jumps to the last line and executes the movement there.

After pressing left:

image

After this PR, however.

It shows the cursor directly at the end of the last line, just like selecting that entry in a history menu will do.

The same thing happens for vi normal mode.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1311,6 +1311,7 @@ impl Reedline {
         self.editor.move_to_line_end(false);
         self.editor
             .update_undo_state(UndoBehavior::HistoryNavigation);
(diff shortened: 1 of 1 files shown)


fdncred:
This has annoyed me for years. Glad you fixed it.

Bahex:
Is there a reason we didn't move ahead with this?

PR #900, opened 2025-04-01
fix: adjust cursor start position for large buffers

qilme:
nushell/nushell#15436

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -243,7 +243,7 @@ impl Painter {
 
         // Moving the start position of the cursor based on the size of the required lines
         if self.large_buffer || is_reset() {
-            for _ in 0..screen_height - 1 {
+            for _ in 0..screen_height - lines.required_lines(screen_width, None) {
                 self.stdout.queue(Print(&coerce_crlf("\n")))?;
             }
             self.prompt_start_row = 0;
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -243,6 +243,9 @@ impl Painter {
 
         // Moving the start position of the cursor based on the size of the required lines
         if self.large_buffer || is_reset() {
+            for _ in 0..screen_height - 1 {
+                self.stdout.queue(Print(&coerce_crlf("\n")))?;
+            }
             self.prompt_start_row = 0;
         } else if required_lines >= remaining_lines {
             let extra = required_lines.saturating_sub(remaining_lines);


fdncred:
Thanks for trying to make nushell and reedline better. Please help me understand how this fixes the issue. If my screen height is 50 how is queuing 50 `\n`'s going to help? Is there a way we can easily try this in the reedline demo to see if it works?

qilme:
Since reedline uses crossterm and [crossterm](https://docs.rs/crossterm/0.28.1/src/crossterm/cursor.rs.html#60) uses ansi escape sequences to implement the function, I can use nushell to emulate the behavior.
1. `scope commands` or other command judged as large buffer.If it is a lagre buffer, then prompt_start_row is 0 and nushell send "ESC [0;0H" to terminal.
2. `ansi -e "H"` is the same as "ESC [0;0H". It clear my viewpoint in the Windows Terminal, vscode, alacritty and termux.
3. Scrool up and content as high as my screen height disappears.

1. `scope commands`
2. `for x in 1..50 { if $x > 30 { break }; print ""}`
3. `ansi -e "H"`
4. Scrool up and content exists

fdncred:
But don't you have a bunch of empty lines after the buffer?

src/painting/painter.rs
```diff
@@ -243,6 +243,9 @@ impl Painter {
 
         // Moving the start position of the cursor based on the size of the required lines
         if self.large_buffer || is_reset() {
+            for _ in 0..screen_height - 1 {
```
Bahex: This assumes prompt + cmdline height is 1 line, causing multi line prompts and multi line cmdlines trigger the opposite of the bug, duplicating lines instead of erasing them:
image

src/painting/painter.rs
```diff
@@ -243,6 +243,9 @@ impl Painter {
 
         // Moving the start position of the cursor based on the size of the required lines
         if self.large_buffer || is_reset() {
+            for _ in 0..screen_height - lines.required_lines(screen_width, None) {
+                self.stdout.queue(Print(&coerce_crlf("\n")))?;
+            }
```
blindFS, line 248: @qilme @Bahex Sorry to step in, I've noticed a panic caused by this change.

Basically we need `saturating_sub` to avoid negative numbers.
Besides, I'm not sure whether we should take `lines.hint` into account here.

blindFS: I've pushed what I consider to be a proper fix to #898 (since there might be a value useful for both issues), Please review that when you get time.

Bahex: @blindFS sorry for the slow reply, your PR was already merged by the time i got the time to take a look so i just forgot about it.

> Besides, I'm not sure whether we should take `lines.hint` into account here.

getting a multiline hint can similarly push text up, not taking that into account can cause lines to disappear as well.

blindFS: > getting a multiline hint can similarly push text up, not taking that into account can cause lines to disappear as well.

Say we have a large_buffer repaint caused by a long multiline hint, if we add the height of self.hint to `lines.required_lines`, we'll end up not doing any scrolling. Is that expected behavior?

PR #902, opened 2025-04-04
fix clippy: call `next_back` instead of `last` on `DoubleEndedIterator`

WindSoilder:
Fixes clippy(version 0.1.86) error:
```
error: called `Iterator::last` on a `DoubleEndedIterator`; this will needlessly iterate the entire iterator
   --> src/core_editor/line_buffer.rs:474:14
    |
474 |             .last()
    |              ^^^^^^ help: try: `next_back()`
```


sholderbach (approved):
Thanks

PR #904, opened 2025-04-15
chore: bump rust to 1.84.1

schlich:
unblocks work on the keybindings, modalkit was upper bound by an old version of rust


fdncred:
We should only bump this to match the version of nushell, which is 1.84.1. Bumping it to 86 means that it's possible to have some syntax incompatibility that may make it unusable to nushell, which is a non-starter for reedline.

fdncred:
@schlich After talking about this topic in our team meeting today, we decided 3 things.

1. If you're going to try to implement modalkit in reedline, we'd like to bump to the minimum version modalkit requires, which I think is 1.74. My apologies for asking you to bump to the latest nushell version.
2. We also decided that it's best to have that reedline bump inside the PR that implements modalkit instead of having a PR to only bump reedline. If something happens and you're not able to complete the PR(s), we didn't bump reedline unnecessarily. We feel it's safer to the consumers of reedline this way. My apologies for asking you to do it differently.
3. We also need to make sure whatever changes in relation to modalkit and vim keybindings is "easily" translatable to nushell, which is probably reedline's biggest "customer". That probably goes without saying. I'm looking forward to these changes to reedline and nushell. Thanks for your interest.

With that said, I'm going to close this PR and hope you're still onboard with trying to implement these changes.

PR #905, opened 2025-04-15
feat: allow user to Select MatchDirection in Menu

divanshu-go:
* Added a new `MatchDirection` enum to specify the direction for searching matches in suggestions. (`src/menu/mod.rs`)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -277,8 +277,8 @@ pub use validator::{DefaultValidator, ValidationResult, Validator};
 
 mod menu;
 pub use menu::{
-    menu_functions, ColumnarMenu, DescriptionMenu, DescriptionMode, IdeMenu, ListMenu, Menu,
-    MenuBuilder, MenuEvent, MenuTextStyle, ReedlineMenu,
+    menu_functions, ColumnarMenu, DescriptionMenu, DescriptionMode, IdeMenu, ListMenu,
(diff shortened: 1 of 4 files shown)


ysthakur:
Thanks for trying to improve Reedline. I'm not sure this is the right way to go, since searching from the right doesn't fully fix the issue. Suppose you type in `apply` and you have a suggestion for a file named `apply-apply.rs`. If you search from the right, the second `apply` would be highlighted rather than the first, which seems unintuitive.

I looked at the conversation on the linked PR, and this comment seems pretty sensible to me: https://github.com/antinomyhq/forge/pull/650#issuecomment-2791458972
> How about we create our own `CompletionMenu` and use that instead of the default one? This way we won't need to wait for the PR to be merged in nushell.

The menus don't get updated too often, so I don't believe it would be a huge maintenance burden. And it would only be temporary, because https://github.com/nushell/reedline/pull/798 is on the way. If #798 is merged, it *will* require Forge to make changes downstream to take advantage of the new `match_indices` field, but once you do that, you won't need to worry about inaccurate highlighting.

To clarify, I'm not shooting down this PR, I'm still open to merging it, but I'd like to avoid adding a new option to our menus. If #798 gets merged, the match direction stuff would just end up being removed again (even if #798 isn't merged, we'll probably end up making a similar change making the match direction stuff obsolete).

PR #911, opened 2025-05-23
fix: selection in emacs/viins mode

### LINKED ISSUE #893: Line editor selects two characters instead of one
### Describe the bug

I normally use shift + arrow keys to select one character at a time in the line editor. However, now when I start to select text, two characters are always _initially_ selected (one extra to the right of the caret), except when selecting the very last character (in which case it works as expected, since there is no extra character to the right of it). 

This seems to be a regression. Everything works fine on version `0.100.0`, but it is not working correctly on `0.103.0`.

### How to reproduce

1. Write a couple of characters in the line editor, e.g. `123456789`.
2. Position the caret in-between, e.g. after the `3`,  `123|456789`.
3. Use e.g. shift + left arrow and observe that both the character `3` and the character `4` have been selected.

### Expected behavior

I expect a single press of shift + left/right arrow to only select at most one character.

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.103.0                                                           |
| major              | 0                                                                 |
| minor              | 103                                                               |
| patch              | 0                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.85.0 (4d91de4e4 2025-02-17) (built from a source tarball) |
| cargo_version      | cargo 1.85.0 (d73d2caf9 2024-12-31)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | standard                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

blindFS:
fixes #893 

A not-so-elegant solution coded by AI agent, just works according to my test.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -2,7 +2,7 @@ use super::{edit_stack::EditStack, Clipboard, ClipboardMode, LineBuffer};
 #[cfg(feature = "system_clipboard")]
 use crate::core_editor::get_system_clipboard;
 use crate::enums::{EditType, UndoBehavior};
-use crate::{core_editor::get_local_clipboard, EditCommand};
+use crate::{core_editor::get_local_clipboard, EditCommand, PromptEditMode, PromptViMode};
 use std::ops::DerefMut;
 
 /// Stateful editor executing changes to the underlying [`LineBuffer`]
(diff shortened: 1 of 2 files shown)


src/core_editor/editor.rs
```diff
@@ -626,20 +626,35 @@ impl Editor {
     /// If a selection is active returns the selected range, otherwise None.
     /// The range is guaranteed to be ascending.
     pub fn get_selection(&self) -> Option<(usize, usize)> {
+        // Default to Vi-style behavior for backward compatibility
+        self.get_selection_with_mode(&PromptEditMode::Vi(PromptViMode::Normal))
```
sholderbach, line 630: Is there an explicit need to be backwards compatible here?

Maybe the correct thing would be to split it into true selection (which covers the characters in a consistent way with a range using indices used by the edit ops), and visual selection which is mode aware.

It seems the bug prone part was the visual selection, if there are no other users in reedline itself we should consider what the right behavior would be for a public user of `get_selection` (and if that is unlikely we can make things private.

sholderbach (commented):
Thanks for trying to tackle this (and not just doing the revert game) Haven't run it yet so I can't tell if the AI is doing slop here.

If it works I don't mind the use of it but maybe you could go through the tests and check which are truly necessary to cover the relevant bits and be explanatory, but we don't need too much redundant prose in the form of tests.

collinmurch:
https://github.com/nushell/reedline/pull/926

PR #913, opened 2025-06-03
Handle suggestions that are substrings of first suggestion

ysthakur:
Currently, when finding a common prefix for partial completions, we assume that suggestions after the first one won't be substrings of the first suggestion. This PR handles that case.

We did have code handling the case when the first suggestion is a substring of a later suggestion, but it doesn't do case-insensitive comparison, so this PR takes care of that too.

### DIFF
--- a/src/menu/menu_functions.rs
+++ b/src/menu/menu_functions.rs
@@ -585,7 +585,7 @@ mod tests {
     }
 
     #[rstest]
-    #[case(&["nushell", "NULL"], 2)]
+    #[case(&["nushell", "nush", "NULL"], 2)]
     #[case(&["ｎｕｓｈｅｌｌ", "ｎｕｌｌ"], 6)]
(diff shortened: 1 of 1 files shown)


src/menu/menu_functions.rs
```diff
@@ -162,38 +162,23 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult {
 
 /// Finds index for the common string in a list of suggestions
 pub fn find_common_string(values: &[Suggestion]) -> (Option<&Suggestion>, Option<usize>) {
```
ysthakur, line 164: Side note: I would much rather this returned a single `Option<(Suggestion, usize)>` (or actually, an `Option<&str>`/`Option<usize>` would be even better, since callers can just get the first suggestion themselves). Didn't modify it in this PR because even though this helper's only used in one place in Reedline, consumers of Reedline might be using it for their menus.

PR #915, opened 2025-06-05
Protect against invalid suggestion spans

ysthakur:
This PR protects against some panics by suggestion spans whose start or end isn't on a character boundary. It does this by using the highest index greater than or equal to the span start/end that's on a character boundary. There's already a [`floor_char_boundary`](https://doc.rust-lang.org/std/primitive.str.html#method.floor_char_boundary) method on `str` for doing this, but it's nightly-only right now, so I had to add a new public helper to `menu_functions`. This helper will be removed when `str.floor_char_boundary` becomes stable, though.

Indices that go beyond the end of the buffer are replaced with the length of the buffer.

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -1,7 +1,9 @@
 use super::{Menu, MenuBuilder, MenuEvent, MenuSettings};
 use crate::{
     core_editor::Editor,
-    menu_functions::{can_partially_complete, completer_input, replace_in_buffer},
+    menu_functions::{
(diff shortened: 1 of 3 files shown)


Juhan280:
Is there anything blocking this? There seem to be no activity for a month

ysthakur:
@Juhan280 Nothing blocking it, just waiting for a review. We don't have a lot of core team members that are familiar with Reedline, and those that are can be very busy

ysthakur:
Reporting bad spans to the implementer would definitely be good if there's a nice way to do it. In Java land I'd just throw an exception and call it a day, not sure how to do it in Rust.

PR #916, opened 2025-06-06
Only repaint at the end when processing a batch of events

### LINKED ISSUE #914: Performance issue for repeated key presses when using reedline through a slow ssh connection
When using reedline through ssh with a bad internet connection, it takes a very long time to handle a sequence of repeated key presses. This happens for example when holding down an arrow key to navigate through the line of text. Reedline can take several seconds to finish processing the sequence of key presses.

This latency is significantly higher than the latency at which the key presses are actually arriving through the connection, or the latency at which bash will process them. It's also not a matter of the key press rate being too fast for reedline to process, because it handles the same input just fine when it's coming from a ssh connection with good internet.

I've observed this on two different machines, with both an artificially throttled internet connection and a real bad internet connection. I've also observed it both when running nushell and when running reedline on it's own.

new-years-eve:
Reedline will batch KeyEvents when they arrive in quick successions. It then further accumulates edit events into a single edit event. This allows it to only perform a single repaint for the whole batch, instead of always needing one repaint per event. However, as of now, events which aren't edit event, such as arrow key presses, still trigger one repaint per event. This commit will change that, and always repaint once at the end of the batch (if a repaint is needed).

- fixes #914

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -789,6 +789,7 @@ impl Reedline {
             }
 
             // Handle reedline events.
+            let mut need_repaint = false;
             for event in reedline_events {
                 match self.handle_event(prompt, event)? {
                     EventStatus::Exits(signal) => {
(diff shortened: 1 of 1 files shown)


eMPee584:
Does this also fix line duplication (even text disappearance (in `aichat`) under circumstances I've yet to figure out) when resizing terminal?

PR #918, opened 2025-06-11
Fix panic when using fuzzy completion in a folder containing files starting with a multibyte unicode char

dburburan:
This fixes https://github.com/nushell/reedline/issues/919 see the issue for a full description

I see a lot of discussion about newer code coming related to fuzzy matching, so this is just to temporarily fix the immediate issue of the panic. I think the most useful thing here is the addition of the test case to catch it.

I wasn't sure if I should handle the match_position = 0 case specially, or modify match_len to mean characters, not bytes. I went with the latter as I thought it looked cleaner, but there's definitely an argument for keeping the old code path for the majority of cases.

This is my first PR, very open to any feedback. Did my best but there's a lot of guidelines to absorb and I don't know most of the code base.

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -826,4 +826,16 @@ mod tests {
         menu.update_values(&mut editor, &mut completer);
         assert!(menu.menu_string(2, true).contains("おは"));
     }
+
+    #[test]
+    fn test_menu_create_string_starting_with_multibyte_char() {
+        // https://github.com/nushell/nushell/issues/15938
+        let mut completer = FakeCompleter::new(&["验abcdef/", "abcdef/"]);
+        let mut menu = ColumnarMenu::default().with_name("testmenu");
+        let mut editor = Editor::default();
+
+        editor.set_buffer("ac".to_string(), UndoBehavior::CreateUndoPoint);
+        menu.update_values(&mut editor, &mut completer);
+        assert!(menu.menu_string(10, true).contains("验"));
+    }
 }
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -305,7 +305,7 @@ impl ColumnarMenu {
             let shortest_base = shortest_base
                 .strip_prefix(is_quote)
                 .unwrap_or(shortest_base);
-            let match_len = shortest_base.len();
+            let match_len = shortest_base.chars().count();
 
             // Find match position - look for the base string in the suggestion (case-insensitive)
             let match_position = suggestion
@@ -315,8 +315,15 @@ impl ColumnarMenu {
                 .unwrap_or(0);
 
             // The match is just the part that matches the shortest_base
-            let match_str = &suggestion.value[match_position
-                ..match_position + match_len.min(suggestion.value.len() - match_position)];
+            let match_str = {
+                let match_str = &suggestion.value[match_position..];
+                let match_len_bytes = match_str
+                    .char_indices()
+                    .nth(match_len)
+                    .map(|(i, _)| i)
+                    .unwrap_or_else(|| match_str.len());
+                &suggestion.value[match_position..match_position + match_len_bytes]
+            };
 
             // Prefix is everything before the match
             let prefix = &suggestion.value[..match_position];
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -838,4 +838,16 @@ mod tests {
         menu.update_values(&mut editor, &mut completer);
         assert!(menu.menu_string(10, true).contains("验"));
     }
+
+    #[test]
+    fn test_menu_create_string_long_unicode_string() {
+        // https://github.com/nushell/reedline/pull/918
+        let mut completer = FakeCompleter::new(&["验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验abcdef/", "abcdef/"]);
+        let mut menu = ColumnarMenu::default().with_name("testmenu");
+        let mut editor = Editor::default();
+
+        editor.set_buffer("a".to_string(), UndoBehavior::CreateUndoPoint);
+        menu.update_values(&mut editor, &mut completer);
+        assert!(menu.menu_string(10, true).contains("验"));
+    }
 }
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -830,7 +830,7 @@ mod tests {
     #[test]
     fn test_menu_create_string_starting_with_multibyte_char() {
         // https://github.com/nushell/nushell/issues/15938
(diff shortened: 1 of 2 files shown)


ysthakur:
The IDE menu has the same problem. Could you duplicate the logic (and test) there? In the future, we can reduce the duplication a bit, but it's tolerable for now.

src/menu/columnar_menu.rs
```diff
@@ -305,7 +305,7 @@ impl ColumnarMenu {
             let shortest_base = shortest_base
                 .strip_prefix(is_quote)
                 .unwrap_or(shortest_base);
-            let match_len = shortest_base.len();
+            let match_len = shortest_base.chars().count();
 
             // Find match position - look for the base string in the suggestion (case-insensitive)
             let match_position = suggestion
```
ysthakur, line 311: I think this should be string instead of suggestion.value. that should fix the match_position out of bounds issue

dburburan: Do you mean where it goes `let match_position = suggestion...`?
I was worried that changing it to string might do some weird things because if you search for ".." it'll find it in the "..." that gets added to the end when it truncates.

dburburan: Actually now that I'm thinking about it, this would just put the underline on the ".." it wouldn't actually affect what is matched since that happens before. So it's fine to do that, I've made the change and the test still passes.

src/menu/ide_menu.rs
```diff
@@ -1487,4 +1493,50 @@ mod tests {
...
+        let mut editor = Editor::default();
+
+        editor.set_buffer("ac".to_string(), UndoBehavior::CreateUndoPoint);
+        menu.update_values(&mut editor, &mut completer);
+        assert!(menu.menu_string(10, true).contains("验"));
+    }
+
+    #[test]
+    fn test_menu_create_value_string_long_unicode_string() {
+        // https://github.com/nushell/reedline/pull/918
```
ysthakur: I don't think linking to this PR would be too useful to someone who wants to know what this test is doing. Better to say right here that it's testing for a panic

src/menu/ide_menu.rs
```diff
@@ -1487,4 +1493,50 @@ mod tests {
...
+        let mut completer = FakeCompleter::new(&["验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验验abcdef/", "abcdef/"]);
```
ysthakur: Is the second "abcdef/" suggestion necessary? I'd just do something like this:
```suggestion
        let mut completer = FakeCompleter::new(&["验验验验abcdef"]);
```

src/menu/ide_menu.rs
```diff
@@ -1487,4 +1493,50 @@ mod tests {
...
+            cursor_col: 50,
+            menu_width: 50,
+            completion_width: 50,
+            description_width: 50,
+            description_is_right: true,
+            space_left: 50,
+            space_right: 50,
+            description_offset: 50,
+            shortest_base_string: String::new(),
+        };
```
ysthakur: I think you just need to assign `completion_width`. If you make it something small (e.g., 5), you won't have to make the suggestion above super long
```suggestion
        menu.working_details.completion_width = 5;
```

src/menu/columnar_menu.rs
```diff
@@ -819,4 +826,28 @@ mod tests {
...
+        let mut menu = ColumnarMenu::default().with_name("testmenu");
+        let mut editor = Editor::default();
+
+        editor.set_buffer("ac".to_string(), UndoBehavior::CreateUndoPoint);
+        menu.update_values(&mut editor, &mut completer);
+        assert!(menu.menu_string(10, true).contains("验"));
+    }
+
+    #[test]
+    fn test_menu_create_string_long_unicode_string() {
```
ysthakur, line 843: Was this ever a problem for the columnar menu? If not, I think you can just remove this test

src/menu/ide_menu.rs
```diff
@@ -1487,4 +1493,50 @@ mod tests {
...
```
ysthakur: Is this necessary? Is the issue not triggered without this?

src/menu/columnar_menu.rs
ysthakur: Although in the future, it might be a good idea for the columnar menu to also start trimming long suggestions. We need to keep this case in mind if we do that.

ysthakur (changes requested):

src/menu/columnar_menu.rs
ysthakur: Yeah the highlighting won't be perfect but whatever

src/menu/ide_menu.rs
dburburan: Ok, changed.

src/menu/ide_menu.rs
dburburan: The second abcdef/ came from when testing it live in an actual folder, you need two files or the menu won't even appear. But you're right I just tested it here and FakeCompleter runs the menu even with one entry.

I've reduced it to just `FakeCompleter::new(&["验abc/"])` as that's all that's needed

src/menu/ide_menu.rs
dburburan: That's true changing `completion_width` is another way to trigger the crash.

I wanted to clean it up a different way though - I didn't want to make the string shorter as I feel like sending a long string through the code is like the core of the test. It should be a stress test regardless of defaults. With that in mind I changed it to this -

`FakeCompleter::new(&[&("验".repeat(205) + "abc/")])`

And then I got rid of setting the `completion_width` at all, as 205 long will be way longer than any reasonable default even if someone tweaks things in the future.

This also keeps the test code the same between ide and columnar, since in columnar there's no `completion_width` to set. You mentioned the duplication will be removed in future - doing that will be easy and obvious if the duplicated test code is identical.

Also if anyone adds truncation to columnar, or changes the defaults in ide, this will continue to stress test it.

dburburan: Actually I needed to set `completion_width` or the test never passes - I guess there's no default value. I still set it to 50 though for the above reasons.

src/menu/columnar_menu.rs
dburburan: Yep there was no issue with columnar. The test is just for future proofing.

src/menu/ide_menu.rs
dburburan: Yep not necessary - removed.

src/menu/columnar_menu.rs
ysthakur: Alright, definitely doesn't hurt to have

src/menu/ide_menu.rs
ysthakur: Gotcha. `.repeat` is fine too

ysthakur:
LGTM! Thanks for working on this.

PR #926, opened 2025-06-28
Fix shift selection  (selecting to the left includes character to the right)

collinmurch:
Fixes shift selection where holding shift and pressing the left arrow includes the character to the right of the cursor.

See https://github.com/nushell/reedline/issues/893 for more info.


blindFS:
Have you tested this on vi normal mode?

Issue #922, opened 2025-06-12
Selection in emacs mode includes one additional character


Jiogo18:
**Platform** Windows 10, Arch Linux
**Terminal software** Windows Terminal, Konsole, Alacritty

With Emacs edit mode, the first shift+left/right select 2 characters.
Git bisect found #867, specifically [editor.rs](https://github.com/nushell/reedline/pull/867/files#diff-61fb10a0f52fcd5f4dba65e7fb16ebc1801213e15c2a82423b21af6f927a4187R565-R582).
It seems Vi mode and Emacs mode should behave differently here.

## Steps to reproduce
1. `cargo run --example basic` or nushell with `$env.config.edit_mode = "emacs"`
2. Write text `0123456789`
3. Select text with shift+left or shift+right
4. An additional character to the right is selected

## Screenshots/Screencaptures

https://github.com/user-attachments/assets/b44f8646-9c08-4a70-94da-afe55a390adc
Keys:
- shift+right, right
- delete
- undo
- shift+left, left
- delete
- undo

collinmurch:
I implemented & tested https://github.com/nushell/reedline/pull/927, which fixes this described issue.

PR #927, opened 2025-06-30
Fix shift selection in vi (insert) & emacs mode

collinmurch:
This PR fixes shift selection in vi (insert) & emacs mode. The gist is that we were incorrectly selecitng an additional character when selecting leftwards (holding shift and pressing left arrow). One extra character to the right of the start location was getting highlighted and would be affected by any delete or other operations.

Tested on macOS with both vi & emacs mode. No regression of vi (normal) shift selection functionality.

Fix for this issue: https://github.com/nushell/reedline/issues/922 which I think is a duplicate of https://github.com/nushell/reedline/issues/893.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -598,32 +598,39 @@ impl Editor {
     /// If a selection is active returns the selected range, otherwise None.
     /// The range is guaranteed to be ascending.
     pub fn get_selection(&self) -> Option<(usize, usize)> {
-        self.selection_anchor.map(|selection_anchor| {
-            // Use the mode that was active when the selection was created, not the current mode
-            let inclusive = matches!(
-                self.selection_mode.as_ref().unwrap_or(&self.edit_mode),
-                PromptEditMode::Vi(PromptViMode::Normal)
-            );
-            let buffer_len = self.line_buffer.len();
+        let selection_anchor = self.selection_anchor?;
 
-            if self.insertion_point() > selection_anchor {
-                let end_pos = if inclusive {
-                    self.line_buffer.grapheme_right_index().min(buffer_len)
-                } else {
-                    self.insertion_point().min(buffer_len)
-                };
-                (selection_anchor, end_pos)
+        // Use the mode that was active when the selection was created, not the current mode
+        let inclusive = matches!(
+            self.selection_mode.as_ref().unwrap_or(&self.edit_mode),
+            PromptEditMode::Vi(PromptViMode::Normal)
+        );
+
+        let selection_is_from_left_to_right = selection_anchor < self.insertion_point();
+
+        let start_pos = if selection_is_from_left_to_right {
+            selection_anchor
+        } else {
+            self.insertion_point()
+        };
+
+        let end_pos = if selection_is_from_left_to_right {
+            if inclusive {
+                self.line_buffer.grapheme_right_index()
             } else {
-                let end_pos = if inclusive {
-                    self.line_buffer
-                        .grapheme_right_index_from_pos(selection_anchor)
-                        .min(buffer_len)
-                } else {
-                    selection_anchor.min(buffer_len)
-                };
-                (self.insertion_point(), end_pos)
+                self.insertion_point()
             }
-        })
+        } else {
+            // selection is from right to left
+            if inclusive {
+                self.line_buffer
+                    .grapheme_right_index_from_pos(selection_anchor)
+            } else {
+                selection_anchor
+            }
+        };
+
+        Some((start_pos, end_pos.min(self.line_buffer.len())))
     }
 
     fn delete_selection(&mut self) {


blindFS:
Thanks, that seems similar but clearer than the solution by Claude #911 .

Noticed one minor issue, steps to reproduce:

1. selection done in vi normal mode
2. press `c`, the last selected character is not deleted.

src/core_editor/editor.rs
```diff
@@ -625,21 +633,39 @@ impl Editor {
...
     pub fn get_selection(&self) -> Option<(usize, usize)> {
+        let inclusive = matches!(self.edit_mode, PromptEditMode::Vi(PromptViMode::Normal));
+        self.get_selection_with_mode(inclusive)
+    }
+
+    /// If a selection is active returns the selected range, otherwise None.
+    /// The range is guaranteed to be ascending.
+    /// `inclusive` controls whether the selection includes the character at the end position (Vi-style)
+    /// or excludes it (Emacs-style).
+    pub fn get_selection_with_mode(&self, inclusive: bool) -> Option<(usize, usize)> {
```
blindFS: I think we don't need this helper function with `self.edit_mode` available.

collinmurch: Sure,  I consolidated the logic.

collinmurch:
Yeah, you're totally right; I missed this test case. Should be fixed now, all the modes seem to work for me.

src/core_editor/editor.rs
```diff
@@ -205,12 +208,18 @@ impl Editor {
...
+        if select {
+            if self.selection_anchor.is_none() {
+                self.selection_anchor = Some(self.insertion_point());
+            }
         } else {
-            None
-        };
+            self.selection_anchor = None;
+        }
+    }
```
blindFS, line 230: The logic is exactly the same here. Does this change suppose to fix the `c` issue of vi norm selection?

collinmurch:
Okay I pushed up a commit that actually does fix it. My prior one did fix shift selection going leftwards but seems to have just shifted the issue to being present moving to the right for vi normal mode + cut operations. This new commit fixes it for both, but adds some complexity that I don't love.

The core issue is that when we perform the cut operation, we _immediately_ go into insert mode which then messes up the calculation for text to delete. Since we're in insert mode, it selects one character short even though the selection _was_ made with vi normal mode.

Let me know what you think of this. Also, I'm not sure exactly what you're referring to with

> There's another complication: inconsistent behavior of selection in vi normal mode triggered by v or shift + arrow key.

From my experience, I can't even get the two to behave the same on a fresh build of the main branch. It seems that shift selection for vi normal mode has some other bug that makes it operate inconsistently with the cut and other operations.

collinmurch:
Will fix the above in a sec I'm not on my normal machine with aggressive autoformatting 😅

blindFS:
I used to take this issue to test AI agents' capabilities, haha. Never found one that's good enough. Claude 4.0 >> Gemini 2.5 pro on this issue, but still lame.

collinmurch:
Bumping this.

collinmurch:
Good suggestion! I've implemented and retested with that suggestion. Let me know what you think.

Sorry for the delay, I've been travelling the last few weeks and haven't had much free time.

src/core_editor/editor.rs
```diff
@@ -16,7 +26,8 @@ pub struct Editor {
     system_clipboard: Box<dyn Clipboard>,
     edit_stack: EditStack<LineBuffer>,
     last_undo_behavior: UndoBehavior,
-    selection_anchor: Option<usize>,
+    selection: Option<Selection>,
+    edit_mode: PromptEditMode,
```
blindFS, line 22: @Bahex Should we consider moving `edit_mode` from https://github.com/nushell/reedline/blob/adeb8ea52ff78917bb5f48d8904adf6cbd29bcf9/src/engine.rs#L128-L129 here?
It doesn't make much sense to me that `edit_mode` is not native to the `editor`.

collinmurch:
Okay I updated the enum shape to be logically distinct. Noteably though this will require more than just a version bump in nushell as it changes the way we interface with the cursor. The changes are pretty small though:

https://github.com/nushell/nushell/compare/main...collinmurch:nushell:main

Which would you prefer? Reverting to the old one with the coupled enum shape or this new one, which requires nushell changes? 

I'd like to get this bug fixed. I can look at the vi issue that's currently present perhaps after.

src/enums.rs
```diff
@@ -352,47 +243,89 @@ pub enum EditCommand {
...
-            EditCommand::MoveToPosition { .. } => {
-                write!(f, "MoveToPosition  Value: <int>, Optional[select: <bool>]")
-            }
-            EditCommand::MoveLeftUntil { .. } => {
-                write!(f, "MoveLeftUntil Value: <char>, Optional[select: <bool>]")
-            }
-            EditCommand::MoveLeftBefore { .. } => {
-                write!(f, "MoveLeftBefore Value: <char>, Optional[select: <bool>]")
-            }
+            EditCommand::Move { movement, .. } => match movement {
```
blindFS: Sorry for not making it clear, but this isn't what I meant: 

1. We probably don't need the bool to mark whether a cursor movement is triggered with selection. 
2. Instead, selection status should be a native property of the editor itself.
3. In vi normal mode, press `v` to toggle selection on/off (inclusive selection)
4. In all modes, shift + arrow keys to firstly turn selection (exclusive) on, then handle the movement.

blindFS: Maybe it's too much for this bug fixing PR, I'll leave it for other reviewers to judge.

collinmurch: Ah I got it. Yeah I hear you I can probably take a stab at this too and throw up a draft MR. Otherwise I'd probably fall back on the previous implementation

blindFS: > Ah I got it. Yeah I hear you I can probably take a stab at this too and throw up a draft MR. Otherwise I'd probably fall back on the previous implementation
> 
> 

Yeah, I think reverting to previous implementation makes it easier to review.

collinmurch:
> It works well for me.
> 
> The implementation looks not very straightforward but I guess we don't have much options without completely redesigning the code base of selection and movement with selection, which I'm afraid none of us has the passion to work on.
> 
> We could use some TODO comments if we have the vision of a "better" design.
> 
> After all, I know little about the reedline code, let's await for Bahex to make the call.

Yeah agreed. I just want this darn issue fixed. I've been using my own fork of `nushell` with this fix in place for so long now. Anyways, updated again. There was a clippy failure in the CI/CD.

Bahex:
Thanks for all your work on this!

PR #929, opened 2025-07-07
Fix default prompt path

skull-squadron:
Prompt previously showed absolute path instead of `~`.

### DIFF
--- a/src/prompt/default.rs
+++ b/src/prompt/default.rs
@@ -116,21 +116,17 @@ impl DefaultPrompt {
 }
 
 fn get_working_dir() -> Result<String, std::io::Error> {
-    let path = env::current_dir()?;
-    let path_str = path.display().to_string();
-    let homedir: String = match env::var("USERPROFILE") {
-        Ok(win_home) => win_home,
-        Err(_) => match env::var("HOME") {
-            Ok(maclin_home) => maclin_home,
-            Err(_) => path_str.clone(),
-        },
-    };
-    let new_path = if path_str != homedir {
-        path_str.replace(&homedir, "~")
-    } else {
-        path_str
-    };
-    Ok(new_path)
+    let mut path = env::current_dir()?;
+
+    if let Some(home) = env::home_dir() {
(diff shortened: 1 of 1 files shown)


kronberger-droid:
Thanks for pointing out that bug.
`cd ~` reproduces the obvious one, and the old `String::replace` also misfires when a path shares a string prefix with HOME (e.g. HOME=`/home/alice`, cwd=`/home/alicebob/x` → `~bob/x`).
Your `strip_prefix` shape fixes both.

Just needs `#[allow(deprecated)]` on `env::home_dir()`, since it deprecated under our MSRV (1.63), un-deprecated in 1.85, but we don't want to bump right now. (@fdncred would this be okay for you?)

And a regression test for the `cwd == home` case.

Happy to land once those are in.

fdncred:
sounds fine

fdncred:
It looks like Windows support was completely removed. Am I reading it wrong? Also, there are no tests for Windows. Please add some.

kronberger-droid:
Hmm good call. This was not on my mind. I will take a look.

kronberger-droid:
Seems like `home_dir()` was buggy for 1.63 (why it was depreciated) and in 1.85 was stabilized for windows. So yeah you are right, we dropped windows support implicitly by staying on 1.63, but assuming behavior of 1.85.
I will take a look how to best solve it.
Thanks for pointing it out.

kronberger-droid:
Back to reading env.
Bug was essentially just for cwd == home, plus the sub-string problem I pointed out.
Ready to land if you are fine with it. I only could test it interactively on linux since I don't have a windows pc.

fdncred:
it worked good for me on windows with `cargo run --example demo`

fdncred:
lots of warnings we'll need to fix separately in another PR for new compilers

kronberger-droid:
> lots of warnings we'll need to fix separately in another PR for new compilers

I think warnings came from the fact that it was based on an old version of reedline. If i do `cargo check/clippy` on current upstream main no warnings.

Issue #853, opened 2024-11-16
Discussion: will reedline become a full-fledged "text editor?"


Jasha10:
This issue is to discuss direction and scope of reedline's "text editor" and "IDE" features.

The goal of PR #670 was "allowing the user to press "jk" (without modifiers) to move from Vi insert mode to Vi normal mode." In that PR, @benvansleen said:

> Achieving this with the existing keybindings section would require implementing something akin to emacs key-chords (since you press one key, release, then press the next). That would be a much, much wider-reaching change.

This seems related to the question of "how many full-fledged text editor features should be added to reedline?" Heavy-duty text editors like vim essentially allow the user to configure a state machine where a given keypress will have a different effect depending on the current state. (For example, the keypress 'k' would have a different effect depending on if the user is in "insert" mode or "normal" mode, and also depending on whether the 'k' immediately followed a 'j'.)

Does it make sense for reedline to move in the above direction (implementing a user-configurable state machine)? Related are questions regarding how many other "text editor" features should be included (e.g. vim allows for user-configurable callbacks, plugins, keymaps to trigger system calls or perform arbitrary turing-complete logic, etc).

I think it could be helpful to discuss vision for how far reedline will move in the direction of becoming a "text editor" / "IDE." Having a vision for this will help to ensure that:
- new features are added in a manner that's consistent with overall vision for user experience, and
- the implementation decisions made in adding new features will not be counterproductive to progress on other planned features.

Related:
- https://github.com/nushell/reedline/issues/63

fdncred:
We've wanted to have better VI keybinding functionality for a long time. You can see some of our brainstorming ideas https://github.com/nushell/reedline/issues/69, which is linked from the meta issue that you mention above.

I think reedline should cover as much functionality as is supportable. An example of what I mean by that is I like the idea of a user configurable state machine like you mention above. However, if there's no one to write it and no one to maintain it, then it's probably not something reedline should have.

I think this is the problem with many open-source projects. There are a lot of plans and ideas but unless someone comes along to implement and support some feature, it goes undone. Since we're all volunteers, we just work on things that interest us. There is a bug in reedline that has been there for years and prevents WezTerm from working properly. My solution to that was to move to Ghostty because I couldn't figure out the solution.

So, it's probably less a matter of what we want, although that's a good starting place, and more of a matter of who will write it and support it for the community.

HartBlanc:
Has outsourcing the modal editing logic to a third-party dependency been considered as an option?
[...]

PR #932, opened 2025-07-16
feat: add `ViChangeMode` event

### LINKED ISSUE #499: Rebind esc for entering vi normal mode
### Question

How can I change the binding for switching between vi insert mode and normal mode?
I've tried already the following binding with no effect:

```
    {
      name: normal_mode
      modifier: alt
      keycode: char_l
      mode: [vi_insert, vi_normal, emacs]
      event: { send: Esc }
    }
```

_No response_

bonofiglio:
Closes #499.

This PR adds the `ViChangeMode` event. Making it possible to change the mode using the normal event system. To achieve this, a new optional function `handle_mode_specific_event` was added to the `EditMode` trait. Vi mode implements this function and handles the event to change the mode accordingly. 

This will make it very easy to add the possibility of configuring mode-change keybindings in nushell.
Here's an example of how it could look:
```
{
  name: exit_insert_mode_with_ctrl_c
  modifier: control
  keycode: char_c
  mode: vi_insert
  event: {
    send: vichangemode
    mode: normal
  }
}
```

I plan on submitting a PR to nushell if this gets merged to add the functionality there too.

The main difference with #670 is that PR makes it user-configurable instead of it being a hard-coded value.
As for #853, this might prove good enough for the time being, as I'm sure many nushell users (like myself!) would really appreciate a way of remapping the vi mode change keybindings.

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -3,6 +3,8 @@ mod motion;
 mod parser;
 mod vi_keybindings;
 
+use std::str::FromStr;
+
 use crossterm::event::{Event, KeyCode, KeyEvent, KeyModifiers};
 pub use vi_keybindings::{default_vi_insert_keybindings, default_vi_normal_keybindings};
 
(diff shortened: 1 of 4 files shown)


src/edit_mode/vi/mod.rs
```diff
@@ -22,6 +22,17 @@ enum ViMode {
...
+impl ViMode {
+    pub fn from_str<S: AsRef<str>>(s: S) -> Option<ViMode> {
+        match s.as_ref() {
+            "normal" => Some(ViMode::Normal),
+            "insert" => Some(ViMode::Insert),
+            "visual" => Some(ViMode::Visual),
+            _ => None,
+        }
+    }
+}
```
sholderbach: That is a manual [`impl FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html)

bonofiglio: You're right. It felt weird to have the trait impl with an empty error, but I agree it's better.

bonofiglio:
@sholderbach Should be all set!

PR #933, opened 2025-07-16
feat: Builder option to immediately accept input

stuartcarnie:
# Summary

Adds a builder function to immediately accept the input and return from `read_line`. It introduces non-breaking changes to the API in response to [this comment](https://github.com/nushell/nushell/pull/15092#issuecomment-2968244904):

```rust
line_editor = line_editor.with_immediately_accept(true);
let input = line_editor.read_line(...)
```

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -755,8 +755,8 @@ impl Reedline {
                 #[cfg(not(feature = "external_printer"))]
                 events.push(crossterm::event::read()?);
 
-                // a line of input is completed.
                 // Receive all events in the queue without blocking. Will stop when
+                // a line of input is completed.
                 while !completed(&events) && event::poll(Duration::from_millis(0))? {
                     events.push(crossterm::event::read()?);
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -734,30 +744,32 @@ impl Reedline {
...
-            }
 
-            // If we believe there's text pasting or resizing going on, batch
-            // more events at the cost of a slight delay.
-            if events.len() > EVENTS_THRESHOLD
-                || events.iter().any(|e| matches!(e, Event::Resize(_, _)))
-            {
-                while !completed(&events) && event::poll(POLL_WAIT)? {
+                // a line of input is completed.
+                // Receive all events in the queue without blocking. Will stop when
```
132ikl: seems like these lines got switched somehow? otherwise, looks good to me

stuartcarnie: Fixed – good catch.

Issue #848, opened 2024-11-07
feature request: vi/vim mode text objects


SimonYde:
When using the vi/vim mode, I find myself lacking the general text objects that exists in most editors/shells with builtin vim mode. I'm talking about the following:

- diw
- ciw
- dap
- cap
- vaw
- viw

The basic `inside` and `around` text objects. These I find extremely useful when navigating in my shell with the vim mode and want to edit a word I just so happened to be in the middle of. 

## References

[fish](https://fishshell.com/docs/current/interactive.html#vi-mode) has this feature as an example of a shell with vi/vim-like keybinds.

fdncred:
We landed this PR recently which, I think, does some of these. Can you confirm?
https://github.com/nushell/reedline/pull/844

sholderbach:
That was just the inside delimiter text objects: `i"`/`i(` etc.

SimonYde:
> We landed this PR recently which, I think, does some of these. Can you confirm? #844

My request is I suspect an expansion on the pull request you mention. My request is to add the `word` and `paragraph` text objects from vim. Though the pull request goes a long way as is.

SimonYde:
Accidentally closed this

fdncred:
I'd love to have a full/complete vim keybind solution with configurable keybinding. Most of the vim stuff is hard coded. Would love to change that.

oiiiiiiii:
It would be great if one also could customize it. Currently `ci)` works, but I would like to add `cib` or maybe add my own delimiters (like `ciq` for changing stuff in ` `` `. There seems to be an `CutInside Value: <char> <char>` edit I could bind, but how can I bind it to `ciq`?

JonLD:
I'm currently working on (and almost finished) #939  to add word (w), WORD (W), bracket (b) and quote (q) text objects. This includes adding jumping to next quote/bracket/pair if not currently in one. I haven't added paragraph (p) at the moment but that should be quite easy to add in later.

kronberger-droid:
implemented by #939 and some more capabilities in this direction to come in #1188

PR #939, opened 2025-07-30
Vi mode text objects for word, WORD, brackets and quotes

JonLD:
### Overview
Implements the vim style inner and around (`i` and `a`) text objects for:
- word bound to `w` (a sequence of letters, digits and underscores, separated with white space)
- WORD `W` (a sequence of non-blank characters, separated with white space)
- brackets bound to `b` (any of `( )`, `[ ]`, `{ }`)
- quote bound to `q` (any of `" "`, `' '`, ` `` `)
-  prior "pair" motion for specific matching pairs (e.g. `di(` or `ci"`) extended for around versions that cover the pair characters

This addresses most of #848 although it doesn't add paragraph text object, that would be fairly straightforward to add later.

Additionally, the pair motions will now jump to the next pair if non are found within search range. For symmetric pairs like quotes searching is restricted to the current line. For asymmetric pairs like brackets search is multi-line across the whole buffer. Symmetric pairs searching does not do any parsing or matching of pairs based on language/groupings. It simply searches for the previous and next matching characters (this is consistently with vim but could be improved, perhaps using tree sitter).
### Repeats
While it might not be the intend to achieve identical behaviour as vim I thought it useful to compare to the default vim behaviour of repeats.
Around brackets:
In the current implementation repeat motions for `around` text objects will cause a jump to the next bracket. 
e.g. 2daw on "(first)between(second)" while in the first bracket will delete both the first and second bracket, included the parentheses themselves, and move the cursor to the end "between".
Vim works around this by not letting you do repeats for "around" motions unless inside a nested bracket.
Inner words:
Repeat motions for words also differ from default vim behaviour for repeated word text objects. In vim a `2diw` performs a single around diw command and then the repeat command is `dw`. The difference in behaviour is that in the reedline implementation `2daw` on `delete th|is word` will result in `delete|word` whereas in vim it would be `delete |word`.

This could perhaps be worked around by adding a method of passing an alternate command for repeats to the command module which could be made in a future change. I think the current behaviour is satisfactory for most use cases.

### Word identification
Currently due to using the unicode-segmentation crate for splitting at word word boundaries word does not have the same meaning as Vi/Vim.  This means `diw` on `not.a..wo|rd` will delete the whole WORD rather than just the word under the cursor.

### DIFF
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -30,6 +30,10 @@ pub enum TextObjectType {
     Word,
     /// WORD (delimited only by whitespace)
     BigWord,
+    /// (, ), [, ], {, }
+    Brackets,
+    /// ", ', `
+    Quote,
 }
 
 /// Text objects that can be operated on with vim-style commands
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -527,9 +527,9 @@ impl Display for EditCommand {
             #[cfg(feature = "system_clipboard")]
             EditCommand::PasteSystem => write!(f, "PasteSystem"),
             EditCommand::CutInsidePair { .. } => write!(f, "CutInside Value: <char> <char>"),
-            EditCommand::CopyInsidePair { .. } => write!(f, "YankInside Value: <char> <char>"),
-            EditCommand::CutAroundPair { .. } => write!(f, "CutAround Value: <char> <char>"),
-            EditCommand::CopyAroundPair { .. } => write!(f, "YankAround Value: <char> <char>"),
+            EditCommand::CopyInsidePair { .. } => write!(f, "CopyInsidePair Value: <char> <char>"),
+            EditCommand::CutAroundPair { .. } => write!(f, "CutAroundPair Value: <char> <char>"),
+            EditCommand::CopyAroundPair { .. } => write!(f, "CopyAroundPair Value: <char> <char>"),
             EditCommand::CutTextObject { .. } => write!(f, "CutTextObject Value: <TextObject>"),
             EditCommand::CopyTextObject { .. } => write!(f, "CopyTextObject Value: <TextObject>"),
         }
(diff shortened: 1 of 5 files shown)


ayax79 (approved):
I have been wanting this for a while. This much better than the hack a while back.
The code looks clean and seems to work fine in nushell (minus a couple minor changes).

Great work!

src/enums.rs
```diff
@@ -546,8 +614,12 @@ impl EditCommand {
             EditCommand::CopySelection => EditType::NoOp,
             #[cfg(feature = "system_clipboard")]
             EditCommand::CopySelectionSystem => EditType::NoOp,
-            EditCommand::CutInside { .. } => EditType::EditText,
-            EditCommand::YankInside { .. } => EditType::EditText,
+            EditCommand::CutInsidePair { .. } => EditType::EditText,
+            EditCommand::CopyInsidePair { .. } => EditType::EditText,
+            EditCommand::CutAroundPair { .. } => EditType::EditText,
+            EditCommand::CopyAroundPair { .. } => EditType::EditText,
```
Juhan280, line 620: I believe the EditType for Copy command should be a noop.

```suggestion
            EditCommand::CopyInsidePair { .. } => EditType::NoOp,
            EditCommand::CutAroundPair { .. } => EditType::EditText,
            EditCommand::CopyAroundPair { .. } => EditType::NoOp,
```

fdncred:
Let's move forward with this. Thanks!

fdncred:
@JonLD would you mind looking at this PR https://github.com/nushell/nushell/pull/16791. I left out https://github.com/nushell/reedline/pull/939/files#diff-2dbf826b03a6d3f8d1d1e31324494945d307b79a9592d7eddf70f1d9bdfd1473R408-R416 because i wasn't sure how to implement them with TextObject in nushell.

oiiiiiiii:
Is it possible to customize the keys for it somehow?

PR #941, opened 2025-08-07
Update rstest from 0.23 to 0.26, the current version

musicinmybrain:
I confirmed that `cargo test` still passes.

https://github.com/la10736/rstest/blob/v0.26.1/CHANGELOG.md

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -752,19 +752,18 @@ checksum = "ba39f3699c378cd8970968dcbff9c43159ea4cfbd88d43c00b22f2ef10a435d2"
 
 [[package]]
 name = "rstest"
-version = "0.23.0"
+version = "0.26.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0a2c585be59b6b5dd66a9d2084aa1d8bd52fbdb806eafdeffb52791147862035"
+checksum = "f5a3193c063baaa2a95a33f03035c8a72b83d97a54916055ba22d35ed3839d49"
 dependencies = [
  "rstest_macros",
- "rustc_version",
 ]
 
 [[package]]
 name = "rstest_macros"
-version = "0.23.0"
+version = "0.26.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


musicinmybrain:
Looks like `build-lint-test` is failing with some new clippy lints; since this PR doesn’t touch the source code, those would be best fixed in a separate PR.

kronberger-droid:
I am fine with landing this.
We lose one transitive dependency.
But would bump dev MSRV 1.67.1 -> 1.70.0
Clippy is happy with one `cargo clippy fix` and tests still pass afterwards.
Are you fine with the MSRV bump for devs @fdncred?

fdncred:
If we think it's necessary. We want to try to keep reedline with as low a MSRV as possible so more people can use it. But since 1.70 was released about 2 years ago, it's probably ok.

kronberger-droid:
Well from my side I have no reason to update yet.
@musicinmybrain did you have something in mind you want to use from the new version?
Otherwise I would close it and we wait until we have a real use case for the features in 0.26.

musicinmybrain:
> Well from my side I have no reason to update yet. @musicinmybrain did you have something in mind you want to use from the new version? Otherwise I would close it and we wait until we have a real use case for the features in 0.26.

Not urgently, no. We patched the [`rust-reedline` package in Fedora](https://src.fedoraproject.org/rpms/rust-reedline) to use a newer `rstest` because we try to avoid the proliferation of unnecessary compat packages, which greatly increase our downstream packaging and maintenance burden. I opened this PR because we have a policy of [staying close to upstream](https://docs.fedoraproject.org/en-US/package-maintainers/Staying_Close_to_Upstream_Projects/) and [offering changes upstream](https://docs.fedoraproject.org/en-US/packaging-guidelines/PatchUpstreamStatus/). I have no complaints about you choosing to ignore it for MSRV reasons or any other reasons. Thanks for taking the time to consider it. We can keep adjusting this version bound in a downstream-only patch.

kronberger-droid:
@musicinmybrain Thanks for your understanding. It just doesn't make sense to bump our MSRV if we don't use any of the features of the new version.
Maybe we will in the future.

PR #942, opened 2025-08-10
Fix missing use in README.md

CastilloDel:
Hey! I was looking at the instructions and noticed this was missing to have it compile. There was a correct example [here](https://docs.rs/reedline/latest/reedline/index.html#integrate-with-custom-tab-completion) though


sholderbach (approved):
Thanks! good catch.

I was under the illusion that we already embedded the README in the lib.rs and this would be doctested but well nope

PR #952, opened 2025-09-17
fix: set default prompt colors to reset

AnandajithS:
The default left and right prompt colors were set as `Green` and `AnsiValue(5)` respectively, which caused the prompt to be colored even when it was reset. This fix makes the prompts respect the terminal default colors when style="none"

Fixes nushell/nushell#16384

### DIFF
--- a/src/prompt/base.rs
+++ b/src/prompt/base.rs
@@ -9,10 +9,10 @@ use {
 };
 
 /// The default color for the prompt, indicator, and right prompt
-pub static DEFAULT_PROMPT_COLOR: Color = Color::Green;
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. The only problem I see here is that neither the left nor right prompt has default colors any longer.

AnandajithS:
Apologies for the delay in following up.and thank you for the feedback. In this fix, the prompts will take on the default colors of the terminal when `style="none"`. The colors can still be changed either by applying a style or by adjusting the terminal settings as needed. If this is acceptable, could it be merged?

PR #953, opened 2025-09-21
vi-mode: add gg/G motions and fix ^ motion behavior

Juhan280:
This PR adds the `gg` and `G` motions and adjusts the `^` motion in `vi-mode`. The `gg` motion moves the cursor to the first character of the buffer and `G` moves it to the last. Unlike Vim, `G` takes the cursor to the end of the buffer instead of the first character of the last line. Tests are included for `dgg`, `dG`, `cgg`, and `cG`.

The `^` motion now moves the cursor to the first non-blank character of the line instead of the line start. Test cases for `d^` and `c^` are updated accordingly.

Also, the name `NonBlankStart` could be better. Please tell me if you come up with something better.

##### This is the first pr I made in rust, if I did any mistake please point it out. Thanks

### DIFF
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -3,6 +3,28 @@ use serde::{Deserialize, Serialize};
 use std::fmt::{Display, Formatter};
 use strum_macros::EnumIter;
 
+/// Which mouse button was pressed.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum MouseButton {
+    /// Left mouse button
+    #[default]
+    Left,
+    /// Right mouse button
+    Right,
+    /// Middle mouse button
+    Middle,
+}
+
+impl From<crossterm::event::MouseButton> for MouseButton {
+    fn from(button: crossterm::event::MouseButton) -> Self {
+        match button {
+            crossterm::event::MouseButton::Left => Self::Left,
+            crossterm::event::MouseButton::Right => Self::Right,
+            crossterm::event::MouseButton::Middle => Self::Middle,
+        }
+    }
+}
+
 /// Valid ways how `Reedline::read_line()` can return
 #[derive(Debug)]
 pub enum Signal {
@@ -808,8 +830,15 @@ pub enum ReedlineEvent {
     /// Esc event
     Esc,
 
-    /// Mouse
-    Mouse, // Fill in details later
+    /// Mouse click event with screen coordinates
+    Mouse {
+        /// Column (x) position, 0-indexed from left
+        column: u16,
+        /// Row (y) position, 0-indexed from top
+        row: u16,
+        /// Which mouse button was clicked
+        button: MouseButton,
+    },
 
     /// trigger terminal resize
     Resize(u16, u16),
@@ -899,7 +928,11 @@ impl Display for ReedlineEvent {
             ReedlineEvent::Submit => write!(f, "Submit"),
             ReedlineEvent::SubmitOrNewline => write!(f, "SubmitOrNewline"),
             ReedlineEvent::Esc => write!(f, "Esc"),
-            ReedlineEvent::Mouse => write!(f, "Mouse"),
+            ReedlineEvent::Mouse {
+                column,
+                row,
+                button,
+            } => write!(f, "Mouse({}, {}, {:?})", column, row, button),
             ReedlineEvent::Resize(_, _) => write!(f, "Resize <int> <int>"),
             ReedlineEvent::Edit(_) => write!(
                 f,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -748,7 +748,7 @@ dependencies = [
 
 [[package]]
 name = "reedline"
-version = "0.44.0"
+version = "0.45.0"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ license = "MIT"
 name = "reedline"
 repository = "https://github.com/nushell/reedline"
 rust-version = "1.63.0"
-version = "0.44.0"
+version = "0.45.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [lib]
--- a/examples/mouse_click.rs
+++ b/examples/mouse_click.rs
@@ -1,0 +1,27 @@
+// Enable mouse click-to-cursor support.
+// cargo run --example mouse_click
+
+use reedline::{DefaultPrompt, MouseClickMode, Reedline, Signal};
+use std::io;
+
+fn main() -> io::Result<()> {
+    let mut line_editor = Reedline::create().with_mouse_click(MouseClickMode::EnabledWithOsc133);
+    let prompt = DefaultPrompt::default();
+
+    println!("Mouse click-to-cursor enabled.");
+    println!("Type some text, then click in the line to move the cursor.");
+    println!("Press Enter to submit, Ctrl-D/Ctrl-C to exit.");
+
+    loop {
+        let sig = line_editor.read_line(&prompt)?;
+        match sig {
+            Signal::Success(buffer) => {
+                println!("We processed: {buffer}");
+            }
+            Signal::CtrlD | Signal::CtrlC => {
+                println!("\nAborted!");
+                break Ok(());
+            }
+        }
+    }
+}
--- a/examples/semantic_prompt_interactive.rs
+++ b/examples/semantic_prompt_interactive.rs
@@ -1,0 +1,51 @@
+// Semantic prompt interactive example
+// Run with: cargo run --example semantic_prompt_interactive
+//
+// This example demonstrates OSC 133 semantic prompt markers.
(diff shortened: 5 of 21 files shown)


src/core_editor/editor.rs
```diff
@@ -852,6 +871,15 @@ impl Editor {
         self.copy_range(copy_range);
     }
 
+    pub(crate) fn copy_from_line_non_blank_start(&mut self) {
+        let offset_a = self.line_buffer.insertion_point();
```
ysthakur: Not a blocker, but I'd prefer more descriptive names than `offset_a` and `offset_b`.

I do like this implementation more than that of `copy_from_line_start`. Not sure why it introduces a new scope and creates that `temp_pos` variable.

Juhan280: I wanted to name them `start_offset` and `end_offset`, but their role changes depending on the position of the cursor. So, this is what I settled with. Naming variables are hard, if you have any suggestion, let me know.

ysthakur: Ack, makes sense. Okay to leave it as is then

ysthakur: Ack, makes sense. Okay to leave it as is then

src/core_editor/editor.rs
```diff
@@ -357,6 +372,22 @@ impl Editor {
         }
     }
 
+    fn cut_from_start_linewise(&mut self) {
+        // TODO: make delete motion not leave a empty line
```
Juhan280: I can't figure out how I can leave an empty line when doing `cgg` but not leave any when doing `dgg`, like `cc` and `dd`.

src/core_editor/editor.rs
Juhan280: I renamed them to `cursor_pos` and `other_pos`

ysthakur:
> It is because, if you copy something linewise in vim, it will also paste linewise i.e. it won't paste it between two characters of a line. Edit: I was going through the source code again, and then discovered reedline actually support linewise copy. I will update my code to use it.

Great! Thanks for the explanation before the edit.
 
> > For `G`, though, I'd like to verify that Vim doesn't go to the first column. Because if it does, the implementation in this PR doesn't match.
> 
> Yes, this is where I intentionally make it different. In a commandline context, needing to go at the absolute end of the buffer is more frequent. You can go the the first character of the line using `G0`. In vi mode of fish shell, `G` also takes the cursor to the absolute end of the buffer.

Makes sense
 
> Understood. If you ask me, I can try to split this PR. I already made separate commits for them.

Nah, you're good. This is a fairly small PR anyway.

src/edit_mode/vi/parser.rs
```diff
@@ -541,20 +541,34 @@ mod tests {
...
     #[case(&['c', 'F', 'a'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::CutLeftUntil('a')]), ReedlineEvent::Repaint]))]
     #[case(&['c', 'T', 'a'], ReedlineEvent::Multiple(vec![ReedlineEvent::Edit(vec![EditCommand::CutLeftBefore('a')]), ReedlineEvent::Repaint]))]
+    #[case(&['c', 'g', 'g'], ReedlineEvent::Multiple(vec![
+        ReedlineEvent::Edit(vec![EditCommand::CutFromStartLinewise]),
+        ReedlineEvent::Repaint,
+    ]))]
+    #[case(&['c', 'G'], ReedlineEvent::Multiple(vec![
+        ReedlineEvent::Edit(vec![EditCommand::CutToEndLinewise]),
+        ReedlineEvent::Repaint,
+    ]))]
```
ysthakur, line 567: I see that there's no cases for yank, but can you check if it's possible to add tests for it, or if there's an actual reason why there aren't tests yet? If the former is the case, testing `y^`, `ygg`, and `yG` would be good

src/core_editor/editor.rs
ysthakur: Might be missing something here, but what about adding a parameter to `CutFromStartLinewise` that indicates whether we're `d`eleting or `c`hanging?

Juhan280: It seems like `dd` and `cc` are using different enum variants entirely

```rs
// dd:
Motion::Line => Some(vec![ReedlineOption::Edit(EditCommand::CutCurrentLine)]),
// cc:
Motion::Line => Some(vec![
    ReedlineOption::Edit(EditCommand::MoveToLineStart { select: false }),
    ReedlineOption::Edit(EditCommand::CutToLineEnd),
]),
```
That also means that the `cc` command is **not** copying in `ClipboardMode::Lines` mode

src/edit_mode/vi/parser.rs
Juhan280: I don't know why there is not test cases for other yank motions. I just assumed there is a reason for that. I added the new test cases for `y^`, `ygg`, and `yG` and I don't see any obvious problems with that. The tests are passing.

ysthakur: Cool, thanks. I'm assuming the `y` tests were just forgotten

src/core_editor/editor.rs
ysthakur: > That also means that the cc command is not copying in ClipboardMode::Lines mode

Interesting. We should probably fix that, but pretty low priority I guess.

---

Can we make `EditCommand::CutFromStartLinewise` look like this?
```
/// Cut from the start of the buffer to the line of insertion point
CutFromStartLinewise { include_blank_line: bool }
```

And then in `to_reedline_with_motion()`, in the `Delete` case, we map `Motion::FirstLine` to `CutFromStartLinewise { include_blank_line: false }`, but in the `Change` case, we map it to `CutFromStartLinewise { include_blank_line: true }`. Would that work?

src/enums.rs
```diff
@@ -499,11 +505,15 @@ impl Display for EditCommand {
             EditCommand::Complete => write!(f, "Complete"),
             EditCommand::CutCurrentLine => write!(f, "CutCurrentLine"),
             EditCommand::CutFromStart => write!(f, "CutFromStart"),
-            EditCommand::CutFromStartLinewise => write!(f, "CutFromStartLinewise"),
+            EditCommand::CutFromStartLinewise { .. } => {
+                write!(f, "CutFromStartLinewise leave_blank_line: <bool>")
```
Juhan280: I am not really sure how the display for variants with fields supposed to look like. And what determines a field is optional. `MoveLeft` has one field `select` and it says optional in the display. And then `MoveRightBefore` has two fields but only one is mentioned in the display and that's required.

ysthakur: Yeah, I'm not sure about what makes a field optional either. As for `MoveRightBefore` mentioning both fields, I think that might just be because we forgot to update this method after updating the `MoveRightBefore` variant at some point.

I do think you're supposed to use `Value: <bool>` rather than `leave_blank_line: <bool>`. I'll ask on Discord to see if anyone knows about this

src/core_editor/editor.rs
```diff
@@ -372,12 +376,15 @@ impl Editor {
...
-        // TODO: make delete motion not leave a empty line
+    fn cut_from_start_linewise(&mut self, leave_blank_line: bool) {
         let insertion_offset = self.line_buffer.insertion_point();
         let end_offset = self.line_buffer.get_buffer()[insertion_offset..]
             .find('\n')
-            .map_or(self.line_buffer.len(), |offset| insertion_offset + offset);
+            .map_or(self.line_buffer.len(), |offset| {
+                // When leave_blank_line is true, we do **not** add 1 to the offset
+                // So there will remain an empty line after the operation
+                insertion_offset + offset + !leave_blank_line as usize
```
ysthakur: Converting bools to usizes makes me nervous, I'd like to make this and the one below an if statement instead

src/enums.rs
ysthakur: [Asked on Discord](https://discord.com/channels/601130461678272522/855886335980994600/1466267512340742461), fdncred found that  https://github.com/nushell/reedline/pull/689 is the source of the `Optional[]` stuff, but I'm still not sure what it means.

I think we can change this line to the following and that'll be good enough:
```
write!(f, "CutFromStartLinewise Value: <bool>")
```

Juhan280: I could do that, but then I feel like there wouldn't be a user facing documentation explaining what that bool does

Juhan280: > As for `MoveRightBefore` mentioning both fields, I think that might just be because we forgot to update this method after updating the `MoveRightBefore` variant at some point.

I think using a derive macro to generate the `Display` impl is the best way to ensure that doesn't happen in the future. It will put them in one place.

ysthakur: Yeah, a macro could be good

I do agree that using the actual field name is good, but at this point, every other command uses value, so I'd like to stay consistent

ysthakur:
> Also are we okay with the name NonBlankStart

Sounds fine to me

> Also are we sure the value of CutFromStartLinewise and CutToEndLinewise should be required. Or should it be optional? If so, what should be the default?

Making it required seems best to me. Might be slightly more annoying, but I don't think either true or false is clearly the better default

PR #958, opened 2025-10-07
fix: bashism parsing

migraine-user:
The current parsing behavior of bashism such as `!-<n>` or `!<n>` is faulty. `!<n>---` is parsed as a whole, and rewritten the same as `!<n>`, and `!-----<n>` is the same as `!-<n>`. You can see what other buggy behavior is possible if you take a closer look at the changed code.
Now they all correctly parse.


fdncred:
I haven't tested yet. Does `!term` still work too?

migraine-user:
Yes. I don't think I changed anything relating to `!term`

PR #960, opened 2025-10-10
Feature: Helix mode

schlich:
Edit 2/17/2026: PR in progress!!! See comments below


schlich:
This is getting pretty close to complete w/comprehensive tests for the commands listed, but i'm gonna mark this as draft temporarily since i want to try to reconcile upstream Helix's architecture a little more closely with reedline's interface... hopefully it will reduce the footprint of this PR.

schlich:
Hi, I am although I haven't had too much free time for open source lately.  If youd like to pick it up I think right now it would just be helpful to try the demo and double check feature parity.  I'm a total rust noob too so any refactoring that's more rust-idiomatic or reduces the size of the PR would definitely be welcome.  Let me know if there's anything else I can help with, I'm gonna try to use some of the holiday season to catch up on things

kronberger-droid:
Hey there, I have been dreaming of a helix mode in nushell for a long time now. There is the possibility to use ctrl-o to open helix and edit the prompt, but it does not feel natural.
I have been looking for projects to contribute to and this feature is close to my heart. I would start with some of the things mentioned in @adaschma's post.
@schlich @freepicheep are any of you already working on some of those?

fdncred:
I could be wrong because I haven't studied it deeply but I thought modalkit was only for vim. Do you think it could work for helix as well?

Ratatui definitely can use crossterm as a backend, we use it in nushell commands like `explore`, `explore regex` and `explore config` and clearly reedline uses crossterm so clearly there is some overlap. Of course, we don't want to turn reedline into a tui but since it already uses crossterm there should be a way to make it work, but I'm just guessing.

schlich:
Excellent suggestions.  Perhaps the new plan could just be to essentially start from a fresh branch of main and do small red/green/refactor cycles for review, that way we can get this feature out and accomplish the refactoring in parallel

schlich:
alrighty, i just pushed up a total rework focused on a 100% implementation of helix mode based on modalkit's primitives.  it took me a while to figure out some patterns and i still dont think i'm all the way there, but this feels a lot more sustainable and workable, with the bonus that it's not going to rely on any vi/emacs code so we can fully gate it on the feature flag.   

This has been a great project for learning me some Rust!

schlich:
@kronberger-droid best place to start would probably be in the prelude section
 https://docs.rs/modalkit/latest/modalkit/prelude/index.html

fdncred:
I'm up for gutting how reedline handles all the keystroke related stuff in favor of modalkit or some other crate that standardizes a way to use keystrokes for emacs, vim, and helix (and works with well-known vim/helix/emacs standards). However, we need to figure out a way to do that in a somewhat non-breaking change way. What I'm thinking is feature-gating the changes so the default is close to what it is today, and the feature is the gutted version that may work very different on the inside but to the end user, it just works (tm) just better.

The thing that makes major code surgery difficult is we don't have enough test cases to ensure non-breaking changes plus it's extremely difficult to review huge PRs. If it's in a feature, we could ship close-to-working stuff that isn't ready for prime-time (but compiles), piece-by-piece without the fear of it breaking every user who uses reedline whether in nushell or just as a crate in their cool app.

Thoughts?

kronberger-droid:
You mean the missmatch between vi and hx or the missmatch between modalkit and reedline?

I think it makes sense to use modalkit as a backend and make it feature gated. But right now it feels like we would have to do major changes in the Editor, i am not sure if it can be easily feature gated without a lot of code dublication. But it is definitely the best option.

I will now work on a clean branch and implement a feature gated helix mode, since i have now identified a possible minimal implementation, based on some new EditorCommands and a new Selection struct in the Editor.
But will now spend some time reading helix's implementation, but tell you if i have something working.

src/edit_mode/hx/mod.rs
```diff
@@ -0,0 +1,1941 @@
...
+        let event = hx.parse_event(key_press(KeyCode::Backspace, KeyModifiers::NONE));
+        assert_eq!(
+            event,
+            ReedlineEvent::Edit(vec![
+                EditCommand::Backspace,
+                EditCommand::HxExtendSelectionToInsertionPoint,
+            ])
+        );
+        // insert_style stays After — selection is still tracked.
+        assert_eq!(hx.insert_style, InsertStyle::After);
```
schlich: is this just making sure that the selection anchor aligns with the cursor e.g. width-1 selection? not sure how the context for this & the previous test case differ, or how insert_style is being set/what that relates to

src/edit_mode/hx/mod.rs
```diff
@@ -0,0 +1,1941 @@
...
+            ReedlineEvent::Edit(vec![
+                EditCommand::Backspace,
+                EditCommand::HxShiftSelectionToInsertionPoint,
+            ])
+        );
+        assert_eq!(hx.insert_style, InsertStyle::Before);
+    }
+
+    #[test]
+    fn delete_in_insert_clears_selection() {
```
schlich: is this right? for me, trying this out in helix only deletes the singular character under the cursor

src/edit_mode/hx/mod.rs
```diff
@@ -0,0 +1,1941 @@
...
+    //   ]text[   -- backward selection: `]` = head, `[` = anchor
+    //   [h]ello  -- 1-wide selection (anchor=0, head=1)
+    //
+    // Helix uses CharClass-based word boundaries (Word/Punctuation/Whitespace),
+    // so "foo.bar" is three words: "foo", ".", "bar".
+    // WORD motions (W/B/E) split only on whitespace.
+
+    /// Parse selection notation into (buffer, anchor, head).
+    ///
+    /// `[` marks the **anchor** position and `]` marks the **head** position
```
schlich, line 734: love this convenience!

src/edit_mode/hx/mod.rs
kronberger-droid: InsertStyle tracks how insert mode was entered. It determines whether each keystroke also adjusts the selection. This test checks that backspace in a-mode still emits the selection tracking command (which it did not due to a bug -- that's why this test now exists), and that the style isn't accidentally reset.

Try in helix to have some non colapsed selection and use a/i, we encounter two different behaviours:
i - we push the selection forward
a - we pull the selection with us

Maybe the Name is confusing it would make more sense to say shrinks not extends.

src/edit_mode/hx/word.rs
```diff
@@ -0,0 +1,792 @@
+//! Helix-style word boundary detection and motion target computation.
+//!
+//! Helix classifies characters into three categories and treats transitions
+//! between classes as word boundaries:
+//! - **Word**: alphanumeric + underscore
+//! - **Punctuation**: everything else that isn't whitespace
+//! - **Whitespace**: spaces, tabs, etc.
+//!
+//! "Small word" motions (w/b/e) break at any class transition.
+//! "Big WORD" motions (W/B/E) only break at whitespace boundaries.
```
schlich, line 10: assuming this is different from vi mode, might be worth a note on how exactly its different

src/edit_mode/hx/word.rs
```diff
@@ -0,0 +1,792 @@
...
+        }
+        (a, b) => a != b,
+    }
+}
+
+/// Precomputed byte-offset table for char-index ↔ byte-offset conversion.
+///
+/// Reedline uses byte offsets everywhere (insertion_point, HxRange),
+/// but word motion logic iterates over `Vec<char>` with char indices
+/// (matching Helix's char-index Rope API). This table bridges the two
```
schlich, line 64: nice!

src/edit_mode/hx/word.rs
```diff
@@ -0,0 +1,792 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // ── Test runner ─────────────────────────────────────────────────────
+    //
+    // Adapted from helix-core/src/movement.rs test infrastructure.
```
schlich: permalink would be good to include here

src/edit_mode/hx/word.rs
```diff
@@ -0,0 +1,792 @@
...
+    // Internally word motions use char indices but convert to byte offsets
+    // via CharOffsets. All ASCII test strings have identical char/byte values.
+
+    /// Run a motion function against a table of (sample, scenarios).
```
schlich, line 392: i wonder if rtest fixtures would be appropriate here

src/edit_mode/hx/mod.rs
kronberger-droid: You are right! Handler and test need to be changed.

src/enums.rs
```diff
@@ -466,6 +466,107 @@ pub enum EditCommand {
...
+    /// selection tracks text that was pushed right by an insertion before it.
+    /// Used after `i` (insert) so the selection stays on the same text.
+    #[cfg(feature = "hx")]
+    HxShiftSelectionToInsertionPoint,
+}
+
+/// Whether a motion resets the selection anchor (Move) or keeps it (Extend).
+#[cfg(feature = "hx")]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Movement {
```
schlich, line 564: Maybe this can be called something like MotionType?

src/edit_mode/hx/word.rs
kronberger-droid: - Underscore: Helix treats _ as a word character. Unicode segmentation treats it as a boundary. So hello_world is one word in Helix but    
  three segments (hello, _, world) in Unicode.                                                                                             
- Apostrophes: Unicode has special MidLetter rules, so don't stays as one word. Helix sees three segments: don (word), ' (punct), t 
  (word).                                                                                                                                    
- Line endings: Helix has a separate Eol class distinct from whitespace, so \n creates a boundary with adjacent spaces. Unicode lumps them 
  together.

```suggestion
  //! "Small word" motions (w/b/e) break at any class transition.
  //! "Big WORD" motions (W/B/E) only break at whitespace boundaries.
  //!
  //! Reedline's existing word motions use Unicode word boundary rules.
  //! Helix uses a simpler three-class model where `_` is a word character
  //! (so `hello_world` is one word) and apostrophes are punctuation
  //! (so `don't` is three segments). Line endings are also a distinct class.
```

schlich (commented):
Overall this is great!  Glad you were able to get some of the less prominent binds/actions in on this first round.  Definitely a lot of room to cut down on LOC between repeated patterns and stuff that can easily be subbed out with imports from the modalkit/keybindings crates but the logic wasn't hard to follow for either the implementation or the testing.  Even still there might be more syntactic sugar we can use to make the intent of the test cases more clear like you did with the bracket notation,  but i'm satisfied the test cases are thorough.

Pleasantly surprised we have enough features implemented that I think a merge could probably actually be used by most people without too many hiccups!  also a good stopping point to work on refactoring before the remaining features!!

I have a decent idea of where the seams are to start on the refactoring work but that can wait for a different pr/issue 🙂

thanks for picking up the torch!

src/enums.rs
kronberger-droid: Yeah probably more descriptive if read out of context.

src/edit_mode/hx/mod.rs
```diff
@@ -0,0 +1,1941 @@
...
+                    ]),
+                    InsertStyle::After => ReedlineEvent::Edit(vec![
+                        EditCommand::Backspace,
+                        EditCommand::HxExtendSelectionToInsertionPoint,
+                    ]),
+                    InsertStyle::Plain => ReedlineEvent::Edit(vec![EditCommand::Backspace]),
+                },
+                (KeyCode::Delete, _) => {
+                    self.insert_style = InsertStyle::Plain;
+                    ReedlineEvent::Edit(vec![EditCommand::Delete, EditCommand::HxClearSelection])
```
kronberger-droid:
```suggestion
                    ReedlineEvent::Edit(vec![EditCommand::Delete])
```

src/edit_mode/hx/mod.rs
```diff
@@ -0,0 +1,1941 @@
...
+            insert_style: InsertStyle::Before,
+            ..Default::default()
+        };
+        let event = hx.parse_event(key_press(KeyCode::Delete, KeyModifiers::NONE));
+        assert_eq!(
+            event,
+            ReedlineEvent::Edit(vec![EditCommand::Delete, EditCommand::HxClearSelection])
+        );
+        assert_eq!(hx.insert_style, InsertStyle::Plain);
+    }
```
kronberger-droid:
```suggestion
    #[test]
    fn delete_in_insert_preserves_selection() {
        let mut hx = Helix {
            mode: HelixMode::Insert,
            insert_style: InsertStyle::Before,
            ..Default::default()
        };
        let event = hx.parse_event(key_press(KeyCode::Delete, KeyModifiers::NONE));
        assert_eq!(
            event,
            ReedlineEvent::Edit(vec![EditCommand::Delete])
        );
        assert_eq!(hx.insert_style, InsertStyle::Before);
    }
```

src/edit_mode/hx/mod.rs
schlich: Ah I see!  not sure if i completely noticed this behavior before.  I agree that something like InsertContext or InsertMethod might be better than InsertStyle.

I'm also not completely convinced that After and Before are actually two separate behaviors -- if we defined the backspace with respect to the anchor position instead of in absolute terms (which im pretty sure is possible if not with existing objects then definitely with modalkit primitives) it might be consistent in both cases and we would only need "reset" and "maintained" enum fields or something like that.  The test assertions could test absolute distance from the anchor or something.  

Anyways i'm probably overthinking that, this looks good as-is for now besides the potential renaming if we can agree on something

src/edit_mode/hx/word.rs
kronberger-droid:
```suggestion
    // Adapted from [helix-core/src/movement.rs](https://github.com/helix-editor/helix/blob/51ec572a27a8c1267afbc07e6c1583585c6363dc/helix-core/src/movement.rs) test infrastructure.
```

src/edit_mode/hx/mod.rs
kronberger-droid: You're right that they could be unified. Helix itself has no InsertStyle equivalent. But we don't have much of the infrastructure helix has to achieve this.

Any suggestions for naming?

schlich: Thought about this for a few minutes and i feel like it really could be named any number of ways but perhaps something like this is a little clearer...

```rust
enum SelectionAdjustmentType {
    /// No selection tracking (entered via `I`, `A`, `c`, or other commands
    /// that clear the selection before entering Insert mode).
    #[default]
    None,
    /// How selections adjust to text insertions/deletions after `i`: text is inserted/deleted *before* the selection, so both
    /// anchor and head must be shifted forward/backwards by the inserted byte length.
    Shifting,
    /// Entered via `a`: text is inserted *after* the selection, so the
    /// selection extends/contracts along with the head cursor while the selection anchor cursor remains in place.
    Anchored,
}
```

My rust noobishness is showing here bc i dont know if you can return None for an enum but it *seems* more correct and useful to have a None / non-value there  The Optional enum doesn't seem to do what I thought it did even though that was my first thought? but maybe it needs to be restructured along those lines?  Regardless not sure "Plain" is very accurate but if we cant use None we can go with something equally meaningless

schlich: ps I think [CursorAdjustment](https://docs.rs/modalkit/latest/modalkit/editing/cursor/enum.CursorAdjustment.html) and [SelectionResizeStyle](https://docs.rs/modalkit/latest/modalkit/prelude/enum.SelectionResizeStyle.html) are the closest modalkit corroloaries, this feels like maybe something in between

kronberger-droid: The rust way would be an `Option<SelectionAdjustmentType>`
Then you can do `if let Some(x) = ... { match x {..}}` when handling it or just use `.is_some()`.

Adding None in an Enum is an anti pattern since you rob yourself from the quality of life features the option Enum provides.

I would also drop the Type Suffix `SelectionAdjustment` is just fine since Snake-case already shows its a type.

Otherwise I agree with the naming.

fdncred:
I think I've landed all the helix PRs now. Are we ready to move on to the next step? What is the next step?

kronberger-droid:
Hey @schlich @fdncred,
great to see the first steps towards helix mode being merged.

I guess the natural next step would be to introduce the selection/range model, since right now motions always collapse the selection.

@schlich do you already have in mind how to implement this and where to put it?

One option would be inside `HelixMachine`, keeping selection logic local.
Or we'd have to promote it to the `LineBuffer` and helix is just the mode that does not collapse the selection, but this would mean we have to touch all other modes, which probably makes no sense for now (but longterm might).

One thing worth keeping in mind early is how we emit selection-extending motions through the state machine.
We could grow new variants `<motion>Extend` or make actions carry intent `Motion { extend: bool }`.

Thanks for your effort!

schlich:
Hey @kronberger-droid from a high level that all sounds fine! I didn't really have a specific path planned. And I likely won't have much time until next week to get back on this so if you want to open a PR please have at it!

kronberger-droid:
Hey there, while i was trying to expand on #1061 and I hit a structural question:

`EditMode::parse_event(&mut self, event) -> ReedlineEvent` is stateless with respect to the buffer.
That means `HelixRange` sitting on `HelixMachine` (where #1061 puts it) cannot actually be maintained from inside `HelixMachine`:

- Helix's `put_cursor` semantics need **new cursor position after motion runs** which is computed by the engine, after `parse_event` has already returned.
- Word motions (`w/b/e`) compute target positions from buffer contents, which the edit mode doesn't have.

## Options

**1. Selection on `Editor`.** Add `helix_selection: Option<HelixRange>` to `Editor` and the engine maintains it via new `EditCommand` variants. Roughly what my `hx-mode` POC did. `HelixRange` relocates to `core_editor/`. Helix-specific logic ends up in or called from `engine.rs`.

**2. Opt-in buffer view via default method.** No signature break:

```rust
pub trait EditMode: Send {
    fn parse_event(&mut self, event: ReedlineRawEvent) -> ReedlineEvent;

    fn parse_event_with_buffer(
        &mut self,
        event: ReedlineRawEvent,
        _editor: &Editor,
    ) -> ReedlineEvent {
        self.parse_event(event)
    }
}
```

Engine calls `parse_event_with_buffer`. `Vi`/`Emacs` inherit the default. `Helix` overrides, reads the buffer, maintains `HelixMachine.selection` in place, emits `EditCommand`s as before. `EditCommand` stays clean; all helix logic stays in `src/edit_mode/helix/`, placement stays as is in #1061.

## Could help with #1059

The Vi word boundaries PR is related to this, since its workaround was flat `MoveViWord*` variants on `EditCommand` (option 1). With option 2, Vi computes word targets locally and emits primitive moves. The "flat vs parameterized" fork collapses: the logic doesn't belong in `EditCommand`.

## Questions to you

- **Is the default-method trait expansion acceptable?**
- **Any concern about `EditMode` impls reading `&Editor` directly** (vs the current intent-only contract)?
- **If both are on the table, which do you prefer?**

---

I lean towards the second option, since it makes it possible to have a more general `EditCommand` set. Right now its a mix of buffer primitives and Vi-flavored motions. If edit modes can read `&Editor`, some mode-specific logic moves into the mode itself which owns the answer.

kronberger-droid:
closed in favor of #1138

PR #963, opened 2025-10-15
docs: improve CONTRIBUTING.md to more closely match CI execution

schlich:
This pull request updates the development workflow instructions in `CONTRIBUTING.md` to ensure consistency with the project's CI checks and to help contributors catch issues earlier. The changes clarify and strengthen requirements for formatting, linting, testing, and lockfile validation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,30 +32,39 @@ We follow the standard rust formatting style and conventions suggested by [clipp
 
 ### To make the CI gods happy
 
-Before submitting a PR make sure to run:
+> The commands below reflect what CI runs. Not all flags are always necessary for local development (e.g. feel free to omit `--all-features` or `--all-targets` during iteration), but running these exact commands before opening a PR ensures you match CI behavior.
 
-- for formatting
+Before opening a PR, run these locally:
+
+- Format:
 
   ```shell
   cargo fmt --all
   ```
 
-- the clippy lints (with warnings treated as errors)
+  > `--all` formats all packages in the workspace.
+
+- Lint (Clippy):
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -37,17 +37,25 @@ Before submitting a PR make sure to run:
 - for formatting
 
   ```shell
-  cargo fmt --all
+  cargo fmt --all -- --check
```
sholderbach: As we want everyone to format the code we can provide the guideline to run the formatting. The format check in CI is just the verification of that.
```suggestion
  cargo fmt --all
```

CONTRIBUTING.md
```diff
@@ -37,17 +37,25 @@ Before submitting a PR make sure to run:
...
 
   ```shell
-  cargo test
+  cargo test --all
   ```
+
+- the lockfile check
+
+  ```shell
+  cargo check --locked --all-targets --all
```
sholderbach: Not sure how many times that is actually relevant? Have you run into that failing?

sholderbach (commented):
Thanks for improving that!

CONTRIBUTING.md
schlich: hi @sholderbach thanks for the feedback!  I think i was a bit thrown off by the github actions CI calling this command, as parity with CI was the priority for this PR.  It seems like clippy covers this behavior (checking that `Cargo.lock` matches `Cargo.toml`) and so i've edited the instructions with some consolidated commands and clarifying notes.  While I believe this should cover things as far as matching CI, I wonder if some of the steps in the gh workflow can be consolidated?

PR #965, opened 2025-10-24
Add ReedlineEvent::SubmitOrSpace

zknx:
This PR solves https://github.com/nushell/nushell/issues/9227 without changing the default behavior.

## Context
The issue reports that when tab completion is active, pressing space doesn't accept the currently selected option (unlike pressing Enter). Instead, the completion menu persists, which differs from the expected behavior in most shells.

I've been struggling with this issue myself. While it seems most users haven't encountered it, I'll leave it up to you whether to approve this change.

## What this PR does
With this fix, pressing space during tab completion will accept the current selection and close the completion menu.
Example with nushell:

```nushell
$env.config.keybindings ++= [
  {
    name: submit_or_space
    modifier: none
    keycode: space
    mode: [emacs vi_normal vi_insert]
    event: [
      {send: SubmitOrSpace}
    ]
  }
]
```

```sh
touch a1 a2
cat a<tab><space>
```
will update the buffer to:
```sh
cat a1 [cursor here]
```

similar to the behavior in zsh or fish or pwsh.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1183,7 +1183,7 @@ impl Reedline {
                     }
                 }
                 self.run_edit_commands(&[EditCommand::InsertChar(' ')]);
-                return Ok(EventStatus::Handled);
+                Ok(EventStatus::Handled)
             }
             ReedlineEvent::ExecuteHostCommand(host_command) => {
                 self.suspended_state = Some(self.painter.state_before_suspension());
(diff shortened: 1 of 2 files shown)


zknx:
Ok, if I enabled this, it breaks like
```
$ cat "b<tab><space>2<enter>
```
cz I cannot input `<space>` character after I press `<tab>` key.
So it breaks something and shouldn't become default behavior.

kronberger-droid:
Closing in favour of #1203, which adds `MenuAccept` as a plain event.
Your space binding becomes `Multiple([MenuAccept, InsertChar ' '])`, so the accept and the space stay separate
and no `Submit` variant has to carry the menu logic.

Thanks for digging into nushell#9227.

PR #967, opened 2025-10-25
feat!: new `ReedlineEvent::PartialComplete` event

### LINKED ISSUE #14152: tab completion on suggestions stack on the second element
### Describe the bug

When I hit tab, I always get the second completion result, though `shift+tab` works correctly. The problem is with paths completions and with commands completions. 
With the history menu `ctrl + r` there is no problem.

It's a strange bug that appeared only in 0.99.x. 
Before that on the same config `tab` for given cases worked correctly.

https://github.com/user-attachments/assets/ce26f58f-16e8-474e-a1f2-3bb093fb24cf

On the same machine in `nu -n` tab works normally. 

On the discord @fdncred pointed out these two commits:

- https://github.com/nushell/reedline/pull/835
- https://github.com/nushell/reedline/pull/828

I tried to disable and rename the config shortcut (below), but with no success.

```
    {
        name: completion_next_menu
        modifier: none
        keycode: tab
        mode: [emacs vi_normal vi_insert]
        event: {
            until: [
                { send: menu name: completion_menu }
                { send: menunext }
                { edit: complete }
            ]
        }
    }
```

### How to reproduce

enter some file path - click tab more than once. 
@fdncred wasn't able to reproduce the bug both on windows and on Mac.

### Expected behavior

I expect `tab` to scroll through selections

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.99.1                                   |
| major              | 0                                        |
| minor              | 99                                       |
| patch              | 1                                        |
| branch             |                                          |
| commit_hash        | 8812072f065be283df3e9a81f3945d0364f2f247 |
| build_os           | macos-aarch64                            |
| build_target       | aarch64-apple-darwin                     |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10)      |
| rust_channel       | 1.79.0-aarch64-apple-darwin              |
| cargo_version      | cargo 1.79.0 (ffa9cf99a 2024-06-03)      |
| build_time         | 2024-10-22 18:30:13 -03:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash                   |
| installed_plugins  | polars 0.99.0                            |

blindFS:
Move the partial completion logic from the handling of `ReedlineEvent::MenuNext`, so that we can fix https://github.com/nushell/nushell/issues/14152 while keeping the following behavior.

> When the only valid options are say `foo` and `foooo`, a typical bash behavior of `f#tab` would complete the repl text to `foo` without moving the menu item towards `foooo`.

Closes #881

### DIFF
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -912,6 +912,9 @@ pub enum ReedlineEvent {
 
     /// Change mode (vi mode only)
     ViChangeMode(String),
+
+    /// Complete the command given the common prefix of items in current open menu
+    PartialComplete,
 }
 
 impl Display for ReedlineEvent {
@@ -960,6 +963,7 @@ impl Display for ReedlineEvent {
             ReedlineEvent::ExecuteHostCommand(_) => write!(f, "ExecuteHostCommand"),
             ReedlineEvent::OpenEditor => write!(f, "OpenEditor"),
             ReedlineEvent::ViChangeMode(_) => write!(f, "ViChangeMode mode: <string>"),
+            ReedlineEvent::PartialComplete => write!(f, "PartialComplete"),
         }
     }
 }
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -904,6 +904,9 @@ pub enum ReedlineEvent {
     /// Move to the previous history page
     MenuPagePrevious,
 
+    /// Complete the command given the common prefix of items in current open menu
+    PartialComplete,
+
     /// Way to bind the execution of a whole command (directly returning from [`crate::Reedline::read_line()`]) to a keybinding
     ExecuteHostCommand(String),
 
@@ -912,9 +915,6 @@ pub enum ReedlineEvent {
 
     /// Change mode (vi mode only)
     ViChangeMode(String),
-
-    /// Complete the command given the common prefix of items in current open menu
-    PartialComplete,
 }
 
 impl Display for ReedlineEvent {
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -520,6 +520,7 @@ impl Menu for ColumnarMenu {
         if can_partially_complete(self.get_values(), editor) {
             // The values need to be updated because the spans need to be
             // recalculated for accurate replacement in the string
+            // TODO: recalculate the spans instead of calling the completer 1 more time
             self.update_values(editor, completer);
 
             true
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -514,16 +514,9 @@ impl Menu for ColumnarMenu {
     /// in the given line buffer
     fn can_partially_complete(
         &mut self,
-        values_updated: bool,
         editor: &mut Editor,
         completer: &mut dyn Completer,
     ) -> bool {
-        // If the values were already updated (e.g. quick completions are true)
-        // there is no need to update the values from the menu
-        if !values_updated {
-            self.update_values(editor, completer);
-        }
-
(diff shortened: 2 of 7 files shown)


src/menu/columnar_menu.rs
```diff
@@ -501,16 +501,9 @@ impl Menu for ColumnarMenu {
...
-        values_updated: bool,
         editor: &mut Editor,
         completer: &mut dyn Completer,
     ) -> bool {
-        // If the values were already updated (e.g. quick completions are true)
-        // there is no need to update the values from the menu
-        if !values_updated {
-            self.update_values(editor, completer);
-        }
-
```
blindFS, line 526: I'm not very sure that it's safe to remove this. But I think it's a bad idea to handle `quick completion` related menu updates here.

ysthakur: I get that it's not very nice to call `update_values` here, but can you guarantee that the suggestions will always be up-to-date at this point?

As far as I can tell, `values_updated` is only true when the event is `Menu` and quick completions are enabled

src/menu/mod.rs
```diff
@@ -106,7 +106,6 @@ pub trait Menu: Send {
     /// in the given line buffer
     fn can_partially_complete(
         &mut self,
-        values_updated: bool,
```
ysthakur, line 109: nit: this is a breaking change for anyone making their own menus, so if you're following the Conventional Commits spec, the PR title should begin with `feat!`, I think

src/enums.rs
```diff
@@ -833,6 +833,9 @@ pub enum ReedlineEvent {
 
     /// Change mode (vi mode only)
     ViChangeMode(String),
+
+    /// Complete the command given the common prefix of items in current open menu
+    PartialComplete,
```
ysthakur: nit: can you move this up next to the other menu events?

src/menu/columnar_menu.rs
blindFS: I think I've called `update_values` at the caller site right before this function whenever the `values_updated` are set to false.

I might need to double check it since my memory gets  faded.

src/menu/mod.rs
blindFS: Nice catch!

src/enums.rs
blindFS: Sure

src/menu/columnar_menu.rs
blindFS: There was indeed a wrong span issue with `quick_completions` set to false. I've fixed it in the last commit.

The code looks so bizarre, I feel there's something fundamentally wrong here... But it should be logically identical now. Please confirm it before merging.

ysthakur: Hmm, I see `update_values` being called in the tests, but not in the engine itself. I'll run this locally though, I just haven't had the time to do so yet

blindFS: > Hmm, I see `update_values` being called in the tests, but not in the engine itself. I'll run this locally though, I just haven't had the time to do so yet

NVM, it's not an urgent feature anyway.

fdncred:
are we ready to land this?

kronberger-droid:
@blindFS Whats the state of this? Are you fine landing in while its a long time until the next release?

blindFS:
> @blindFS Whats the state of this? Are you fine landing in while its a long time until the next release?

Actually I don't remember the details, but I think it's pretty risky to merge without reviewing, I'll mark it as draft for now in case anybody willing to solve the issue based on my attempt.

PR #969, opened 2025-10-25
fix: wrong answer of `find_common_string`

blindFS:
Fixes a bug described [here](https://github.com/nushell/nushell/pull/16765#issuecomment-3384411809).

Update:

Fixes https://github.com/nushell/nushell/issues/15535

### DIFF
--- a/src/menu/menu_functions.rs
+++ b/src/menu/menu_functions.rs
@@ -185,8 +185,7 @@ pub fn find_common_string(values: &[Suggestion]) -> Option<(&Suggestion, usize)>
                 .value
                 .char_indices()
                 .zip(current_suggestion.value.chars())
-                .find(|((_, lhs), rhs)| rhs != lhs)
-                .map(|((idx, _), _)| idx)
+                .find_map(|((idx, lhs), rhs)| (rhs != lhs).then_some(idx))
                 .unwrap_or(max_len);
             if new_common_prefix_len == 0 {
                 Done(0)


src/menu/menu_functions.rs
```diff
@@ -168,39 +172,29 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult<'_> {
...
             } else {
-                first
-                    .value
-                    .char_indices()
-                    .zip(suggestion.value.char_indices())
-                    .find(|((_, mut lhs), (_, mut rhs))| {
-                        lhs.make_ascii_lowercase();
-                        rhs.make_ascii_lowercase();
-
-                        lhs != rhs
```
blindFS, line 187: I think this should be case sensitive, correct me if I'm wrong.

src/menu/menu_functions.rs
```diff
@@ -345,15 +339,17 @@ pub fn replace_in_buffer(value: Option<Suggestion>, editor: &mut Editor) {
...
         let start = floor_char_boundary(editor.get_buffer(), span.start).min(end);
 
         // make sure that the partial completion does not overwrite user entered input
-        let extends_input = matching.starts_with(&editor.get_buffer()[start..end])
-            && matching != &editor.get_buffer()[start..end];
+        let entered_input = &editor.get_buffer()[start..end];
+        let extends_input = matching
+            .to_ascii_lowercase()
+            .contains(&entered_input.to_ascii_lowercase())
+            && matching != entered_input;
```
blindFS, line 352: And this should be case insensitive...

src/menu/menu_functions.rs
```diff
@@ -345,15 +339,17 @@ pub fn replace_in_buffer(value: Option<Suggestion>, editor: &mut Editor) {
...
         let end = floor_char_boundary(editor.get_buffer(), span.end);
         let start = floor_char_boundary(editor.get_buffer(), span.start).min(end);
 
         // make sure that the partial completion does not overwrite user entered input
-        let extends_input = matching.starts_with(&editor.get_buffer()[start..end])
-            && matching != &editor.get_buffer()[start..end];
+        let entered_input = &editor.get_buffer()[start..end];
+        let extends_input = matching
+            .to_ascii_lowercase()
+            .contains(&entered_input.to_ascii_lowercase())
```
sholderbach: Should we use proper Unicode case folding here? Also why do we need the change to the case sensitivity?

blindFS: Unicode case folding sounds like a better idea.

The case sensitivity changes are based on the following presumption:

1. If the candidates are `foo` and `Foo`, we shouldn't partially complete it to any of them.
2. If the candidates are `Foo`, `Fooo`, and the user input is `foo` (which implies case insensitive completion config). We should partially complete it to `Foo`.

You can check the new tests in columnar_menu.rs and menu_functions.rs

blindFS: ~Also the `contains` check seems too strict for fuzzy scenarios.~

~I mean maybe we don't need this extends_input? I suppose input matching is already done elsewhere.~

blindFS:
@ysthakur Would you please take a look at this when you get time?

Cargo.lock
ysthakur, line 1: Unrelated to this PR, but we should make a `.gitattributes` and add this as a generated file in there, not sure how we haven't done that until now

src/menu/menu_functions.rs
```diff
@@ -168,39 +173,28 @@ pub fn parse_selection_char(buffer: &str, marker: char) -> ParseResult<'_> {
...
 
 /// Finds index for the common string in a list of suggestions
-pub fn find_common_string(values: &[Suggestion]) -> (Option<&Suggestion>, Option<usize>) {
-    let first = values.iter().next();
-
-    let index = first.and_then(|first| {
-        values.iter().skip(1).fold(None, |index, suggestion| {
-            if suggestion.value.starts_with(&first.value) {
-                Some(first.value.len())
+pub fn find_common_string(values: &[Suggestion]) -> Option<(&Suggestion, usize)> {
```
ysthakur, line 176: Much more sensible return type

src/menu/columnar_menu.rs
```diff
@@ -822,6 +822,24 @@ mod tests {
...
+    }
+
+    partial_completion_tests! {
+        name: partial_completion_unicode_case_folding,
+        completions: ["ßar", "ßaz"],
+
+        test_cases:
+            partial_completes_case_insensitive: ("ss", "ßa"),
+    }
+
```
ysthakur, line 842: From https://github.com/nushell/reedline/pull/969#discussion_r2496571136:

> 1. If the candidates are foo and Foo, we shouldn't partially complete it to any of them.
> 2. If the candidates are Foo, Fooo, and the user input is foo (which implies case insensitive completion config). We should partially complete it to Foo.

Could you add tests for this? For the second, the user input `foo` would be completed to `Fooo`, but `f` should be partially completed to `Foo`

ysthakur (commented):
Thanks, LGTM, just need a couple more test cases

src/menu/columnar_menu.rs
```diff
@@ -822,6 +822,24 @@ mod tests {
...
 
+    // https://github.com/nushell/nushell/issues/15535
+    partial_completion_tests! {
+        name: partial_completion_with_quotes,
+        completions: ["`Foo bar`", "`Foo baz`"],
+
+        test_cases:
+            partial_completes_prefix_with_backtick: ("F", "`Foo ba"),
+            partial_completes_case_insensitive: ("foo", "`Foo ba"),
+    }
```
blindFS, line 833: @ysthakur the 2nd of [suggested test cases](https://github.com/nushell/reedline/pull/969/files#r2512872770) is just an easier version of this?

src/menu/menu_functions.rs
```diff
@@ -718,46 +714,23 @@ mod tests {
...
-    fn find_common_string_with_non_ansi() {
-        use crate::Span;
-
-        let input: Vec<_> = ["ｎｕｓｈｅｌｌ", "ｎｕｌｌ"]
+    #[rstest]
+    #[case::ascii(vec!["nushell", "null"], 2)]
+    #[case::non_ascii(vec!["ｎｕｓｈｅｌｌ", "ｎｕｌｌ"], 6)]
+    // https://github.com/nushell/nushell/pull/16765#issuecomment-3384411809
+    #[case::unsorted(vec!["a", "b", "ab"], 0)]
+    #[case::should_be_case_sensitive(vec!["a", "A"], 0)]
```
blindFS, line 722: @ysthakur and the 1st of [suggested test cases](https://github.com/nushell/reedline/pull/969/files#r2512872770) is covered by this?

src/menu/columnar_menu.rs
ysthakur: Whoops, didn't notice that, that's covered then

src/menu/menu_functions.rs
ysthakur: I did see this and dismissed it because `find_common_string` is pretty low in the call chain. But I guess it's unlikely that `find_common_string` will be passed lowercased suggestion values or something, so whatever, I guess this is good enough

src/menu/columnar_menu.rs
ysthakur: See comments below, @blindFS pointed out that this was already handled

PR #973, opened 2025-11-17
fix: Make ColumnarMenu suggestions always take 1 line

### LINKED ISSUE #972: Columnar completion menu can make prompt disappear in narrow terminals
**Platform** macOS
**Terminal software** ghostty

1. Image

2. Press tab in a narrow terminal window
3.  Image

Basically some items take more than 1 line, which may lead to wrong estimation of this for columnar menu:

https://github.com/nushell/reedline/blob/7888dbf7ed519dd8d0cab6146130e3c59c42324d/src/menu/columnar_menu.rs#L700-L702

ysthakur:
Fixes #972

## Description

When the longest suggestion in a columnar menu exceeds the screen width, all suggestions will take up more than 1 line. None of the suggestion values will be truncated, and even short suggestion values will be padded to the same length as the longest suggestion. This PR fixes that by truncating values to the screen width.

This PR is just a fix, so I didn't want to change behavior too much, but in the future, we might want to consider basing the padding not on the longest suggestion, but the longest *visible* suggestion (when there are descriptions). When there aren't descriptions, the number of columns would probably have to be based on the longest suggestion overall, since it might be confusing if you scroll down and the number of columns changes.

## Screenshots

Here's an example of how the values get truncated (note that these suggestions have descriptions, they're just not shown because the longest suggestion is too long):

image

When the longest suggestion isn't too long, you see the descriptions (in this PR, I changed it so that they end in `"..."` when truncated):
image

And this is what you get without descriptions:
image

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -874,8 +874,7 @@ mod tests {
         painter.handle_resize(terminal_size.0, terminal_size.1);
 
         menu.menu_event(MenuEvent::Activate(false));
-        menu.update_working_details(editor, completer, &mut painter);
-        // menu.update_values(editor, completer);
+        menu.update_working_details(editor, completer, &painter);
     }
 
     #[test]


src/menu/menu_functions.rs
```diff
@@ -468,6 +469,31 @@ pub fn get_match_indices<'a>(
     }
 }
 
+/// Truncate a string with no ANSI escapes to the given max width, which must be >=3.
+///
+/// If `s` is longer than `max_width`, the resulting string will end in "..."
+/// and have width at most `max_width`.
+pub(crate) fn truncate_no_ansi(s: &str, max_width: usize) -> Cow<str> {
```
ysthakur: It's named `truncate_no_ansi` because I intend to add a `truncate_with_ansi` helper that truncates strings that do contain ANSI escapes (so it must ignore the ANSI escapes)

src/menu/ide_menu.rs
```diff
@@ -506,17 +506,7 @@ impl IdeMenu {
...
-                .value
-                .chars()
-                .take(max_string_width.saturating_sub(3))
-                .collect::<String>();
-            chars.push_str("...");
-            chars
-        } else {
-            suggestion.value.clone()
-        };
+        let string = truncate_no_ansi(&suggestion.value, max_string_width);
```
ysthakur, line 509: I used the truncation helper here because the original code looks at number of chars rather than width

blindFS:
Nice! Works well for me.

src/painting/painter.rs
```diff
@@ -517,9 +517,12 @@ impl Painter {
         //
         // I assume this is a bug with the position() call but haven't figured that
         // out yet.
-        if let Ok(position) = cursor::position() {
-            self.prompt_start_row = position.1;
-            self.just_resized = true;
+        #[cfg(not(test))]
```
ysthakur, line 520: Couldn't figure how to mock the `Painter` struct with `mockall` so I figured this was easier

WindSoilder (approved):
Generally the implementaion looks good to me, let's land it.

PR #984, opened 2025-12-14
feat: handle alt key escapes in vi mode

### LINKED ISSUE #983: vi mode: use alt + key to execute normal mode
## Description
Allow executing (and entering normal mode) normal mode commands by holding down alt in insert mode (or any other modes for that matter).

In regular vim this works as usual, however isn't handled by reedline.

## Usecase
I commonly use this to go back to the previous command (with alt + j or alt + k), or go to the beginning with alt + I or delete the entire line and and go back to insert mode with alt + S

## References

bash's readline and zsh's line editor both do this by default.

tmuaz:
My implementation is a little crude, I'm open to any feedback to make changes to it before opening the PR.
fixes #983

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -75,9 +75,11 @@ impl Vi {
 impl EditMode for Vi {
     fn parse_event(&mut self, event: ReedlineRawEvent) -> ReedlineEvent {
         match event.into() {
-            Event::Key(KeyEvent {
-                code, modifiers, ..
-            }) => match (self.mode, modifiers, code) {
+            Event::Key(
+                key_event @ KeyEvent {
+                    code, modifiers, ..
+                },
+            ) => match (self.mode, modifiers, code) {
                 (ViMode::Normal, KeyModifiers::NONE, KeyCode::Char('v')) => {
                     self.cache.clear();
                     self.mode = ViMode::Visual;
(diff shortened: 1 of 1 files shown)


src/edit_mode/vi/mod.rs
```diff
@@ -134,23 +138,28 @@ impl EditMode for Vi {
                     self.insert_keybindings
                         .find_binding(modifier, KeyCode::Char(c))
                         .unwrap_or_else(|| {
-                            if modifier == KeyModifiers::NONE
```
blindFS, line 137: I don't understand this part, with or without your changes. Do you have any clue?

highjeans: As far as I can tell, it seems to be handling inputs with key modifiers. His change checks for alt at the beginning and handles it like vim does there, then does the other checks reedline does already.

tmuaz: It seems to just be checking if it can enter that key as a regular input (not handling cases where just the control modifier is held) and also making the key uppercase if shift is held.

My function does "consume" alt inputs (by flipping the alt bit) so checking if alt is pressed in that list is a bit redundant hence the comment.

blindFS: I mean what's so special about `ctrl + alt` and `ctrl + alt + shift`? Why not just check whether shift is pressed? The comments before make me even more confused...

tmuaz: honestly not entirely sure but I've decided to keep the original logic in-case it's just something I don't understand

kronberger-droid:
superseded by #1207

PR #986, opened 2025-12-16
fix: move cursor left when exiting vi insert mode with Escape

andrewgazelka:
## Summary
- In vim, pressing Escape to exit insert mode moves the cursor left by one position
- This is because insert mode positions the cursor between characters, while normal mode positions it on a character
- Without this fix, repeatedly pressing `i <esc> i <esc>` would not move the cursor backward as expected

## Test plan
- [x] Added test `esc_from_insert_mode_moves_cursor_left` 
- [x] Updated existing test to explicitly test Esc from normal mode
- [x] All 756 tests pass

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -156,8 +156,20 @@ impl EditMode for Vi {
                 }
                 (_, KeyModifiers::NONE, KeyCode::Esc) => {
                     self.cache.clear();
+                    let was_insert = self.mode == ViMode::Insert;
                     self.mode = ViMode::Normal;
-                    ReedlineEvent::Multiple(vec![ReedlineEvent::Esc, ReedlineEvent::Repaint])
+                    // In vim, exiting insert mode moves cursor left by one position
+                    // because insert mode cursor is between characters while normal mode
+                    // cursor is on a character
+                    if was_insert {
+                        ReedlineEvent::Multiple(vec![
+                            ReedlineEvent::Edit(vec![EditCommand::MoveLeft { select: false }]),
+                            ReedlineEvent::Esc,
+                            ReedlineEvent::Repaint,
+                        ])
+                    } else {
+                        ReedlineEvent::Multiple(vec![ReedlineEvent::Esc, ReedlineEvent::Repaint])
+                    }
                 }
                 (ViMode::Normal | ViMode::Visual, _, _) => self
                     .normal_keybindings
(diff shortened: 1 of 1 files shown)


blindFS:
Thanks for trying to improve vi keybindings.

I see the point of aligning the behaviors of reedline vi-mode with those of vim/neovim.

But I don't think this specific one is very useful. It somehow reveals that we may have a bigger issue of wrong calculations of cursor positions within vi mode, see also #927. It would be better to fix them all-together with some fundamental changes/redesigning than this kind of case-wise tuning.

andrewgazelka:
Hmmm makes sense.

andrewgazelka:
> Thanks for trying to improve vi keybindings.
> 
> I see the point of aligning the behaviors of reedline vi-mode with those of vim/neovim.
> 
> But I don't think this specific one is very useful. It somehow reveals that we may have a bigger issue of wrong calculations of cursor positions within vi mode, see also #927. It would be better to fix them all-together with some fundamental changes/redesigning than this kind of case-wise tuning.

you want literally 1:1? what must be added

sim590:
I'd like to pick up this work and address the concerns raised here.

First, I want to clarify the Vi semantics at play: in Vim, pressing Escape to leave insert mode **always** moves the cursor one position left. This is not a cosmetic detail — it's fundamental to how Vi works. Insert mode places the cursor *between* characters; normal mode places it *on* a character. This is standard behavior in Vim, Neovim, and every Vi-compatible editor.

Regarding the regression in #772 (prefix history search breaking after `k` in normal mode): I understand that some users may want to keep prefix-based history search available. This feature already works with the up arrow in insert mode, so there's no loss of functionality. The `k` key behavior in normal mode should not be a blocker for fixing a real Vi bug.

That said, I'm also willing to adjust the `is_cursor_at_buffer_end()` logic so that prefix history search still works in Vi normal mode — by treating "cursor on the last character" as equivalent to "cursor at end" when in normal mode. This way we get correct Vi behavior without regressing the prefix search feature.

Would this approach be acceptable?

blindFS:
@sim590 thanks for your efforts, but I'm afraid your fix shares the same case-wise tuning nature with this PR. Like commented above, we'd favor a more systematic solution because people come with patches here and there without studying the whole picture of the underlying problem, making the code harder and harder to maintain. Hope you understand the difficult choice. Of course it will be great if you can dig deeper into the vim-mode implementation and make it right in a more general sense.

sim590:
@blindFS Thanks for the honest feedback. I've dug deeper and opened #1066 with a root cause analysis and two proposed systematic approaches. Would appreciate your thoughts there.

kronberger-droid:
fixed in #1100 + #1101

PR #989, opened 2025-12-23
feat: add initial support for multi-stroke chord keybindings in emacs mode

reubeno:
Adds basic support for adding multi-stroke (chord) keybindings, e.g.:

```rust
// Add chord bindings
// Ctrl+X Ctrl+C: Quit
keybindings.add_sequence_binding(
    &[ctrl('x'), ctrl('c')],
    ReedlineEvent::CtrlD,
);
```

Changes to public API:

* _(Breaking change)_ Remove the existing `pub` field of `Keybindings` so we can unify the implementation between single-keystroke bindings and multi-keystroke bindings.

  I did some [searching on GitHub](https://github.com/search?q=%22.keybindings%22+%22reedline%22+lang%3Arust+path%3A**%2F*.rs+%28NOT+is%3Afork%29+%28NOT+is%3Aarchived%29+-repo%3Anushell%2Fnushell+-repo%3Anushell%2Freedline&type=code) and couldn't immediately find specific cases that would be broken by this -- but of course, that doesn't prove it won't break a client. (Most of the search hits that came up seemed to be vendored/forked copies of `reedline` or `nushell`, or similar field names in different types.)
* Adds `Keybindings::add_sequence_binding()` and `Keybindings::remove_sequence_binding()` to bind (or unbind) a _sequence_ of 1 or more `KeyCombination`s to a `ReedlineEvent`; this supports both single-keystroke bindings and multi-keystroke chord-style bindings.
* Retains `Keybinding::add_binding()` and `Keybinding::remove_binding()` for backwards compatibility, replacing their bodies with simple wrapping calls to the new sequence-enlightened calls.

General approach:

* Do not allow a registered keystroke to be a strict prefix of another registered keystroke.
* Support arbitrarily long chords.
* Use a trie to store key bindings for efficient matching.
* Unify handling of single-keystroke keybindings with multiple-keystroke keybindings (do not special-case one vs. the other).
* Focus on `emacs` mode first; do not worry about aligning with existing multi-keystroke processing already present in vi mode for handling commands.

I'm looking for feedback on the general approach as well as the implementation before getting any deeper into this. For reference, I've demonstrated how this support could be used in `nushell` via [this commit in my `nushell` fork](https://github.com/reubeno/nushell/commit/ae4f46c2000459b767788f426d748c3f97269947). There's also a new example included in this draft PR to demonstrate and test.

_(Aside: This is my first feature-level contribution. I'm eager to get as much feedback as I can!)_

### DIFF
--- a/examples/chords.rs
+++ b/examples/chords.rs
@@ -1,0 +1,81 @@
+// Example demonstrating multi-keystroke chord support in reedline
+//
+// This example shows how to configure key chords (multi-key sequences)
+// that trigger actions. For example, pressing Ctrl+X followed by Ctrl+C
+// can be bound to a specific event.
+
+use crossterm::event::{KeyCode, KeyModifiers};
+use reedline::{
+    default_emacs_keybindings, DefaultPrompt, Emacs, KeyCombination, Reedline,
+    ReedlineEvent, Signal,
+};
+use std::io;
+
+/// Helper to create a KeyCombination for Ctrl+<char>
+fn ctrl(c: char) -> KeyCombination {
+    KeyCombination {
+        modifier: KeyModifiers::CONTROL,
+        key_code: KeyCode::Char(c),
+    }
+}
+
+fn main() -> io::Result<()> {
+    println!("Reedline Chord Example");
+    println!("======================");
+    println!();
+    println!("This example demonstrates multi-keystroke chord bindings.");
+    println!();
+    println!("Available chords:");
+    println!("  Ctrl+X Ctrl+C - Quit");
+    println!("  Ctrl+X Ctrl+Y Ctrl+Z Ctrl+Z Ctrl+Y - Report that nothing happens");
+    println!();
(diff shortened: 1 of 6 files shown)


kronberger-droid:
Nice foundation.
Heads up that @benvansleen's #1016 is solving the same chord problem from the other end and I think the trie you put in Keybindings is the cleaner base of the two, so it'd be great to join forces.

A couple of things #1016 already handles that this is missing:
a timeout flush (needed for jj-in-insert, where a lone j has to self-insert after a pause), and the case where a key is both a binding and a prefix (j→Down and jj→Esc, which the trie currently disallows).
Could the two be combined... your trie as storage, his resolver + timeout on top as one shared piece the modes embed?

Up for coordinating?

kronberger-droid:
Closing this, design discussions are now at #1204.
If Anyone wants to pick it up feel free.

PR #992, opened 2025-12-29
fix: reset ANSI styling before printing prompt

majiayu000:
## Summary
Reset any ANSI styling that may have been left by external commands before printing the prompt. This ensures the prompt is not affected by previous output styling that was not properly reset.

## Problem
When external commands output text with ANSI styling but don't include a reset sequence, the styling continues to affect the prompt display. This was reported in nushell/nushell#16384.

## Solution
Added ANSI reset (`SetAttribute(Attribute::Reset)` and `ResetColor`) at the beginning of `repaint_buffer()` function to ensure any leftover styling is cleared before the prompt is rendered.

## Test Plan
- All existing tests pass
- Manual testing: run commands that output styled text without reset, verify prompt displays correctly

Fixes nushell/nushell#16384

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -210,9 +210,8 @@ impl Painter {
     ) -> Result<()> {
         // Reset any ANSI styling that may have been left by external commands
         // This ensures the prompt is not affected by previous output styling
-        self.stdout
-            .queue(SetAttribute(Attribute::Reset))?
-            .queue(ResetColor)?;
+        // Note: Attribute::Reset (SGR 0) resets all attributes including colors
+        self.stdout.queue(SetAttribute(Attribute::Reset))?;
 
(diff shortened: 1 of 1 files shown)


src/painting/painter.rs
```diff
@@ -208,6 +208,12 @@ impl Painter {
         use_ansi_coloring: bool,
         cursor_config: &Option<CursorConfig>,
     ) -> Result<()> {
+        // Reset any ANSI styling that may have been left by external commands
+        // This ensures the prompt is not affected by previous output styling
+        self.stdout
+            .queue(SetAttribute(Attribute::Reset))?
+            .queue(ResetColor)?;
```
ysthakur: Are you sure `SetAttribute(Attribute::Reset)` isn't enough? Is `ResetColor` necessary?

fdncred:
Thanks!

PR #993, opened 2025-12-30
feat: add auto_complete_menu option for IDE-style completions

oxysoft:
## Summary
- Adds `with_auto_complete_menu(enable: bool)` builder method to Reedline
- When enabled, completion menu appears automatically as you type (like fish/IDEs)
- Menu activates after 1+ character typed for any argument
- Menu deactivates on space (waiting for next word)
- First word (commands) never auto-completes
- Bonus: adds `CompletionHinter` for inline gray-text suggestions

## Motivation
Many users coming from fish shell expect completions to appear automatically without pressing Tab. This provides an opt-in way to enable that behavior.

## Test plan
- [x] Build with `cargo build`
- [x] Tested in nushell with `$env.config.completions.auto_menu = true`
- [x] Verified menu appears for flags (`ls -`) and files (`ls d`)
- [x] Verified menu deactivates on space (`ls --all `)
- [x] Verified no scroll jank (menu only activates once per word)

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -143,6 +143,9 @@ pub struct Reedline {
     hinter: Option<Box<dyn Hinter>>,
     hide_hints: bool,
 
+    // Auto-show completion menu as you type (IDE/fish style)
+    auto_complete_menu: bool,
+
     // Use ansi coloring or not
     use_ansi_coloring: bool,
 
@@ -233,6 +236,7 @@ impl Reedline {
             visual_selection_style,
             hinter,
             hide_hints: false,
+            auto_complete_menu: false,
             validator,
             use_ansi_coloring: true,
             cwd: None,
@@ -323,6 +327,16 @@ impl Reedline {
     #[must_use]
     pub fn disable_hints(mut self) -> Self {
         self.hinter = None;
+        self
+    }
+
+    /// Enable auto-showing completion menu as you type (IDE/fish style)
+    ///
+    /// When enabled, the completion menu named "completion_menu" will automatically
+    /// appear as you type, without needing to press Tab.
+    #[must_use]
+    pub fn with_auto_complete_menu(mut self, enable: bool) -> Self {
+        self.auto_complete_menu = enable;
         self
     }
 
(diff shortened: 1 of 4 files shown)


blindFS:
A potentially stupid question: I don't see any `ide_menu` specific change, does this feature work for columnar completion menu?

Or, by `IDE-style`, you mean something unrelated to `ide_menu`?

src/engine.rs
```diff
@@ -1209,11 +1223,65 @@ impl Reedline {
...
+                    let should_show = if let Some(last_word_start) = buffer.rfind(' ') {
+                        let last_word = &buffer[last_word_start + 1..];
+                        // Show if: flags (starts with -) OR has 1+ char typed
+                        !last_word.is_empty()
+                    } else {
+                        // First word - don't auto-complete commands
+                        false
+                    };
+
+                    if let Some(menu) = self.menus.iter_mut().find(|m| m.name() == "completion_menu") {
```
ysthakur, line 1261: Not a fan of the menu name being hardcoded here. I'd rather `auto_complete_menu` be an `Option<String>` holding the name of the menu to activate automatically. That way, users can choose which menu they want to always be active.

src/engine.rs
```diff
@@ -143,6 +143,9 @@ pub struct Reedline {
     hinter: Option<Box<dyn Hinter>>,
     hide_hints: bool,
 
+    // Auto-show completion menu as you type (IDE/fish style)
+    auto_complete_menu: bool,
```
ysthakur, line 147: I think this name can confuse people. `auto_complete_menu` sounds like it's a menu for autocompletion, which all the menus are for. I'd prefer `always_on_menu` or something like that

src/engine.rs
```diff
@@ -143,6 +143,9 @@ pub struct Reedline {
     hinter: Option<Box<dyn Hinter>>,
     hide_hints: bool,
 
+    // Auto-show completion menu as you type (IDE/fish style)
```
ysthakur, line 146: You say "IDE" here, but looking at your code, it won't necessarily pick the `IdeMenu`, it'll pick whichever menu is named `completion_menu`. I think you just happened to see the IDE menu because that's what you have set in your Nushell config.

But also, let's not restrict this feature to the IDE menu. I prefer the columnar menu myself and would like to use that. So you can just remove the "(IDE/fish style)" bit from this comment

src/engine.rs
```diff
@@ -1209,11 +1223,65 @@ impl Reedline {
...
                     } else {
                         menu.menu_event(MenuEvent::Edit(self.quick_completions));
                     }
+                } else if self.auto_complete_menu {
+                    // Auto-activate completion menu (IDE style)
+                    // - Flags: show immediately on "-" or "--"
+                    // - Files/args: show after 1+ chars typed (not just space)
+                    // - Commands (first word): don't auto-complete
+                    let buffer = self.editor.line_buffer().get_buffer();
+                    let should_show = if let Some(last_word_start) = buffer.rfind(' ') {
```
ysthakur, line 1252: I'd prefer to make it so that the menu is *always* active, even if it's after a space. Let's leave it up to the menu and completer to decide whether to show anything.

I know this means users will see "NO RECORDS" all the time, but I'm sure we can work around that (or just let it be).

@blindFS thoughts?

src/engine.rs
```diff
@@ -1209,11 +1223,65 @@ impl Reedline {
...
+                                );
+                            } else {
+                                // Menu already active, just send edit event to update
+                                menu.menu_event(MenuEvent::Edit(self.quick_completions));
+                                menu.update_values(
+                                    &mut self.editor,
+                                    self.completer.as_mut(),
+                                    self.history.as_ref(),
+                                );
+                            }
```
ysthakur, line 1279: Nitpick (pardon the indentation)
```suggestion
                            if !menu.is_active() {
                                // Only activate if not already active
                                menu.menu_event(MenuEvent::Activate(self.quick_completions));
                            } else {
                                menu.menu_event(MenuEvent::Edit(self.quick_completions));
                              }
                              menu.update_values(
                                  &mut self.editor,
                                  self.completer.as_mut(),
                                  self.history.as_ref(),
                              );
```

ysthakur (changes requested):
Thanks for working on this!

I like the idea of a `CompletionHinter`, but it seems unrelated to the `auto_complete_menu` change and should be split into a separate PR

src/engine.rs
blindFS: It's a bit tricky. 

On one side, there should be a configurable min-triggering-characters to lower the frequency.

However spaces are sometimes meaningful characters for `fuzzy`/`substring` matcher.

Seems to me a pure reedline solution won't be "smart" enough in that sense.

ysthakur: It's not just for the fuzzy/substring matchers, I'm thinking about a scenario where someone has a command `foo` that can only take one fruit as an argument: `apple` or `banana`. So if I type `foo <TAB>`, I want to see the options apple and banana, even if I haven't typed anything yet

And yeah, it's out of scope for Reedline to guess how a completer works. The only time that disabling the menu is probably safe is when the user hasn't typed anything at all yet (empty buffer)

src/engine.rs
tobiasBora: I think this comment just explain what the author means by "as you type", since in IDE like vscode completion is shown as you type.

blindFS: > I think this comment just explain what the author means by "as you type", since in IDE like vscode completion is shown as you type.

The thing is that "IDE" in reedline context has sort of [special meaning](https://github.com/nushell/reedline/blob/main/src/menu/ide_menu.rs).

tobiasBora: Surely, but given that it appears next to `fish` I don't think it relates to this internal meaning. Nevermind it's just a comment.

blindFS: > Surely, but given that it appears next to `fish` I don't think it relates to this internal meaning. Nevermind it's just a comment.

I'd prefer precise comments.

fdncred:
closed in favor of https://github.com/nushell/reedline/pull/1118. After that PR, if there are differences that we still need from this one, they can be added in a separate PR.

tobiasBora:
@fdncred I'm sorry, but why do you close this one? The merged feature has nothing to do with auto complete always on as far as I understand. At most shouldn't we just say that this PR must be updated to take into account the new option added in #1118 ?

fdncred:
because it seems like the same feature to me, is it not?

tobiasBora:
It's not. Here the goal is to ALWAYS show the completion menu. In the linked PR, they just fix a "bug" that was removing the completion menu when pressing the backspace key. In the thread the author itself says that it is unrelated to the current PR.

fdncred:
I don't see the author responding to anything. Lots have changed in reedline since the end of last year. Probably best to resubmit on the latest main if the author is still around.

tobiasBora:
> I don't see the author responding to anything. 

I was refering to https://github.com/nushell/reedline/pull/1118#issuecomment-4986736474

> In this PR we added a setting that makes reedline keep the previously spawned menu open when all the characters entered before are deleted. So it has nothing to do with automatic activation.

fdncred:
agreed. I'm saying the author still cares nothing about this pr because they haven't responded to any of the comments.

Issue #995, opened 2026-01-02
IDE completion type has noticable seizure risk


alicealysia:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

when scrolling through completions in my terminal with the ide type enabled, I wound up experiencing an absence seizure and wound up with a migraine for the rest of the day.

This is primarily due to the way that the description and completion menu move around the screen by default. Often failing to resize and reposition in a sensible way, and instead switching positions such that you wind up with a completion menu extremely far away from the cursor, and a description to the left, despite there being plenty of space to position them.

this issue can be resolved immediately by setting the following values:
```
    completion_width: 33
    description_width: 33
```

Can these please be made the defaults?

Some other things worth considering that would both further reduce the risk, and also make this look nicer:

Allow users to set the background color for an entire row, and the entire prompt. By having the color bar be a constant width, it's not wiggling about in a semi random pattern.

Add an option to reserve vertical space either at the end of the viewport, or before the prompt so that triggering the menu doesn't shift the viewport.

### How to reproduce

1. enable the ide completion menu layout.
```
$env.config.menus ++= [{
  name: completion_menu
  only_buffer_difference: false
  marker: "| "
  type: {
    layout: ide
  }
  style: {
  }
}]
```
2. press tab
3. For maximum effect zoom in (ctrl++), you can even add some extra color to the different completions if you wanna get real spicy with it.
4. hold the down arrow.

You'd also need to be photosensitive ofc, not sure how easy it is to reproduce that one.

### Expected behavior

I  expect reedline not to be the first seizure inducing line editor?

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.109.1                                                                                   |
| major                | 0                                                                                         |
| minor                | 109                                                                                       |
| patch                | 1                                                                                         |
| branch               | makepkg                                                                                   |
| commit_hash          | cde5f7be8c813173e95ea65f96e79a133f587de7                                                  |
| build_os             | linux-x86_64                                                                              |
| build_target         | x86_64-unknown-linux-gnu                                                                  |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07) (Arch Linux rust 1:1.91.1-2)                          |
[...]

PR #996, opened 2026-01-04
fix(IdeMenu): Don't make description position depend on description width

ysthakur:
Goes towards fixing https://github.com/nushell/reedline/issues/995 (description box can change sides for every suggestion with the IdeMenu, causing seizures for photosensitive users)

This PR makes it so that which side the description box is on is not influenced by the description itself (only applicable when the `description_mode` is `PreferRight`, the default). This means that when scrolling down the list of completions, it will never change sides. The PR also fixes a bug where, if the description menu was on the left, it would sometimes push the completions box to the right, causing more motion.

### Breaking change

However, a breaking change caused by this is the change in the meaning of `min_description_width`. Previously, it was set to 0 by default, and now I'm setting it to 15 (arbitrary choice, open to changing it). Opened https://github.com/nushell/nushell/pull/17280 to update this on the Nushell side. I'm unsure what exactly its purpose was before, but it's now used as the threshold for deciding when to switch the description box to the left.

For example, if `min_description_width` is 15, then as long as we have 15 columns available on the right of the completions box, the description box can stay on the right. Once we have 14 or fewer columns available, the description box switches to the left. <s>Setting `min_description_width` to 0 would effectively force the description box to stay on the right.</s>

### Recordings

Here's a recording of what the default IdeMenu looks like in Nushell (warning: I decreased the playback speed but the recording includes scrolling down the menu, not sure if that can trigger seizures. Recommendation from alicealysia is to decrease window size first if you're photosensitive):

[asciicast](https://asciinema.org/a/Xp5dRxilcvWxy0slaORyLiXFk?t=10)

And here it is with `correct_cursor_pos: true`:
[asciicast](https://asciinema.org/a/Co0PA6qM9KsDBoMRYg4muQorh?t=3)

### DIFF
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -721,11 +721,13 @@ impl Menu for IdeMenu {
             self.working_details.completion_width = completion_width;
 
             // Columns at which completion box begins
-            let mut completion_pos = self
+            let mut completion_pos = (self
                 .working_details
                 .cursor_col
-                .saturating_add_signed(self.default_details.cursor_offset)
-                .saturating_sub(total_border_width / 2);
+                .saturating_sub(total_border_width / 2)
+                as i16
+                + self.default_details.cursor_offset)
+                .max(0) as u16;
 
             if self.default_details.correct_cursor_pos {
                 let base_string = &self.working_details.shortest_base_string;


alicealysia:
Re the warning, yeah, slowing down playback helps quite a bit, something extra that I will advise here is that viewers with photo-sensitivity should still reduce the window size if using a larger screen for the sake of saftey, but I was able to watch comfortably.

fdncred (approved):
Nice work Yash! Looks like lots of fiddly bits to get this working. I'm fine with moving forward with this and dogfooding it a for a bit in nushell.

Issue #997, opened 2026-01-04
Chinese (Full-width) Character Width Issue in IDE Layout Mode


IvolliousElectrolux:
**Platform** Windows 11
**Terminal software** Windows Terminal Preview 1.24.2682.0

Describe the problem you are observing.
When there are full-width characters in the file/directory names, the completion-menu's ide layout was broken.

## Steps to reproduce
1. change the layout of completion_menu to ide:
```
$env.config = {
    menus: [
        {
            name: completion_menu
            only_buffer_difference: false
            marker: "| "
            type: {
                layout: ide
            }
            style: {
                text: green
                selected_text: green_reverse
                description_text: yellow
            }
        }
    ]
}
```
2. cd into a new directory
3. create some files or directories with multiple lengths of Full-Width charaters, such as 你好.txt 我很好.txt 哈哈haha.txt, you can also create some files named only in english (half-full-width) to compare.
4. type .\ then type Tab, then you'll see the layout problem

## Screenshots/Screencaptures

![Image](https://github.com/user-attachments/assets/573cabfd-487b-4503-afc9-aa853998ee11)

![Image](https://github.com/user-attachments/assets/3a4be6e1-912e-4a98-9887-78bc91d6886c)

ysthakur:
@IvolliousElectrolux No need to apologize, you didn't do anything wrong, I made that PR *after* you created this issue anyway.

Also, let's keep this issue open until the PR is merged. Oftentimes there are delays with merging PRs, or they get cancelled entirely.

ysthakur:
Could you see which branch of the Reedline fork you're using? The PR uses `ide-menu-width`, not `main`

IvolliousElectrolux:
yes it's On branch ide-menu-width

ysthakur:
Shoot, I was trigger-happy and accidentally merged the PR before remembering your comment. But I still can't reproduce the bug after the Reedline changes. This is what it looks like for me:

<img width="410" height="311" alt="Image" src="https://github.com/user-attachments/assets/38ae2faa-2826-4927-8a10-027420da71bb" />

Instead of `cargo build` and then `nu`, could you instead try `cargo run` in the folder where you cloned the `nushell`?

PR #998, opened 2026-01-04
fix(IdeMenu): Use width instead of bytes/chars

ysthakur:
Fixes https://github.com/nushell/reedline/issues/997

We were looking at number of bytes/chars to get the width of the suggestion values and base strings. This didn't work with multibyte characters. The fix is to use the Unicode width of these strings.

This is what the IDE menu looks like with suggestions containing `验` (a multibyte character)
```
~/reedline| he
             ╭────────────────────────────────────────────────╮
             │hello another very large option for hello wor...│
             │hello world 1                                   │
             │hello world 2                                   │
             │hello world another                             │
             │hello world reedline验                          │
             │hello world something                           │
             │hello world验验验                               │
             ╰────────────────────────────────────────────────╯
```

```
~/reedline| hello wo
                   ╭──────────────────────╮
                   │hello world 1         │
                   │hello world 2         │
                   │hello world another   │
                   │hello world reedline验│
                   │hello world something │
                   │hello world验验验     │
                   ╰──────────────────────╯
```

### DIFF
--- a/src/menu/ide_menu.rs
+++ b/src/menu/ide_menu.rs
@@ -496,7 +496,7 @@ impl IdeMenu {
             .unwrap_or_default();
 
         let padding_right = (self.working_details.completion_width as usize)
-            .saturating_sub(suggestion.value.chars().count() + border_width + padding);
+            .saturating_sub(suggestion.value.width() + border_width + padding);
 
         let max_string_width =
             (self.working_details.completion_width as usize).saturating_sub(border_width + padding);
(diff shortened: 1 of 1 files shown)


fdncred:
I'm fine with this but I always wonder with things like this, should it be width, unicode-width, or graphemes?

fdncred (approved):
@ysthakur yes, i just wanted to make sure that `width()` was at least `unicode-width`'s width and not something else. Your explanation of graphemes seems logical to me.

PR #1002, opened 2026-01-11
Allow overriding suggestion display value

ysthakur:
This PR adds an optional `display_override` field to `Suggestion`. If set, the string set in this field will be displayed to users rather than the actual `value` of the suggestion. However, when a suggestion is chosen, its `value` will be put into the buffer, not its `display_override`.

Idea came from [this comment by @qfel](https://github.com/nushell/reedline/pull/798#issuecomment-2249222249) on an older PR.

## Why?

Reason 1: On the Nushell side, we often have to provide completions wrapped in quotes. This can cause problems on the Reedline side. We [strip quotes](https://github.com/nushell/reedline/blob/6ceda50065fce791bb0b89c76fe1c45789ea0d5c/src/menu/columnar_menu.rs#L391) in the columnar and IDE menus to account for this when figuring out which substring to highlight. I'm not sure if the quote stripping is necessary anymore since we're doing substring matching anyway, but it doesn't handle escapes in strings. With `display_override`, we can show suggestions that don't contain any quotes or backslashes while still keeping the suggestions' `value`s the same.

Relatedly, [`find_common_string`](https://github.com/nushell/reedline/blob/6ceda50065fce791bb0b89c76fe1c45789ea0d5c/src/menu/menu_functions.rs#L177), which is used for partial completions, is unable to compare file names wrapped in quotes to file names not wrapped in quotes. But here, the best solution here would be to allow completers to provide their own partial completions, which I plan to implement soon. Right now, partial completions are still based on `value`, not display value

Reason 2: Allowing Nushell users to customize their completions. The new field allows including ANSI highlights, so we can now let users create custom/external completers that provide suggestions with `display_override`s styled however they want. They can get styled strings from some other source and provide those to Nushell.

## Testing

As an example, here's a custom completer that uses lolcat. I just used lolcat for a laugh, but the lolcat colors look surprisingly good:

https://asciinema.org/a/CTmdQR9nboLjm5zn?t=4

Completer:
```nu
def comp [] {
  let completions = ^ls crates/*/*/*.rs | lolcat -f | lines | each {|file|
    {
      value: ($file | ansi strip),
      display_override: $file
    }
  }
  {
    completions: $completions,
    options: {
      completion_algorithm: "fuzzy"
    }
  }
}
def foo [ arg: string@comp ] {}
```

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -565,7 +565,11 @@ impl Menu for ColumnarMenu {
         let (values, base_ranges) = completer.complete_with_base_ranges(&input, pos);
 
         self.values = values;
-        self.display_widths = self.values.iter().map(|sugg| sugg.display_value().width()).collect();
+        self.display_widths = self
+            .values
+            .iter()
+            .map(|sugg| sugg.display_value().width())
+            .collect();
         self.working_details.shortest_base_string = base_ranges
             .iter()
             .map(|range| {


blindFS:
> But here, the best solution here would be to allow completers to provide their own partial completions, which I plan to implement soon. Right now, partial completions are still based on value, not display value

I'm confused at this part. Do you mean display value based partial completion is not good enough?

PR #1004, opened 2026-01-12
Add small LSP client 

wvhulle:
This PR adds a small LSP client that can be constructed from a path to any LSP server binary.
The screenshots in this thread are demonstrations on how user interaction with this LSP client would look.

I am looking for feedback on how to show the diagnostics and auto-fixes. Do you have any suggestions?

image

Fix menu:

image

There is an example provided that uses the `nu-lint` LSP server because it has a lot of lints and fixes available and this shows how multiple violations are rendered vertically and horizontally. It is possible to use `nu-lsp` but `nu-lsp` has fewer rules, so the density of the user interface cannot be tested so much.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -4,7 +4,10 @@ use itertools::Itertools;
 use nu_ansi_term::{Color, Style};
 
 #[cfg(feature = "lsp_diagnostics")]
-use crate::lsp::{message_style, underline_style, Diagnostic, LspDiagnosticsProvider};
+use crate::lsp::{
+    format_diagnostic_messages, range_to_span, underline_style, DiagnosticSeverity,
+    LspDiagnosticsProvider,
+};
 use crate::{enums::ReedlineRawEvent, CursorConfig};
 #[cfg(feature = "bashisms")]
 use crate::{
@@ -1813,8 +1816,9 @@ impl Reedline {
         #[cfg(feature = "lsp_diagnostics")]
         if let Some(ref mut provider) = self.lsp_diagnostics {
             for diag in provider.diagnostics() {
-                let style = underline_style(diag.severity);
-                styled_text.style_range(diag.span.start, diag.span.end, style);
+                let severity = diag.severity.unwrap_or(DiagnosticSeverity::WARNING);
+                let span = range_to_span(buffer_to_paint, &diag.range);
+                styled_text.style_range(span.start, span.end, underline_style(severity));
             }
         }
 
@@ -1852,7 +1856,7 @@ impl Reedline {
 
         // Format diagnostic messages for display below the prompt
         #[cfg(feature = "lsp_diagnostics")]
-        let diagnostic_display = self.format_diagnostic_messages(prompt);
+        let diagnostic_display = self.get_diagnostic_display(prompt);
         #[cfg(not(feature = "lsp_diagnostics"))]
         let diagnostic_display = String::new();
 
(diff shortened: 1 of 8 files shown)


blindFS:
> > @wvhulle Would you like to add the code action feature to `nu-lsp` and reuse it for reedline instead of implementing another lsp server? Or is there any technical issue that blocks this kind of communication?
> 
> 
> 
> Isn’t nu-lsp an lsp server and this a small lsp client? Nu-lsp could be used as the real-time diagnostics provider (as lsp server) for this client instead of nu-lint. So you wouldn’t be bound to nu-lint with the current implementation and the actual lsp server is swappable.

That's what I meant, better to reuse what's already present.

> If I would add code actions to nu-lsp (although I don’t mind doing that) it would not be possible to have this type of real-time diagnostics and fixes in the reedline prompt. It’s a slightly different feature I think?

Why's that?

blindFS:
> Showing realtime diagnostics while typing a prompt in reedline requires more than just adding code actions to nu-lsp. They seem to be orthogonal projects.

I don't get it, what's the special source on the nu-lint side to make it real-time?

blindFS:
@wvhulle I think you misunderstood my point. I'm not saying we don't need this lsp-client on reedline side.

examples/lsp_diagnostics.rs
```diff
@@ -0,0 +1,85 @@
+//! Example demonstrating LSP diagnostics integration with reedline.
+//!
+//! This example spawns an LSP server (nu-lint) and displays diagnostics inline
+//! in real-time as you type. It also demonstrates the diagnostic fix menu.
+//!
+//! Run with: cargo run --example lsp_diagnostics --features lsp_diagnostics
+//!
+//! Prerequisites:
+//! - nu-lint must be installed and available in PATH with LSP support
```
blindFS: You mean nu-lint is not necessary and could be replaced with nu-lsp?

wvhulle: This particular example references nu-lint binary but it is easy to switch out and replace with nu-lsp.

blindFS: Fair enough, I think I've made myself clear. Please don't make this feature defaults to using a non-official lsp server.

wvhulle: I understand, however, the official LSP does not have many auto-fixes available (as far as I know), so `nu-lint` is kind of needed in the example to show how the fix menu feature works (with any LSP that has fixes).

blindFS: Feel free to enhance `nu-lsp` then.

I mean it's OK to leave this example depending on nu-lint. But don't make this whole feature that way.

wvhulle: > Feel free to enhance `nu-lsp` then.

I'd love to, but some people use nu-lint in CLI mode (locally or in CI), which seems to be out of scope for nu-lsp.

blindFS: > > Feel free to enhance `nu-lsp` then.
> 
> 
> 
> I'd love to, but some people use nu-lint in CLI mode (locally or in CI), which seems to be out of scope for nu-lsp.

As you said, it's swappable, right? People who favor `nu-lint` can config in their own ways. And you can continue working on it, but I don't see the beauty of it. I'd prefer an all-in-one and configurable `nu-lsp` and all people contributing to it instead of a divided community.

wvhulle: Whether nu-lsp or nu-lint are the default, or whether nu-lsp is the official LSP server seems a bit unrelated to this PR to me. No? This PR is just about adding a minimal LSP client in Reedline, which is unrelated to LSP servers. The screenshots however, should give an idea how it might look to the users when the LSP client is active.

blindFS: > Whether nu-lsp or nu-lint are the default, or whether nu-lsp is the official LSP server seems a bit unrelated to this PR to me. No? This PR is just about adding a minimal LSP client in Reedline, which is unrelated to LSP servers. The screenshots however, should give an idea how it might look to the users when the LSP client is active. 

If it's configurable, then yes, it's unrelated.
But as I pointed out, this comment said `nu-lint` is marked as prerequisite, that's why we're having this conversation.

wvhulle:
> Besides, the requests are blocking, right? I'd prefer a more responsive shell.

I'll make it non-blocking.

blindFS:
> > Besides, the requests are blocking, right? I'd prefer a more responsive shell.
> 
> 
> 
> I'll make it non-blocking.

That's cool, but would you like to wait for some review feedbacks before you move on?

blindFS:
> Of course, I can wait a bit. 

Thanks, I just want to make sure we all agree on the direction to go.

And maybe making the PR a bit easier to review.

fdncred:
We talked about this quite a bit yesterday in our meeting. Everyone agrees that this is a really cool example of what you can do with reedline. However, we just think this is too much of a distraction for users. For that reason, we're closing this PR. Thanks for your great work!

Closing this PR makes me wonder if reedline should have some type of plugin architecture to allow others to experiment and use cool tools like nu-lint without all that code being embedded into reedline. 🤔

PR #1014, opened 2026-01-20
Add `libc` feature to enable crossterm's libc backend

NiltonVolpato:
This adds a `libc` feature that enables crossterm's `libc` feature, allowing reedline to work in environments where rustix's raw syscall approach fails.

## Motivation

When running under x86 emulation (e.g., CheerpX/WebVM, QEMU user-mode), reedline fails with:

```
Reedline error: Invalid argument (os error 22)
```

The root cause is that rustix uses the `TCGETS2` ioctl via raw syscalls, which some emulators don't implement. See:
- https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/2133804

Crossterm has a `libc` feature that uses libc's `tcgetattr` instead, which falls back to `TCGETS` and works in these environments.

## Changes

- Adds `libc` feature that enables `crossterm/libc`

## Usage

```toml
reedline = { version = "...", features = ["libc"] }
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,6 +190,7 @@ dependencies = [
  "crossterm_winapi",
  "derive_more",
  "document-features",
+ "libc",
  "mio",
  "parking_lot",
  "rustix 1.1.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ external_printer = ["crossbeam"]
 sqlite = ["rusqlite/bundled", "serde_json"]
 sqlite-dynlib = ["rusqlite", "serde_json"]
 system_clipboard = ["arboard"]
+libc = ["crossterm/libc"]
 
 [[example]]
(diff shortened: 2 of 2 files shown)


joshka:
@NiltonVolpato I'm considering dropping the libc support from crossterm in the next release - this is the only place where it's used across all of github (i.e. feels like a super edge case). Any thoughts of getting the bugs fixed in the emulators instead of having to support a broken stack all the way up (or perhaps getting the falllback added to rustix instead of having to support libc in crossterm for this one library call?)

Edit: actually seems like rustix already has the fixes, so in theory this should already work on your end and this PR can probably be reverted at some point. See https://github.com/crossterm-rs/crossterm/pull/1087 for more details.

fdncred:
@kronberger-droid Let's remember to revert this once a new version of crossterm is published. I'll try to remember too. I'm watching 1087 mentioned above.

Thanks for the heads-up @joshka

PR #1015, opened 2026-01-20
feat: new feature idle_callback to allow to set callback function while waiting for input

eitsupi:
Add an `idle_callback` feature that allows applications to register a callback function that is invoked periodically while waiting for user input.

Some applications need to perform periodic work while the editor is waiting for input. Examples include:                       
                                                                                                                                 
- Processing GUI/window events for embedded graphical applications                                                             
- Updating progress indicators or spinners                                                                                     
- Handling background timers or scheduled tasks                                                                                
- Processing events from external systems                                                                                      
                                                                                                                                 
Currently, `read_line()` blocks until input arrives, making it impossible to handle such periodic tasks.
This feature provides a simple opt-in mechanism for applications that need this capability.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -27,13 +27,18 @@ use {
             FileBackedHistory, History, HistoryCursor, HistoryItem, HistoryItemId,
             HistoryNavigationQuery, HistorySessionId, SearchDirection, SearchQuery,
         },
-        painting::{Painter, PainterSuspendedState, PromptLines},
+        painting::{Painter, PainterSuspendedState, PromptLines, RenderSnapshot},
         prompt::{PromptEditMode, PromptHistorySearchStatus},
         result::{ReedlineError, ReedlineErrorVariants},
-        terminal_extensions::{bracketed_paste::BracketedPasteGuard, kitty::KittyProtocolGuard},
+        terminal_extensions::{
+            bracketed_paste::BracketedPasteGuard,
+            kitty::KittyProtocolGuard,
+            semantic_prompt::{Osc133ClickEventsMarkers, SemanticPromptMarkers},
+        },
         utils::text_manipulation,
-        EditCommand, ExampleHighlighter, Highlighter, LineBuffer, Menu, MenuEvent, Prompt,
-        PromptHistorySearch, ReedlineMenu, Signal, UndoBehavior, ValidationResult, Validator,
+        EditCommand, ExampleHighlighter, Highlighter, LineBuffer, Menu, MenuEvent, MouseButton,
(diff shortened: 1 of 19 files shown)


src/engine.rs
```diff
@@ -745,15 +758,33 @@ impl Reedline {
...
-                // If the `external_printer` feature is enabled, we need to
-                // periodically yield so that external printers get a chance to
-                // print. Otherwise, we can just block until we receive an event.
-                #[cfg(feature = "external_printer")]
-                if event::poll(EXTERNAL_PRINTER_WAIT)? {
+                // Determine if we need to poll (non-blocking) or can block on input.
+                // We need polling if external_printer or idle_callback is configured.
+                #[cfg(all(feature = "external_printer", feature = "idle_callback"))]
+                let poll_config = match (&self.external_printer, &self.idle_callback) {
+                    (Some(_), _) => Some(EXTERNAL_PRINTER_WAIT),
```
ysthakur: Seems like this means that even if you set an idle callback with its own interval, that internal will be overridden by `EXTERNAL_PRINTER_WAIT`. I think that's confusing behavior.

I'd prefer it if we had a separate, shared `poll_interval` setting that was used for both the external printer and idle callback (it can default to what's currently `EXTERNAL_PRINTER_WAIT`). If neither external printer nor idle callback are set, the poll interval can just be ignored.

eitsupi: Thanks, how about 00ed995?

src/engine.rs
```diff
@@ -57,10 +57,10 @@ const POLL_WAIT: Duration = Duration::from_millis(100);
...
 const EVENTS_THRESHOLD: usize = 10;
 
-/// Maximum time Reedline will block on input before yielding control to
-/// external printers.
-#[cfg(feature = "external_printer")]
-const EXTERNAL_PRINTER_WAIT: Duration = Duration::from_millis(100);
+/// Default maximum time Reedline will block on input before yielding control
+/// for features that require periodic processing (e.g., external printer,
+/// idle callback).
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_millis(100);
```
fdncred, line 68: What's the difference between this and POLL_WAIT above? I mean, do we need both?

eitsupi: IIUC, they serve different purposes:

- `POLL_WAIT` is used after receiving events. When paste or resize is detected (many events arriving at once), it　batches additional events together at the cost of a slight delay.
- `DEFAULT_POLL_INTERVAL` is used before any event arrives. It's the maximum time the main loop blocks waiting for the first input event, so that `external_printer`/`idle_callback` get a chance to run periodically.

They happen to have the same value (100ms) but operate at different phases of the event loop.

PR #1017, opened 2026-01-29
feat: Skim as a reedline completion menu

benvansleen:
I've always loved [FZF tab-completion in zsh](https://github.com/lincheney/fzf-tab-completion/tree/master). Can we achieve something similar in nushell?

It turns out: yes, we can! ([And we can configure nushell to allow for opting-into a `skim` menu layout type, too.](https://github.com/nushell/nushell/compare/main...benvansleen:nushell:feat/support-skim-menu-configuration))

skim-demo

I'm going to test-drive this for a bit and see if I find any rough areas. Otherwise, would love to work to get this upstreamed!

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -677,35 +677,9 @@ impl Reedline {
 
         let (result, clear_height) = f(self);
 
-        let mut reuse_prompt = true;
-        if let Some(clear_height) = clear_height {
-            if clear_height > 0 {
-                let (_, screen_height) = terminal::size()?;
-                let clear_height = clear_height.min(screen_height);
-                if clear_height > 0 {
-                    let skim_top = screen_height.saturating_sub(clear_height);
-                    let prompt_start = painter_state.prompt_start_row();
-                    let prompt_end = painter_state.prompt_end_row();
-                    if prompt_end >= skim_top {
-                        reuse_prompt = false;
-                    }
-
-                    let cursor_position = painter_state.cursor_position();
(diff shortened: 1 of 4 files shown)


fdncred:
I think skim is cool and once it supports Windows this will be great. They just added support for ratatui but they don't support crossterm yet so this makes it *nix only which breaks one of our tenants of reedline and nushell, which is to be as cross-platform as possible.

I believe nucleo used in helix is cross platform. Was hoping someone would adpat that to something like this.

blindFS:
I like the ideal of external menus, looks cool! Thanks for implementing it.

I have a small question about the menu updating mechanism: AFAIK, with completion menu turned on, each typed character triggers another completion task in nushell that refreshes menu candidates (correct me if wrong). It seems very wasteful for menus like skim where the matching could be done in reedline itself.

So here's a dilemma, if we enable skim's matching, how do we make it configurable through `$env.config.completion.algorithm`? If don't, it seems it's just made for visual pleasures.
@ysthakur any thought?

benvansleen:
> ... this makes it *nix only which breaks one of our tenants of reedline and nushell, which is to be as cross-platform as possible.

Very reasonable aspiration. This principle should hold even for optional compile-time features?

> I believe nucleo used in helix is cross platform. Was hoping someone would [adapt] that to something like this.

Will have to give that a look. Am not emotionally tied to skim; just thought it would be neat to utilize the crate.

PR #1018, opened 2026-01-31
add jiff-datetime feature

fdncred:
This is my first attempt at switching from the `chrono` crate to the `jiff` crate. It's behind a feature named `jiff-datetime` so that I can experiment with it in nushell.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,18 +15,20 @@ jobs:
         rust:
           - stable
         # Define the feature sets that will be built here (for caching you define a separate name)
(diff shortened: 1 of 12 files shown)


kronberger-droid:
Hey @fdncred,
whats the state of this?
I am just going around and checking on all the stale PRs.
I am totally fine with keeping this in the state it is if you want.
Just checking in.

fdncred:
I'd like to migrate to jiff at some point here and in nushell and away from chrono but it doesn't have to be this PR.

kronberger-droid:
Okay, then lets keep it open for discussion, or until it is superseded by another one.

PR #1019, opened 2026-02-04
add native semantic prompt support initially for OSC 133/633

fdncred:
This PR adds native support for semantic prompt markers (OSC 133/633) to reedline, enabling proper terminal integration with Ghostty, VS Code, and other terminals that support the FinalTerm semantic prompts specification.

- Uses ST (`\x1b\\`) terminator, not BEL (`\x07`)
- Consumers (like nushell) can enable markers via `with_semantic_markers()`
- C/D execution markers remain the responsibility of the shell

There are 2 new examples. `semantic_prompt_verify` is for manually inspecting the output to see what it looks like to humans. `semantic_prompt_interactive` is for using with ghostty or similar tools that provide a debug interface to see the escapes. for ghostty, use the "Terminal Inspector" and then on the Renderer menu make sure you enable Overlay Semantic Prompts. 

I've done my best to ensure that these escapes are per [spec](https://gitlab.freedesktop.org/Per_Bothner/specifications/-/blob/master/proposals/semantic-prompts.md?ref_type=heads). Let me know if you disagree or think I've missed something.
### Example semantic_prompt_verify

```
❯ cargo run --example semantic_prompt_verify
Semantic Prompt Verification
============================

OSC 133 Markers (Standard):
----------------------------
  Primary prompt start:   OSC 133;P;k=i ST = \e]133;P;k=i\e\
  Secondary prompt start: OSC 133;P;k=s ST = \e]133;P;k=s\e\
  Right prompt start:     OSC 133;P;k=r ST = \e]133;P;k=r\e\
  Command input start:    OSC 133;B ST     = \e]133;B\e\

OSC 633 Markers (VS Code):
--------------------------
  Primary prompt start:   OSC 633;A;k=i ST = \e]633;A;k=i\e\
  Secondary prompt start: OSC 633;A;k=s ST = \e]633;A;k=s\e\
  Right prompt start:     OSC 633;P;k=r ST = \e]633;P;k=r\e\
  Command input start:    OSC 633;B ST     = \e]633;B\e\

Expected Prompt Sequence: (133;A is oveerridden with P to avoid newline issues)
------------------------
For a prompt like: '~/src > ls -la'

1. prompt_start(Primary)  -> OSC 133;P;k=i ST  (before '~/src ')
2. [left prompt text]     -> '~/src '
3. [indicator text]       -> '> '
4. command_input_start()  -> OSC 133;B ST     (after indicator)
5. prompt_start(Right)    -> OSC 133;P;k=r ST (before right prompt)
6. [right prompt text]    -> any right prompt

Multiline Continuation:
----------------------
For a multiline prompt continuation:

1. prompt_start(Secondary) -> OSC 133;P;k=s ST (before '::: ')
2. [continuation indicator] -> '::: '
3. command_input_start()   -> OSC 133;B ST    (after indicator)

Raw Byte Sequences:
------------------
  OSC 133;P;k=i ST = ESC ] 1 3 3 ; P ; k = i ESC \\
  OSC 133;P;k=s ST = ESC ] 1 3 3 ; P ; k = s ESC \\
  OSC 133;P;k=r ST = ESC ] 1 3 3 ; P ; k = r ESC \\
  OSC 133;B ST     = ESC ] 1 3 3 ; B ESC \\

Note: C (pre-exec) and D (post-exec) markers are emitted by the shell,
not by reedline, as they relate to command execution lifecycle.
```
### Example semantic_prompt_interactive
image

### DIFF
--- a/src/terminal_extensions/semantic_prompt.rs
+++ b/src/terminal_extensions/semantic_prompt.rs
@@ -9,11 +9,12 @@
 //! ## Protocol Overview
 //!
 //! - `A` - Marks the start of a prompt (with optional `k=` kind parameter)
-//!         Thanks to mitchellh for this tidbit, "The `A` is kind of weird: it
-//!         does a `\r\n` _if the cursor isn't at x=0_ and THEN does a `P`. If
-//!         you want, you can use `P` all the time instead of `A` for example
-//!         if you're not sure if the cursor will be at x=0." So, with that in
-//!         mind, we'll use `P` for right prompts and left prompts.
+//!
+//!  Thanks to mitchellh for this tidbit, "The `A` is kind of weird: it does
+//!  a `\r\n` _if the cursor isn't at x=0_ and THEN does a `P`. If you want,
+//!  you can use `P` all the time instead of `A` for example if you're not
+//!  sure if the cursor will be at x=0." So, with that in mind, we'll use `P`
+//!  for right prompts and left prompts.
 //! - `B` - Marks the end of prompt and start of user input
 //! - `C` - Marks the start of command execution (emitted by shell, not reedline)
 //! - `D` - Marks the end of command execution with exit code (emitted by shell)


src/terminal_extensions/semantic_prompt.rs
```diff
@@ -0,0 +1,208 @@
...
+//! - Selecting command output
+//! - Visual indicators for prompt regions
+//!
+//! ## Protocol Overview
+//!
+//! - `A` - Marks the start of a prompt (with optional `k=` kind parameter)
+//! - `B` - Marks the end of prompt and start of user input
+//! - `C` - Marks the start of command execution (emitted by shell, not reedline)
+//! - `D` - Marks the end of command execution with exit code (emitted by shell)
+//! - `P` - Sets a property (used for right prompt marker `k=r`)
```
mitchellh, line 21: This is technically `P` for `prompt`. So it starts a prompt. The `A` is kind of weird: it does a `\r\n` _if the cursor isn't at x=0_ and THEN does a `P`. If you want, you can use `P` all the time instead of `A` for example if you're not sure if the cursor will be at x=0 😄

fdncred: ok, I've updated the comments with your suggestion and the code, of course. I left the vscode at 633;A but change the others to 133;P all the time. Thanks for the tip.

PR #1022, opened 2026-02-06
feat(emacs): add keybindings for moving to start and end of buffer

Juhan280:
Implements the standard Emacs shortcuts `Alt+<` and `Alt+>` to quickly navigate the cursor.

- Alt+<: Move to the start of the buffer
- Alt+>: Move to the end of the buffer


fdncred:
if they're supposed to work everywhere, vi and emacs. i think that's what common is for, iirc.

PR #1025, opened 2026-02-09
feat: use dedicated events for buffer jumps while preserving selection via EditCommands

Juhan280:
This adds new ReedlineEvent variants for moving to the start and end of the buffer. These are added alongside existing EditCommands to provide a more direct way to trigger these movements. These were added to allow traversing the history after moving the cursor the start or end of the buffer.

The default Emacs and Vi keybindings have been updated to use them. But in visual mode, we are still using the EditCommand variants.

### DIFF
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -864,10 +864,10 @@ pub enum ReedlineEvent {
     /// Move left to the next column, or completion entry
     Left,
 
-    /// Move to the start
+    /// Move to the start of the buffer
     ToStart,
 
-    /// Move to the end
+    /// Move to the end of the buffer
     ToEnd,
 
     /// Navigate to the next historic buffer


Juhan280:
If you have suggestions for more better enum variant name or doc, please share.

Also in `src/edit_mode/vi/motion.rs`, there are todo comments for `Motion::Up` and `Motion::Down` arm of the `to_reedline` method. If what have I done so far makes sense, I can also add their `EditCommand` variants to make them work in visual mode

src/enums.rs
```diff
@@ -864,6 +864,12 @@ pub enum ReedlineEvent {
     /// Move left to the next column, or completion entry
     Left,
 
+    /// Move to the start
```
ysthakur: nitpick: I'd say "Move to the start of the buffer." Enum names seem fine though

Juhan280: Hmm, I thought about that, but decided `ReedlineEvent`s are also used to do actions other than buffer navigation. Like for example jumping to the first/last completion item.
But I guess I can just say buffer for now and we ccan change it in the future if we add more features.

ysthakur:
> Once we use EditCommand, we cannot traverse the history.

Ah, I wasn't aware of this. Just tried it out and you're right. LGTM then, thanks!

I wonder if this is a bigger problem we need to fix, though, or if it's necessary behavior. Need to look into this further

PR #1032, opened 2026-03-01
refactor: move jump-to-buffer start/end bindings to common navigation

Juhan280:
Relocates ALT+< and ALT+> from emacs to common.
Also adds SHIFT+ALT+, and SHIFT+ALT+. to support the Kitty keyboard protocol

cc @weirdan

### DIFF
--- a/src/edit_mode/keybindings.rs
+++ b/src/edit_mode/keybindings.rs
@@ -198,6 +198,13 @@ pub fn add_common_navigation_bindings(kb: &mut Keybindings) {
         KC::Char('n'),
         ReedlineEvent::UntilFound(vec![ReedlineEvent::MenuDown, ReedlineEvent::Down]),
     );
+
+    // Jump to the start/end of the buffer
+    kb.add_binding(KM::ALT, KC::Char('<'), ReedlineEvent::ToStart);
+    kb.add_binding(KM::ALT, KC::Char('>'), ReedlineEvent::ToEnd);
+    // For kitty keyboard protocol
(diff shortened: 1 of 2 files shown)


src/edit_mode/keybindings.rs
```diff
@@ -198,6 +198,13 @@ pub fn add_common_navigation_bindings(kb: &mut Keybindings) {
         KC::Char('n'),
         ReedlineEvent::UntilFound(vec![ReedlineEvent::MenuDown, ReedlineEvent::Down]),
     );
+
+    // Jump to the start/end of the buffer
+    kb.add_binding(KM::ALT, KC::Char('<'), ReedlineEvent::ToStart);
+    kb.add_binding(KM::ALT, KC::Char('>'), ReedlineEvent::ToEnd);
+    // For kitty keyboard protocol
```
fdncred, line 205: if these are for the kitty protocol should we check to ensure kitty protocol is enabled adding using these?

Juhan280: How can I check them within this function? Also the user might enable/disable kitty protocol after startup. Then it won't work for them.

fdncred: ya, it would have to enable/disable bindings whenever the kitty protocol was enabled/disabled. However, I'm not clear on if the keybindings require the kitty protocol or not. What happens if kitty isn't enabled and you have these bindings?

Juhan280: > What happens if kitty isn't enabled and you have these bindings?

You can't trigger those bindings without kitty protocol.
See https://github.com/nushell/nushell/pull/17508#issuecomment-3980982930

PR #1034, opened 2026-03-08
add EnumDiscriminants to ReedlineEvent and EditCommand

Juhan280:
The main purpose of this PR is to facilitate nushell to make its own display implementation for `ReedlineEvent` and `EditCommand`. Currently, nushell determines which fields are optional and which field is required. But for some reason, the display for that is defined in reedline. As a result it causes confussion (see https://github.com/nushell/reedline/pull/953#discussion_r2698499834)

I will make a follow up PR to this in nushell implementing that change.

Also, I would also like to remove the `Display` impl of the mentioned enums. But I am hesitant to do so because that would be a breaking change.

/cc @fdncred

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -390,9 +390,9 @@ dependencies = [
 
 [[package]]
 name = "heck"
-version = "0.4.1"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


fdncred:
Thanks for making the nushell PR. It made it easier to test this out. LGTM!

PR #1035, opened 2026-03-11
feat: add external break signal for `read_line` interruption and mark `Signal` as `non_exhaustive`

eitsupi:
Add a break signal that allows external threads to interrupt `read_line()`, causing it to return a new `Signal::ExternalBreak(String)` variant with the current buffer contents.

## Motivation

Applications that embed a REPL alongside external control systems (e.g., AI agents) need a way to interrupt `read_line()` from another thread.
For example, an AI agent controlling the console may need to inject input while the user is at the prompt — but `read_line()` is a blocking loop with no external interruption mechanism.

Currently, there is no way for an external thread to cause `read_line()` to return.
The `external_printer` feature can print messages above the prompt, but cannot break out of the input loop.
The `idle_callback` feature (#1015) allows periodic processing but also cannot cause an early return.

## Design

The break signal check is placed in the `read_line_helper` loop alongside `idle_callback` and `external_printer`.
When `break_signal` is configured, `needs_polling` is set to `true` so `read_line()` uses `poll()` instead of blocking indefinitely on `crossterm::event::read()`.

## API

```rust
use std::sync::{atomic::AtomicBool, Arc};

let break_signal = Arc::new(AtomicBool::new(false));
let mut editor = Reedline::create()
    .with_break_signal(break_signal.clone());

// From another thread:
break_signal.store(true, Ordering::Relaxed);

// read_line() returns:
// Ok(Signal::ExternalBreak(buffer_contents))
```

## Note

`Signal` is now `#[non_exhaustive]`.
Downstream match arms on `Signal` will need a wildcard `_` pattern.
This is a one-time change that prevents future additions from being breaking.

### DIFF
--- a/examples/break_signal.rs
+++ b/examples/break_signal.rs
@@ -1,0 +1,53 @@
+// Demonstrates the external break signal feature.
+// A background thread sets the break signal after 3 seconds,
+// causing `read_line()` to return `Signal::ExternalBreak` with
+// the current buffer contents.
+//
+// To run:
+// cargo run --example break_signal
+
+use reedline::{DefaultPrompt, Reedline, Signal};
+use std::{
+    io,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    thread,
+    time::Duration,
+};
+
+fn main() -> io::Result<()> {
+    let break_signal = Arc::new(AtomicBool::new(false));
+
+    let mut line_editor = Reedline::create().with_break_signal(break_signal.clone());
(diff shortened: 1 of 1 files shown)


eitsupi:
> Is there any way to test this automatically and/or manually?

I've added an example so you can test it manually.

Automated testing requires a terminal, which is tested in arf test suite (https://github.com/eitsupi/arf/pull/113) using https://github.com/doy/vt100-rust (unfortunately not compatible with Windows).

PR #1040, opened 2026-03-17
fix: Strip ANSI before computing suggestion width in completion menus

ysthakur:
We don't strip ANSI escapes when computing suggestion width, which means that the padding is off when the suggestion has `display_override` set to a string containing ANSI escapes. I didn't detect this during testing because all of the suggestions I tested always had the same number of ANSI escapes.

### Before

Columnar menu:
image

IDE menu:
image

### After

Columnar menu:
image

IDE menu:
image

Completer used:
```nu
def comp [] {
    [
        {
            value: "Cookies",
            description: "A great healthy snack",
        },
        {
            value: "Spinach",
            display_override: $"gah(ansi green)blech(ansi s)hurgh",
            description: "A poisonous leaf",
            style: red
        },
    ]
}
def foo [commit: string@comp] {}
```

### DIFF
--- a/src/menu/columnar_menu.rs
+++ b/src/menu/columnar_menu.rs
@@ -438,8 +438,8 @@ impl ColumnarMenu {
                         format!(
                             "{}{}{}{}{}",
                             styled_value,
-                            RESET,
                             " ".repeat(padding),
+                            RESET,
                             self.settings.color.description_style.paint(desc_trunc),
                             RESET,
                         )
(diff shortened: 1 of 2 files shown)


src/menu/columnar_menu.rs
```diff
@@ -438,8 +438,8 @@ impl ColumnarMenu {
                         format!(
                             "{}{}{}{}{}",
                             styled_value,
-                            RESET,
                             " ".repeat(padding),
+                            RESET,
```
ysthakur, line 442: Put the reset after padding to stay consistent with how selected values are highlighted

WindSoilder (approved):
Thanks!  LGTM

PR #1041, opened 2026-03-18
feat: accept templated command for `OpenEditor`

pickx:
this PR extends `OpenEditor` in the following way:
- `BufferEditor`'s `command` field can now contain the following patterns: `{file}`, `{line}`, `{col}`
- reedline will automatically instantiate the fields from current position when `OpenEditor` is invoked

the usecase here is that pressing Ctrl+O in nushell would open the editor in the current position of the cursor. this branch works as-is in nushell without additional changes to nushell itself, other than the user needing to use a different format for `buffer_editor`.  it is (or _should be_) backward-compatible with existing `buffer_editor` commands.

I tested this with:
- `$env.config.buffer_editor = ["hx" "{file}:{line}:{col}"]`
- `$env.config.buffer_editor = ["code" "--wait" "--goto" "{file}:{line}:{col}"]`

TODOs in the code will be removed in subsequent commits - they're just potential areas of discussion for the PR process.

### DIFF
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -95,7 +95,7 @@ impl LineBuffer {
     /// Zero-based index
     pub fn col(&self) -> usize {
         self.lines[self.line_start()..self.insertion_point]
-            .chars()
+            .grapheme_indices(true)
             .count()
     }
 
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -90,6 +90,15 @@ impl LineBuffer {
         self.lines[..self.insertion_point].matches('\n').count()
     }
 
+    /// Calculates the character index in the line the user is on
+    ///
+    /// Zero-based index
+    pub fn col(&self) -> usize {
+        self.lines[self.line_start()..self.insertion_point]
+            .chars()
+            .count()
+    }
+
     /// Counts the number of lines in the buffer
     pub fn num_lines(&self) -> usize {
         self.lines.split('\n').count()
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -83,7 +83,7 @@ impl LineBuffer {
         self.insertion_point = self.lines.len();
     }
 
-    /// Calculates the current the user is on
+    /// Calculates the current line the user is on
     ///
     /// Zero-based index
     pub fn line(&self) -> usize {
@@ -105,20 +105,21 @@ impl LineBuffer {
         self.insertion_point = 0;
     }
 
+    fn line_start(&self) -> usize {
+        self.lines[..self.insertion_point]
+            .rfind('\n')
+            .map_or(0, |offset| offset + 1)
+        // str is guaranteed to be utf8, thus \n is safe to assume 1 byte long
+    }
+
     /// Move the cursor before the first character of the line
     pub fn move_to_line_start(&mut self) {
-        self.insertion_point = self.lines[..self.insertion_point]
-            .rfind('\n')
-            .map_or(0, |offset| offset + 1);
-        // str is guaranteed to be utf8, thus \n is safe to assume 1 byte long
+        self.insertion_point = self.line_start();
     }
 
     /// Move the cursor before the first non whitespace character of the line
     pub fn move_to_line_non_blank_start(&mut self) {
-        let line_start = self.lines[..self.insertion_point]
-            .rfind('\n')
-            .map_or(0, |offset| offset + 1);
-        // str is guaranteed to be utf8, thus \n is safe to assume 1 byte long
+        let line_start = self.line_start();
 
         self.insertion_point = self.lines[line_start..]
             .find(|c: char| !c.is_whitespace() || c == '\n')
@@ -535,9 +536,7 @@ impl LineBuffer {
     /// extending beyond the potential carriage return and line feed characters
     /// terminating the line
     pub fn current_line_range(&self) -> Range<usize> {
-        let left_index = self.lines[..self.insertion_point]
-            .rfind('\n')
-            .map_or(0, |offset| offset + 1);
+        let left_index = self.line_start();
         let right_index = self.lines[self.insertion_point..]
             .find('\n')
             .map_or_else(|| self.lines.len(), |i| i + self.insertion_point + 1);
--- a/src/core_editor/line_buffer.rs
+++ b/src/core_editor/line_buffer.rs
@@ -196,6 +196,15 @@ impl LineBuffer {
         self.lines[..self.cursor.head()].matches('\n').count()
     }
 
+    /// Calculates the grapheme index in the line the cursor head is on.
+    ///
+    /// Zero-based index
+    pub fn col(&self) -> usize {
+        self.lines[self.line_start_index()..self.cursor.head()]
+            .grapheme_indices(true)
+            .count()
+    }
+
     /// Counts the number of lines in the buffer
     pub fn num_lines(&self) -> usize {
         self.lines.split('\n').count()
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -2214,6 +2214,7 @@ mod tests {
     use super::*;
     use crate::terminal_extensions::semantic_prompt::PromptKind;
     use crate::DefaultPrompt;
+    use rstest::rstest;
 
     #[test]
     fn test_cursor_position_after_multiline_history_navigation() {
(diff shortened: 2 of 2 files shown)


pickx:
I realized I was re-inventing the wheel, so I rebased away unnecessary code.

src/core_editor/line_buffer.rs
```diff
@@ -83,13 +83,22 @@ impl LineBuffer {
...
     pub fn line(&self) -> usize {
         self.lines[..self.insertion_point].matches('\n').count()
     }
 
+    /// Calculates the character index in the line the user is on
+    ///
+    /// Zero-based index
+    pub fn col(&self) -> usize {
+        self.lines[self.line_start()..self.insertion_point]
+            .chars()
```
fdncred: i'm wondering if .chars() is right here. e.g. should it be graphemes or should it call is_valid()?

fdncred:
Can we somehow add a test for this and test this in an example?

src/core_editor/line_buffer.rs
pickx: I'm not sure, actually. would appreciate some help here.

pickx:
as I was writing the examples/tests, I decided that requiring `{file}` is surprising.
instead we now append the filename at the end, _unless_  `{file}` is specified.
tell me if you think we should do things differently

src/core_editor/line_buffer.rs
fdncred: it should probably use is_valid to ensure that the col you're inserting isn't in the middle of a unicode character and impossible to be in the middle of.

pickx: looks like we want `grapheme_indices(true)` here, after testing with a buffer containing non-unicode characters. `chars()` is definitely wrong

fdncred: your change is probably alright but `is_valid() does something similar

pickx: sorry, looks like I'm not getting you 🙂
are you suggesting we validate using this function because even going by grapheme is not a strong enough guarantee? or that we change the impl to something that looks like is_valid?

please spoon feed me, lol. or simply commit whatever change makes sense.

fdncred: No worries. What I saw saying is that I think we already have a function that is doing something similar to what you're trying to do. So, you could call this function to ensure where you were inserting is a good place.
```rust
    /// Check if the line buffer is valid utf-8 and the cursor sits on a valid grapheme boundary
    pub fn is_valid(&self) -> bool {
        self.lines.is_char_boundary(self.insertion_point())
            && (self
                .lines
                .grapheme_indices(true)
                .any(|(i, _)| i == self.insertion_point())
                || self.insertion_point() == self.lines.len())
            && std::str::from_utf8(self.lines.as_bytes()).is_ok()
    }
```
I'm not positive but if we set the insertion_point() and then call is_valid(), it should tell us if that's a good place.

Maybe I'm making too much of this or am just wrong and I'm just missing the point. I'm just trying to have less duplicate/similar code if we can.

We can just go with what you have if I'm being too confusing.

pickx:
I would like to revive this PR.
`main` has drifted since this was opened, but after resolving conflicts, this still works.

1. is this a feature you would like to see in `reedline`? if so, anything else you want to see here? (if not, that's fine, and you can close this.)
2. due to the conflicts: do we prefer a merge commit, or a rebase+squash?
3. suggestions on how to test this? I've already added tests for `render_editor_command`, but I'm not sure if that is sufficient.
4. this substitution thing feels like it can be handled by `ENV_CONVERSIONS`...? of course, not without some additional changes. also, this requires `nushell`, so no longer keeps this contained to `reedline`.

fdncred:
1. I think having this would be cool in nushell
```nushell
$env.config.buffer_editor = ["hx" "{file}:{line}:{col}"]
$env.config.buffer_editor = ["code" "--wait" "--goto" "{file}:{line}:{col}"]
```
However, I'd like it to happen without breaking changes so whatever we have needs to be optional. Maybe it's like that already?

2. I'm fine with either rebase+squash or merge commit since this is so old now.

3. I'm not sure about tests. We'd definitely want some in nushell too.

4. I'm not sure how ENV_CONVERSIONS would help but I'm just waking up too.

I'd like to get @kronberger-droid's thoughts on this one too.

pickx:
(as far as I can tell, this isn't breaking, and is reverse-compatible with existing setups)

kronberger-droid:
Sounds nice, I like the feature.

The new editor backend would maybe change the code just a little, but it general it looks fine already.

pickx:
@kronberger-droid okay, now that the recent `reedline` changes appears to have landed:
I rebased this. not much actual code changes here now.
ready for review.

src/engine.rs
```diff
@@ -241,6 +241,59 @@ struct BufferEditor {
     temp_file: PathBuf,
 }
 
+impl BufferEditor {
+    /// Renders the editor command template,
+    /// substituting `{file}`, `{line}`, and `{col}` where present.
+    pub(crate) fn render_command(&self, line_buffer: &LineBuffer) -> Command {
```
pickx, line 246: right now we do this even if the command is not templated.
is this is considered wasteful, we can avoid it by determining ahead of time if it's templated or not
please tell me if you think it matters.

fdncred:
let's see what copilot thinks.

src/engine.rs
```diff
@@ -241,6 +241,59 @@ struct BufferEditor {
...
+        let mut cmd = Command::new(self.command.get_program());
```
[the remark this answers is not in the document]
pickx: I admit, I don't understand what this is about.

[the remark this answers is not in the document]
fdncred: maybe @kronberger-droid understands?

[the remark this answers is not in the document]
kronberger-droid: Well my guess was building a new command just looses the env and some other stuff around it.

I sadly don't have time right now to go into it to much. But claude agreed and was quite clear that we loose quite some features and some of it is unrecoverable in the way we write it right now:
 
"`render_command` builds a fresh `Command` from `get_program()` and `get_args()`, so everything else the caller set stays behind on the old object. nushell leans on exactly that: `nu-cli/src/repl.rs:713` does `command.args(args).envs(envs)` since `$env` isn't the process env, so the editor ends up with the bare inherited env instead. Copying `get_envs()` and `get_current_dir()` would fix some of it. `env_clear()`, the stdio config and the unix `CommandExt` settings have no getters at all, and a bare `env_clear()` reads back identical to a command that never touched the env, so that part isn't recoverable from a stored `Command` however we copy it."

I will take a closer look later.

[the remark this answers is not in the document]
kronberger-droid: But i think its essentially that that nushell hands its whole `$env` to the command via `.envs(...)`, and a freshly built `Command` has no way to pick that back up.
`get_envs()` would recover that part at least.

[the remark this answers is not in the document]
kronberger-droid: So the env is easy enough to solve. But there are other methods which where possible on the prebuilt Command one would loose. So easiest would be to just ship the default implementation when no template is used and for the template at least add the env. Since nushell depends on it.

I would argue that we don't merge this until we have it ported and working in a branch of nushell.

pickx:
so...

1. while testing, I noticed that a non-existent `$env.config.buffer_editor` command is an unhandled `reedline` error and crashes `nushell`. uh, this really seems like something `nushell` should be able to recover from.
2. the failure to propagate env makes this worse because it can make crashing more common. for example if the editor command was in the previous `$env.PATH` but env wasn't propagated.
3. any optimizations here (such as somehow reusing the `Command`, or storing the args separately, ...) don't really benefit us now, since I believe the REPL recreates this command regardless.

src/engine.rs
kronberger-droid: `get_envs()` covers nushell, but `current_dir`, `env_clear` and the stdio settings are still dropped, and only `current_dir` has a getter.
When no argument contains a placeholder, spawn `self.command` itself, with `temp_file` appended at construction as before. Rebuild only for a template, and copy `get_current_dir()` there.

src/engine.rs
```diff
@@ -685,20 +738,27 @@ impl Reedline {
...
+    /// let mut line_editor = Reedline::create().with_buffer_editor(command, temp.clone());
+    ///
+    /// // if {file} is omitted, the filename is still appended at the end,
+    /// // as in the above example
+    /// let mut command = Command::new("emacs");
+    /// command.arg("+{line}:{col}");
+    /// let mut line_editor = Reedline::create().with_buffer_editor(command, temp);
     /// ```
     #[must_use]
     pub fn with_buffer_editor(mut self, editor: Command, temp_file: PathBuf) -> Self {
```
kronberger-droid, line 761: A caller that passes `temp_file` as an argument, which the old `contains` check allowed, now gets it twice.
Keep that check next to the placeholder one

kronberger-droid (commented):
Sorry for the wait.

1. The crash is a nushell side fix. Worth making an issue or doing a PR there.
2. `get_envs()` should cover it, thanks.
3. The rebuild was not about allocation, it is about what a fresh command drops. Mentioned inline.

After the changes inline i think we can land it.

PR #1042, opened 2026-03-19
fix: preserve prompt position after ExternalBreak

eitsupi:
Follow-up to #1035.

In #1035, when a caller handles `ExternalBreak` by calling `read_line()` again (e.g., to resume editing after inspecting the buffer), the prompt jumps down one line each time.
This is because `move_cursor_to_end()` unconditionally prints a CRLF before returning `ExternalBreak`.

This PR replaces `move_cursor_to_end()` with saving `suspended_state`, matching the pattern used by `ExecuteHostCommand`.
The updated example demonstrates resuming `read_line()` after `ExternalBreak`.

### DIFF
--- a/examples/break_signal.rs
+++ b/examples/break_signal.rs
@@ -1,7 +1,10 @@
 // Demonstrates the external break signal feature.
-// A background thread sets the break signal after 3 seconds,
+//
+// A background thread sets the break signal every 5 seconds,
 // causing `read_line()` to return `Signal::ExternalBreak` with
-// the current buffer contents.
+// the current buffer contents. The example then resumes editing
+// by calling `read_line()` again — the prompt stays on the same
+// line thanks to suspended-state preservation.
 //
 // To run:
 // cargo run --example break_signal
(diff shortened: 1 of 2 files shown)


fdncred:
How does this affect normal use?

eitsupi:
> How does this affect normal use?

This change only affects callers that use the `break_signal` feature (introduced in #1035) and call `read_line()` again after receiving `ExternalBreak`.
For normal nushell use, there is no impact.
For `break_signal` users, the behavior change is: after `ExternalBreak`, the next `read_line()` call reuses the original prompt position instead of advancing one line. This matches how `ExecuteHostCommand` already works.

Apologies for the late fix. I should have caught this issue when working on #1035.

fdncred:
Am I misreading this? I thought this change was for everyone?
https://github.com/nushell/reedline/pull/1042/changes#diff-ff278ec599adfd033d35a083ebe8948db8bf177c9817dee8ddc4f3a9e3f5728eR819-R820

eitsupi:
> Am I misreading this? I thought this change was for everyone?
> [#1042 (changes)](https://github.com/nushell/reedline/pull/1042/changes#diff-ff278ec599adfd033d35a083ebe8948db8bf177c9817dee8ddc4f3a9e3f5728eR819-R820)

The changed code is inside the `if let Some(ref signal) = self.break_signal` block, which is only entered when a caller has configured a break signal via `with_break_signal()`.
Without that, `self.break_signal` is `None` and this path is never reached.

The core fix is replacing `move_cursor_to_end()` with `suspended_state` saving (same pattern as `ExecuteHostCommand`).
The `last_render_snapshot = None` on line 819 was added for consistency with `ExecuteHostCommand`.
it has no functional impact since the snapshot is only used for mouse-click handling and is refreshed on the next `repaint()`.

fdncred:
ah, ok. I missed that. I guess this isn't an issue anymore since you closed it?

eitsupi:
I'm sorry, it was an accident.
I accidentally closed from my repository that was referencing this pull request😓

fdncred:
Thanks

eitsupi:
Thanks for the quick review!

PR #1043, opened 2026-03-30
Update default prompt to use > character

speicherwerk:
The previously used character `RIGHT ANGLE BRACKET` (3009) is not available in many fonts. The usual `GREATER-THAN SIGN` (003e) however is. (Plus it looks nicer).

### DIFF
--- a/src/prompt/default.rs
+++ b/src/prompt/default.rs
@@ -6,9 +6,9 @@ use {
 };
 
 /// The default prompt indicator
-pub static DEFAULT_PROMPT_INDICATOR: &str = "〉";
+pub static DEFAULT_PROMPT_INDICATOR: &str = ">";
 pub static DEFAULT_VI_INSERT_PROMPT_INDICATOR: &str = ": ";
-pub static DEFAULT_VI_NORMAL_PROMPT_INDICATOR: &str = "〉";
+pub static DEFAULT_VI_NORMAL_PROMPT_INDICATOR: &str = ">";
 pub static DEFAULT_MULTILINE_INDICATOR: &str = "::: ";
 
 /// Simple [`Prompt`] displaying a configurable left and a right prompt.
--- a/src/prompt/default.rs
+++ b/src/prompt/default.rs
@@ -6,9 +6,9 @@ use {
 };
 
 /// The default prompt indicator
-pub static DEFAULT_PROMPT_INDICATOR: &str = ">";
+pub static DEFAULT_PROMPT_INDICATOR: &str = "> ";
 pub static DEFAULT_VI_INSERT_PROMPT_INDICATOR: &str = ": ";
-pub static DEFAULT_VI_NORMAL_PROMPT_INDICATOR: &str = ">";
+pub static DEFAULT_VI_NORMAL_PROMPT_INDICATOR: &str = "> ";
 pub static DEFAULT_MULTILINE_INDICATOR: &str = "::: ";
 
 /// Simple [`Prompt`] displaying a configurable left and a right prompt.


fdncred:
Are there people who don't use nerd fonts?

speicherwerk:
> Are there people who don't use nerd fonts?

Yes there are. Me for one. Also as a library with an end-user interface, we should be sensible regarding defaults. I don't think one is to assume that the average end-user of a CLI tool uses a nerd font.

fdncred:
I think it's pretty safe since you're the first person I remember complaining about it since the project started which is more than 5 years now. I'm not saying we can't land your PR but I still think it's pretty safe.

NotTheDr01ds:
@speicherwerk What OS, terminal, and font are you using?  I agree that most Western fonts don't have this character by default, but it still seems to work for most people.  I've even seen it before in stock Linux distro Bash prompts.

For me, even on a fresh Windows system using the default Windows Terminal with Cascadia Cove (which does *not* include that character), it still renders correctly.  It also works in Xterm through WSL2.

I'm not an expert in this area so I'm relying on Gemini for some of this (which could be vastly wrong), but this is because the font engine on the system is supposed to be able to do font-matching or bubbling.  On Windows, I can confirm that this works properly.  On Linux, this seems to be part of [FontConfig](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/fc-match/fc-match.c?ref_type=heads).

To @fdncred's point, we really haven't heard complaints about this not working, and the character has been part of the default prompt for at least 5 years now (probably longer).

For the second part, I agree `> ` is nicer looking, so that's part of my configuration personally.

speicherwerk:
> @speicherwerk What OS, terminal, and font are you using? I agree that most Western fonts don't have this character by default, but it still seems to work for most people. I've even seen it before in stock Linux distro Bash prompts.
> 
> For me, even on a fresh Windows system using the default Windows Terminal with Cascadia Cove (which does _not_ include that character), it still renders correctly. It also works in Xterm through WSL2.
> 
> I'm not an expert in this area so I'm relying on Gemini for some of this (which could be vastly wrong), but this is because the font engine on the system is supposed to be able to do font-matching or bubbling. On Windows, I can confirm that this works properly. On Linux, this seems to be part of [FontConfig](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/fc-match/fc-match.c?ref_type=heads).
> 
> To @fdncred's point, we really haven't heard complaints about this not working, and the character has been part of the default prompt for at least 5 years now (probably longer).
> 
> For the second part, I agree `> ` is nicer looking, so that's part of my configuration personally.

I don't know awfully lot about digital typesettings so I can't really contribute anything here but for what it's worth I was on Alacritty using a pretty recent version of Iosevka Term without any patches (on Arch Linux). Alacritty uses [`crossfont`](https://crates.io/crates/crossfont/0.8.1) for font rendering (as far as I can tell), which claims to use `FreeType` as a backend on Linux.

kronberger-droid:
Hey @speicherwerk, I encountered this problem too, seems the char is not even in nerd fonts.
We have now decided to use > (greater than) to be safe.
But for this to land we need you to add a trailing space behind "> " to match the usual convention.

speicherwerk:
> Hey @speicherwerk, I encountered this problem too, seems the char is not even in nerd fonts.
> 
> We have now decided to use > (greater than) to be safe.
> 
> But for this to land we need you to add a trailing space behind "> " to match the usual convention.

Alright. I don't know how I'd missed the space; I will add it right away.

kronberger-droid:
> Alright. I don't know how I'd missed the space; I will add it right away.

All good, the previous char is two broad by default. Maybe you just did a simple replace.
Thanks!

PR #1044, opened 2026-04-06
feat(list_menu): add DescriptionPosition to control description place…

a-rookie-of-C-language:
Add a `DescriptionPosition` enum that lets callers choose whether the
completion description is rendered before or after the value in a
`ListMenu` row.

Previously the position was hardcoded as `(description) value`.
The new default keeps that behaviour (`Before`) for backwards
compatibility; setting `After` renders `value description` instead,
which is more natural for flag/option completions where the value
(e.g. `--regexp`) should visually come first.

Changes:
- `DescriptionPosition` enum with `Before` (default) and `After` variants
- `ListMenu::description_position` field (default: `Before`)
- `ListMenu::with_description_position()` builder method
- `create_string()` branches on `self.description_position`
- Re-export `DescriptionPosition` from `menu/mod.rs` and `lib.rs`

This mirrors the existing `IdeMenu::with_description_mode()` pattern.

### DIFF
--- a/src/menu/list_menu.rs
+++ b/src/menu/list_menu.rs
@@ -13,6 +13,19 @@ use {
 
 const SELECTION_CHAR: char = '!';
 
+/// Controls where the description is rendered relative to the completion value
+/// in a [`ListMenu`] row.
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+pub enum DescriptionPosition {
+    /// Description is shown **before** the value, wrapped in parentheses:
+    /// `(description) value`  — the original behaviour.
+    #[default]
+    Before,
+    /// Description is shown **after** the value with a leading space:
+    /// `value description`
+    After,
+}
+
 struct Page {
     size: usize,
     full: bool,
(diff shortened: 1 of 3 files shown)


src/menu/list_menu.rs
```diff
@@ -265,40 +290,80 @@ impl ListMenu {
...
-        });
+        match self.description_position {
+            DescriptionPosition::Before => {
+                let description = description.map_or("".to_string(), |desc| {
+                    if use_ansi_coloring {
+                        format!(
+                            "{}({}) ",
+                            self.settings.color.description_style.prefix(),
+                            desc,
+                        )
```
kronberger-droid: You dropped the `RESET` here.

kronberger-droid:
Other than the `RESET` you dropped it looks fine.
If you are motivated you could try to duplicate, since you are doing quite a lot of the same stuff in both match arms, but it makes it easy to read so not really a problem.

src/menu/list_menu.rs
a-rookie-of-C-language: 抱歉,我没太注意,感谢您的指出

kronberger-droid:
Nice, `RESET` placement is actually cleaner now. And it reads well, thanks!

Two small things before we land it:
- The `Before` wraps in parentheses and `After` doesn't, which is reasonable. Just add a small comment so no one is tempted to "fix" it.
- The two new tests are nice, but only cover the ANSI paths. The non-ANSI path is not tested. Would be amazing to add those to.

Thanks!

PR #1046, opened 2026-04-08
prompt: default prompt colors to the terminal foreground

iniw:
https://github.com/nushell/reedline/pull/992 reset stale sgr state before each repaint, which fixed prompt color bleed from previous command output. it did not fix the starship case reported in https://github.com/starship/starship/issues/6560 and https://github.com/nushell/nushell/issues/16384 because reedline still applied its own default prompt colors immediately before printing the prompt.

that meant a prompt implementation that returned ansi text without overriding `get_prompt_color()` still inherited reedline's defaults. for starship, `style = "none"` emits no color sequence at all, so the first visible character inherited reedline's green prompt color instead of the terminal's foreground color.

this patch changes the default prompt, multiline, indicator, and right-prompt colors to the terminal defaults. prompts that want explicit colors can still override the color methods, but unstyled prompts now stay unstyled and no longer need an extra reset sequence to avoid the green first character.

### DIFF
--- a/src/prompt/base.rs
+++ b/src/prompt/base.rs
@@ -8,11 +8,12 @@ use {
     strum::EnumIter,
 };
 
-/// The default color for the prompt, indicator, and right prompt
-pub static DEFAULT_PROMPT_COLOR: Color = Color::Green;
-pub static DEFAULT_PROMPT_MULTILINE_COLOR: nu_ansi_term::Color = nu_ansi_term::Color::LightBlue;
-pub static DEFAULT_INDICATOR_COLOR: Color = Color::Cyan;
-pub static DEFAULT_PROMPT_RIGHT_COLOR: Color = Color::AnsiValue(5);
+/// Terminal-default colors for prompt implementations that do not opt into
+/// explicit styling.
+pub static BASE_PROMPT_COLOR: Color = Color::Reset;
+pub static BASE_PROMPT_MULTILINE_COLOR: nu_ansi_term::Color = nu_ansi_term::Color::Default;
+pub static BASE_INDICATOR_COLOR: Color = Color::Reset;
+pub static BASE_PROMPT_RIGHT_COLOR: Color = Color::Reset;
 
 /// The current success/failure of the history search
 pub enum PromptHistorySearchStatus {
@@ -98,21 +99,21 @@ pub trait Prompt: Send {
         &self,
         history_search: PromptHistorySearch,
     ) -> Cow<'_, str>;
-    /// Get the default prompt color
+    /// Get the base prompt color
     fn get_prompt_color(&self) -> Color {
-        DEFAULT_PROMPT_COLOR
+        BASE_PROMPT_COLOR
     }
-    /// Get the default multiline prompt color
+    /// Get the base multiline prompt color
     fn get_prompt_multiline_color(&self) -> nu_ansi_term::Color {
-        DEFAULT_PROMPT_MULTILINE_COLOR
+        BASE_PROMPT_MULTILINE_COLOR
     }
-    /// Get the default indicator color
+    /// Get the base indicator color
     fn get_indicator_color(&self) -> Color {
-        DEFAULT_INDICATOR_COLOR
+        BASE_INDICATOR_COLOR
     }
-    /// Get the default right prompt color
+    /// Get the base right prompt color
     fn get_prompt_right_color(&self) -> Color {
-        DEFAULT_PROMPT_RIGHT_COLOR
+        BASE_PROMPT_RIGHT_COLOR
     }
 
     /// Whether to render right prompt on the last line
--- a/src/prompt/default.rs
+++ b/src/prompt/default.rs
@@ -2,6 +2,7 @@ use crate::{Prompt, PromptEditMode, PromptHistorySearch, PromptHistorySearchStat
 
 use {
     chrono::Local,
+    crossterm::style::Color,
     std::{borrow::Cow, env},
 };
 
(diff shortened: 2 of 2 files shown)


fdncred:
We've covered this in the past. If we set these to none then there are no colors for the prompt at all for reedline users unless you're using something else to color your prompt.

iniw:
i don't think you're right (but i'm not fully sure either!). AFAIK emitting text without a color (which is the case in this patch) will just default to the terminal's foreground color, right? so the actual color of the prompt will depend on the user's terminal emulator

what i mean is that there _will_ be colors, it just won't be controlled by reedline

fdncred:
This is what I'm talking about. I really don't want to lose the default coloring in reedline.

### Before this PR
image

### After this PR
image

iniw:
i understand where you're coming from but i see it the other way around:

users of reedline (including the examples) should override the `get_*_color` family of functions in the `Prompt` trait to achieve their desired styling.
reedline, by default, shouldn't impose a particular style on the prompt. it's too low-level for that.

this is just my personal opinion, of course :)

do you think me making the examples colorful again (by overriding the colors) but keeping the default color-less would be a good middle ground?

fdncred:
> do you think me making the examples colorful again (by overriding the colors) but keeping the default color-less would be a good middle ground?

maybe. i also wonder if we'd need to do this too in nushell when running `nu -n`

iniw:
okay, I've made the `DefaultPrompt` impl use the old styles. lmk if this is more acceptable now

fdncred:
I think it's good. Can you look into what changes are needed in nushell to support this?

iniw:
> Can you look into what changes are needed in nushell to support this?

I implemented this for nushell and have conflicting thoughts about it now. The more I wrote the more I thought that this fix was happening in the wrong place. Regardless of whether or not reedline has a default prompt color, nushell will still have one, so this fix by itself ends up being pointless.

What I ended up doing was checking whether or not a custom prompt string was set or not (which is easy since nushell represents these as `Option<String>` values), but I don't think this is particularly good since it doesn't sound intuitive for the user to have your prompt colors change based on whether or not you have a custom prompt.

After thinking about it for a bit I think this should be fixed on starship's side.

What do you think? I assume you agree, but I'll wait to read your thoughts before closing this and submitting the starship PR.

fdncred:
I think you should be able to use nushell, oh-my-posh, starship, others to change your prompt and reedline should care that you do that.

iniw:
So you think it's ok to change the prompt color based on whether you have a custom prompt or not? If that's alright then I'm happy to implement that.

kronberger-droid:
Hey there, what's the state on this?
I am not quite sure what you decided while reading through it @fdncred.

My understanding:
the green lives in the trait defaults, so every `Prompt` impl inherits it invisibly.
`NushellPrompt` doesn't override the color methods, which is why starship's unstyled text gets painted green.

This PR moves the opinion to the right layer.
The trait default becomes neutral, `DefaultPrompt` keeps the green, so demos and quickstarts look the same.

In nushell we then just implement the green colors in `NushellPrompt` to keep our defaults.
But then to fix the bug, we have to override all colors if a prompt is set.
Essentially saying, the prompt is now responsible for coloring.
Otherwise we would reintroduce the same bug one layer up.

What do you think about that?

fdncred:
@kronberger-droid i'll defer to what you think is appropriate. i just would like to see the same defaults in nushell.

iniw:
I personally see a few ways forward:

- Nushell could simply _not_ have a default prompt color and fallback to the terminal emulator's default foreground color. 
  This is a breaking change and potentially not very nice for newcomers' first impressions.
- Nushell could remove the default prompt color if a custom prompt is set.
  This may be unintuitive. A minor modification to the prompt would have major effects on how it looks? Doesn't sound very nice. Perhaps documenting it would be enough. Hard to say.
- We leave everything as is and instead fix it in starship by forcing it to override the currently active color.

kronberger-droid:
@iniw
As @fdncred already mentioned twice we don't want to change the default prompt coloring for nushell.

Option 2 is the most sensible to me: Reedline provides clean defaults, nushell impl uses nushell default coloring and the contract for custom prompt is essentially: "You want to customize the prompt, it's all up to you how it looks now."
Which is actually more intuitive than a default prompt being a mixture of the nushell defaults and the configured custom prompt.
Otherwise the starship bug would not be considered a bug, but a feature.

@iniw I can take care of the nushell side if you want. Do you think this PR already reflects that sentiment?

iniw:
Nice to finally have direction on this :)

> @iniw I can take care of the nushell side if you want.

Sounds great to me! I didn't think my solution on the nushell side was good, hence why I never PR'd it.

> Do you think this PR already reflects that sentiment?

Yup, I think this PR makes sense.

kronberger-droid:
Great,while I search for a good solution in nushell I will review this and then land in them in sequence.

kronberger-droid:
Check out my nushell PR nushell/nushell#18686.
As described we check if a custom prompt is defined, if so it will remove the coloring and it is now fully handled by the custom prompt.

We can either land this one for reedline also, or just keep it. I am fine with both, probably landing it makes it cleaner.
@fdncred what do you think?

fdncred:
What I don't understand is why we're not using nu_ansi_term for Color. I don't think we should be using any coloring, even crossterm, with nushell or reedline.

kronberger-droid:
Good point, might be that i just lazily copy how this PR or how the reedline defautl impl  did it.
I did this one before my vacation so i don't know exactly right now.
I will take a look at it today.

kronberger-droid:
Oh sorry, now I see it. The trait is defined really weird.
You are right to question it, but i would investigate it in a follow-up and see if there are any regressions introduced when changing to nu_ansi_term.
You agree, or would you block those, because of it? @fdncred

fdncred:
I'd block it now. I don't want to introduce non-nu-ansi-term coloring. Maybe that's a mistake? Maybe crossterm colors already exist all over the place? I haven't looked. We specifically forked ansi-term for this purpose that's why I object to it. Hopefully crossterm color hasn't gone too far.

kronberger-droid:
We did definitely not introduce it.
This was in the Prompt trait before we touched anything.
So definitely an existing problem.
I have it now changed in reedline and will test how it behaves in nushell.
I will post them soon.

fdncred:
Thanks. Glad you're fixing this. Sorry to be a stickler on this. Wish I would've noticed it when it original creeped in.

kronberger-droid:
All good no problem. Its not that big of a deal to fix now.

kronberger-droid:
One thing that i noticed is that a non VT console like Windows legacy console would have used a `winapi` fallback provided by the `crossterm::Color`. We loose that.
But everything else already does it the same, so it would just have saved a small part of the prompt.
And if someone encounters this problem they can just set `use_ansi_coloring: false`

fdncred:
ok, i'm fine with that. we can land this when you're ready.

PR #1049, opened 2026-04-12
Make `ReedlineEvent::ExecuteHostCommand` return `Signal::HostCommand`  instead of `Signal::Success`

Juhan280:
This allow use to do special actions when the command is send through the keybinding. This also makes it possible to use more complex datatype by serializing it or string or just a key to a lookup table.


fdncred:
Thanks

PR #1051, opened 2026-04-14
refactor!: graceful handle ReedlineEvent::CtrlC

### LINKED ISSUE #1050: Retrieve buffer when receive CtrlC
Now I build the Reedline state like:

```rust
    let mut state = Reedline::create().use_bracketed_paste(true);

    if let Some(history) = make_file_history() {
        state = state.with_history(Box::new(history));
    }

    loop {
        let input = state.read_line(&prompt).expect("failed to read next line");
        let input = match input {
            Signal::Success(input) => input,
            Signal::CtrlC | Signal::CtrlD | Signal::ExternalBreak(_) | _ => {
                println!("Exit");
                break;
            }
        };
        let input = input.trim();
        // ...
    }
```

I'd like to see if there is pending buffered input when received `Signal::CtrlC | Signal::CtrlD`, but now I can't get it.

My goal is to handle input based on whether the user has content in the buffer.

If there is existing buffered input, I want Ctrl-C and Ctrl-D to potentially have no effect. I only want the program to exit if the input is empty.

However, Reedline currently does not provide me with a way to access or retrieve that specific state.

See full code at https://github.com/scopedb/scopeql/blob/7102e802de64e014aa69b780722fd2323e163986/scopeql/src/repl/entrypoint.rs#L90-L110

tisonkun:
This closes https://github.com/nushell/reedline/issues/1050

This aligns the manner of `ReedlineEvent::CtrlD`. I don't know what `self.deactivate_menus();` does so do my guess to put it in the `if self.editor.is_empty()` branch.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1241,8 +1241,8 @@ impl Reedline {
                 }
             }
             ReedlineEvent::CtrlC => {
+                self.deactivate_menus();
                 if self.editor.is_empty() {
-                    self.deactivate_menus();
                     self.editor.reset_undo_stack();
                     Ok(EventStatus::Exits(Signal::CtrlC))
                 } else {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1241,10 +1241,14 @@ impl Reedline {
                 }
             }
             ReedlineEvent::CtrlC => {
-                self.deactivate_menus();
-                self.run_edit_commands(&[EditCommand::Clear]);
-                self.editor.reset_undo_stack();
-                Ok(EventStatus::Exits(Signal::CtrlC))
+                if self.editor.is_empty() {
+                    self.deactivate_menus();
+                    self.editor.reset_undo_stack();
+                    Ok(EventStatus::Exits(Signal::CtrlC))
+                } else {
+                    self.run_edit_commands(&[EditCommand::Clear]);
+                    Ok(EventStatus::Handled)
+                }
             }
             ReedlineEvent::ClearScreen => {
                 self.deactivate_menus();


tisonkun:
cc @cptpiepmatz @fdncred in case if you have time to drop a review

src/engine.rs
```diff
@@ -1241,10 +1241,14 @@ impl Reedline {
                 }
             }
             ReedlineEvent::CtrlC => {
-                self.deactivate_menus();
-                self.run_edit_commands(&[EditCommand::Clear]);
-                self.editor.reset_undo_stack();
-                Ok(EventStatus::Exits(Signal::CtrlC))
+                if self.editor.is_empty() {
+                    self.deactivate_menus();
```
fdncred: I think we should move deactivate_menus() outside is_empty(). The way it is currently in this PR means for non-empty input, CtrlC will clear the buffer but leave any active menus enabled.

fdncred (changes requested):

src/engine.rs
tisonkun: @fdncred Thanks for your review! Updated.

fdncred (approved):
Thanks. LGTM!

tisonkun:
@fdncred I'm waiting for this PR to be released to use in my project.

Are there any other blockers, or is there some estimation I can expect?

fdncred:
I've been waiting for our patch release which happened just a few minutes ago. I'll probably land this PR tomorrow or the next day.

tisonkun:
Thanks for your information!

fdncred:
Thanks!

src/engine.rs
```diff
@@ -1242,9 +1242,13 @@ impl Reedline {
             }
             ReedlineEvent::CtrlC => {
                 self.deactivate_menus();
-                self.run_edit_commands(&[EditCommand::Clear]);
-                self.editor.reset_undo_stack();
-                Ok(EventStatus::Exits(Signal::CtrlC))
+                if self.editor.is_empty() {
+                    self.editor.reset_undo_stack();
```
Juhan280, line 1246: It is a bit late for the review. But I believe CtrlC should always clear undo stack

fdncred: @tisonkun if you want to submit another PR to make that change, we can land it pretty quickly.

Juhan280:
So, I have been running this for a few day now. Not sure I like this change. I would prefer to have a way to get original functionality back somehow. I want to know what others think about this.

CtrlC clears the prompt instead of making a new prompt (that was the point of this PR). But I find myself wanting to copy stuff from my previous prompt which I CtrlC-ed to the current prompt. But its no longer possible since the prompt is gone. Not sure how many people might have similar experience as my but I assume that would be very small number of people.

Because of the point mentioned above, CtrlC should probably not clear undo stack (already current behavior, but still mentioning this because I stated the opposite in my previous comment.)

Previously, CtrlC-ing from vi_normal mode would take me to vi_insert mode in a new prompt. But after this PR, CtrlC-ing the first time only clears the prompt keeping me in my current mode. Only after CtrlC-ing the second time (when the prompt is empty), it takes me to vi_insert mode. Not sure if I like this change either.But I can get used to this. CtrlC sometimes maintain current mode and sometimes doesn't. I think this is a confusing behavior. It should be consistent.

I made this comment to bring attention to the matter that it changes very core behavior the users interact with the most. So we should give it more thorough consideration whether we want this change or want to compromise.

fdncred:
I vote for consistency and making things easier for the user, which falls in line with your idea of not clearing and generally making thigs more consistent.

src/engine.rs
tisonkun: @fdncred See https://github.com/nushell/reedline/pull/1058

Sorry I miss this comment the very first time.

NotTheDr01ds:
@fdncred If I'm reading your comment correctly, you are in support of rolling back this change?  If so, I can take care of it.

@tisonkun Can you look for another solution for your use-case?  For Nushell, we do want to keep the command intact on <kbd>Ctrl</kbd>+<kbd>C</kbd>.

fdncred:
@NotTheDr01ds I'm up for whatever the right fix is. There's also #1058 although we're getting close to a release again without much time to test.

NotTheDr01ds:
@fdncred Since it's just reverting a 4-line change (cumulative, between the two commits), I'm not too worried.  It's also well tested behavior since it's been working this way for years ;-).

I've submitted #1081 for it if you can give it a quick glance.

fdncred:
@NotTheDr01ds i saw it. go for it if that's what you want to do. it's not really bothering me the way it is now.

tisonkun:
> @tisonkun Can you look for another solution for your use-case?

If Ctrl+C is hardcoded as before this PR in reedline, I think there is no escape way :O

PR #1052, opened 2026-04-15
deprecate `get_reedline_reedline_events`, `get_reedline_edit_commands` and `get_reedline_prompt_edit_modes`

Juhan280:
deprecate `get_reedline_reedline_events`, `get_reedline_edit_commands` and `get_reedline_prompt_edit_modes`

Display implementations of these enums were a hack. The strings were supposed represent how nushell (a downstream dependency) parses user inputs. And it often goes out of sync.
But now nushell uses the enum discriminants to display the values.


Juhan280:
Maybe we should deprecate other `get_reedline_*` functions too.

fdncred:
Thanks

PR #1055, opened 2026-04-18
fix(painting): avoid divide-by-zero in estimate_single_line_wraps for 0-column terminals

### LINKED ISSUE #842: Divide by zero when opening history_menu in narrow terminal
**Platform**: macOS
**Terminal software**: all kinds of terminals

Got these when pressing ctrl-r to open history menu in a narrow terminal:

```
▓  󰏫 Error:   × Main thread panicked. dev $!?   ├─▶ at /Users/brew/Library/Caches/Homebrew/
  │   cargo_cache/registry/src/index.crates.io-
  │   6f17d22bba15001f/reedline-0.35.0/src/
  │   painting/utils.rs:62:32
  ╰─▶ attempt to divide by zero
         0: 0x10f12a2b4 - _rust_begin_unwind
         1: 0x10f31a292 -
      core::panicking::panic_fmt::h2bf80d97c31b18
      c0
         2: 0x10f31a796 -
      core::panicking::panic_const::panic_const_d
      iv_by_zero::h6a4a844a42d12df2
         3: 0x10efffd56 -
      reedline::painting::utils::estimate_single_
      line_wraps::he2ce6a3ebf974298
         4: 0x10efde2ca -
      reedline::menu::list_menu::ListMenu::number
      _of_lines::h373afd30d8f7f4d2
         5: 0x10efde523 -
      reedline::menu::list_menu::ListMenu::printa
      ble_entries::hbb1243eafae1abf2
         6: 0x10efdf2ce -
      <reedline::menu::list_menu::ListMenu as
      reedline::menu::Menu>::update_working_detai
      ls::h7975cc53aa7ed67e
         7: 0x10efcfd63 -
      reedline::menu::ReedlineMenu::update_workin
      g_details::ha605d300e0726fd6
         8: 0x10f00c3e6 -
      reedline::engine::Reedline::repaint::hb3e14
      3826652f743
         9: 0x10f008be3 -
      reedline::engine::Reedline::read_line::hfb1
      16570925c290e
        10: 0x10e3ebe49 -
      nu_cli::repl::loop_iteration::h9698a4eb13f8
      568a
        11: 0x10e3e6e3e -
      std::panic::catch_unwind::hb6bb7f322d087986
        12: 0x10e4123d3 -
      nu_cli::repl::evaluate_repl::h4fcb0d54f7ae1
      239
        13: 0x10e3a9a3a -
      nu::run::run_repl::h24ced594c3095e12
        14: 0x10e39bb57 -
      nu::main::hf219ffa11f65f566
  help: set the `RUST_BACKTRACE=1` environment
        variable to display a backtrace.

```
## Steps to reproduce
1. set a long marker to history menu, I used to have ansi colors in the marker so width() will return a large number.
```nu
{
    name: history_menu
    only_buffer_difference: false
    # marker: (prompt_decorator $private_vars.prompt_symbol_color "light_blue" "")
    marker: "------------------------------------------------------------------------------------------"
    type: {
        layout: list
        page_size: 30
    }
    style: {
        text: $private_vars.menu_text_color
        selected_text: light_blue_reverse
        description_text: yellow
    }
}
```
2. Shrink the terminal to be narrower than the marker
3. Open the history menu

I don't get why we need to substract anything from the column number in

https://github.com/nushell/reedline/blob/871075e62b5703cfc584e4e02b14c2473f5b0b63/src/menu/list_menu.rs#L524-L528

and

https://github.com/nushell/reedline/blob/871075e62b5703cfc584e4e02b14c2473f5b0b63/src/menu/list_menu.rs#L180-L185

When I remove those, this bug disappeared and nothing seems to be broken

SAY-5:
### Problem

Opening the history menu (Ctrl-R) in a very narrow terminal, or during
a terminal resize that briefly reports `terminal_columns = 0`, panics
the main thread:

```
thread 'main' panicked at src/painting/utils.rs:62:32:
attempt to divide by zero
stack: ...
    reedline::painting::utils::estimate_single_line_wraps
 -> reedline::menu::list_menu::ListMenu::number_of_lines
 -> ListMenu::printable_entries
 -> <ListMenu as Menu>::update_working_details
```

`estimate_single_line_wraps` does the ceiling-division

```rust
let estimated_line_count = (estimated_width + terminal_columns - 1) / terminal_columns;
```

which is well-defined only for `terminal_columns >= 1`. Reported in
#842.

### Fix

Return `0` extra wraps up front when `terminal_columns == 0`, matching
the "no extra wraps" contract for the degenerate case. Everything
downstream (`estimate_required_lines`, `ListMenu::number_of_lines`)
tolerates a zero wrap count.

```diff
 pub(crate) fn estimate_single_line_wraps(line: &str, terminal_columns: u16) -> usize {
-    let estimated_width = line_width(line);
     let terminal_columns: usize = terminal_columns.into();
+    if terminal_columns == 0 {
+        return 0;
+    }
+    let estimated_width = line_width(line);

     // integer ceiling rounding division for positive divisors
     let estimated_line_count = (estimated_width + terminal_columns - 1) / terminal_columns;
```

Added `estimate_single_line_wraps_zero_columns_does_not_panic` (would
panic under the old code) plus a small rstest table covering the common
width/column combinations to guard against a regression.

Fixes #842

### DIFF
--- a/src/painting/utils.rs
+++ b/src/painting/utils.rs
@@ -53,10 +53,26 @@ pub(crate) fn estimate_required_lines(input: &str, screen_width: u16) -> usize {
 ///
 /// Does not account for any potential line breaks in `line`
 ///
-/// If `line` fits in `terminal_columns` returns 0
+/// If `line` fits in `terminal_columns` returns 0. A zero-width
+/// `terminal_columns` can be reported by terminals mid-resize or when
+/// the size is unknown; return 0 in that case rather than dividing by
+/// zero (see #842).
+///
+/// FIXME: The zero-column guard below papers over a caller bug, it
+/// doesn't solve it. `menu::list_menu::ListMenu::menu_required_lines`
+/// passes `terminal_columns.saturating_sub(indicator_width + count_digits)`,
+/// so on a terminal whose width is not greater than the indicator plus
+/// the entry-index digits this function receives 0 and every entry is
+/// reported as a single non-wrapping line. The real fix is to enforce a
+/// minimum viable column budget in `menu_required_lines` (or to stop
+/// subtracting the indicator width from the entry width). Tracked in
+/// #842 / #428; remove this comment once the caller is fixed.
 pub(crate) fn estimate_single_line_wraps(line: &str, terminal_columns: u16) -> usize {
-    let estimated_width = line_width(line);
     let terminal_columns: usize = terminal_columns.into();
+    if terminal_columns == 0 {
+        return 0;
+    }
+    let estimated_width = line_width(line);
 
     // integer ceiling rounding division for positive divisors
     let estimated_line_count = (estimated_width + terminal_columns - 1) / terminal_columns;
@@ -93,5 +109,27 @@ mod test {
             input != expected || matches!(result, Cow::Borrowed(_)),
             "Unnecessary allocation"
         )
+    }
+
+    /// Narrow-terminal regression: a zero-column terminal used to panic
+    /// with "attempt to divide by zero" inside the ceiling-division
+    /// expression (#842). Return 0 extra wraps instead.
+    #[test]
+    fn estimate_single_line_wraps_zero_columns_does_not_panic() {
+        assert_eq!(estimate_single_line_wraps("hello world", 0), 0);
+        assert_eq!(estimate_single_line_wraps("", 0), 0);
+    }
+
+    #[rstest]
+    #[case("", 80, 0)]
+    #[case("hello", 80, 0)]
+    #[case("abcdefghij", 5, 1)]
+    #[case("abcdefghijk", 5, 2)]
+    fn estimate_single_line_wraps_basic(
+        #[case] line: &str,
+        #[case] columns: u16,
+        #[case] expected: usize,
+    ) {
+        assert_eq!(estimate_single_line_wraps(line, columns), expected);
     }
 }


blindFS:
Hi, thanks for taking on this.

The fix is straightforward enough. However, as far as I recall, the root cause of this problem is quite complicated.

Would you please go over the conversations in #428 to see if you can tackle the underlying problems?

If not, I'd like to have a FIXME/HACK comment somewhere so we don't forget about the true pending issues.

SAY-5:
Thanks @blindFS. I traced the 0-column case back to `ListMenu::menu_required_lines` in `list_menu.rs` — it passes

```rust
terminal_columns.saturating_sub(indicator_width + count_digits(entry_index))
```

to `number_of_lines`, and on a terminal whose width is not greater than the indicator plus the entry-index digits that subtraction underflows to 0. The panic is the visible symptom of that caller-side underflow rather than a real terminal-has-0-columns condition.

I took option 2 from your review: kept the divide-by-zero guard in `estimate_single_line_wraps` (since it's the hot path that actually panics) and added a FIXME comment on the function pointing at `menu_required_lines` so the proper fix (enforce a minimum viable column budget, or stop subtracting the indicator width from the entry width) stays visible. The commit message spells that out too.

Happy to open a follow-up that rewrites `menu_required_lines` to never hand `number_of_lines` a column budget below some floor (e.g. 1) if you'd prefer I do it here or in a separate PR.

fdncred:
Thanks! Also, the follow up PR you mentioned would be great!

PR #1057, opened 2026-04-19
deps: `keybindings` crate is only used for `helix`

Juhan280:
Only use keybidnings crate for helix feature


fdncred:
ya, makes sense. thanks

Issue #556, opened 2023-03-18
Adding fish-like abbreviations to reedline


thomasgoulet:
Abbreviations are really nice to have within shells.

- They reduce the need to declare aliases or functions outside of reedline
- They would allow reedline's completion mechanism to work with the expanded version of the abbreviation. Any shell using reedline would only need to manage completion for the actual command and not aliases
- They allow for more readability for whoever sees the terminal (when screensharing or screenshotting)
- Long commands can be typed quickly and even modified if need be (change a flag or argument after expansion)

---

- `!!` and `!$` are already "abbreviations" in a way.
- Abbreviations should also be expanded when pressing `space`, not just `enter`
- From what I've seen regular expressions are way less common within abbreviations in fish and I don't think they would be needed here. At least not at first.
- There would need to be a mechanism to define if the abbreviation should be expanded only when at the start of the line or also in the middle. Middle of line abbreviations are less common in fish but I've seen a few people use them. Maybe only start of line could be supported at first.

## References

https://fishshell.com/docs/current/cmds/abbr.html

casedami:
Is there any progress on this? I would be interested in helping. 

could we do the following:

- add nushell parsing support for something like `abbr gs = git status`. should be somewhat similar to how the current parsing is done for `alias`
- store abbreviations in the nushell engine state
- when nushell starts the repl, pass abbreviations to reedline something like `with_abbreviations()`
- in `reedline/src/engine.rs` in `handle_editor_event` in the `ReedlineEvent::Edit(commands)` arm, see if the user has pressed space, and if so check if we can expand the last word before the space ie check if an abbreviation exists. this check could also prevent any expansion within an unclosed string for example

thomasgoulet:
@casedami Well played. I couldn't believe my eyes when I read the changelog.

PR #1060, opened 2026-04-21
feat: add support for fish-like abbreviations

casedami:
Following up on my comment in #556 i figured i would go ahead and get something working. This would be the initial step to getting abbreviations in nushell and would require ~~a parsing implementation~~ abbreviations to be stored in nushell config and then passed to reedline. Open to any comments or opinions as far as implementation goes but this is a feature that i would personally love to have in nushell.

I've essentially added a check during the following events that will check if the word before the cursor is an abbreviation and expand it if true:
- `Submit` and `SubmitOrNewline`
- `Enter` 
- `Edit` if the first command in the event is to insert a space

As for adding this to nushell, here's a simple example of how this might look in the nushell config:

```diff
diff --git a/crates/nu-cli/src/repl.rs b/crates/nu-cli/src/repl.rs
index c1fea4150..239153226 100644
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -414,6 +414,7 @@ fn loop_iteration(ctx: LoopContext) -> (bool, Stack, Reedline) {
         )))
         .with_quick_completions(config.completions.quick)
         .with_partial_completions(config.completions.partial)
+        .with_abbreviations(config.abbreviations)
         .with_ansi_colors(config.use_ansi_coloring.get(engine_state))
         .with_cwd(Some(
             engine_state
diff --git a/crates/nu-protocol/src/config/mod.rs b/crates/nu-protocol/src/config/mod.rs
index 37fa177a5..9769160e9 100644
--- a/crates/nu-protocol/src/config/mod.rs
+++ b/crates/nu-protocol/src/config/mod.rs
@@ -63,6 +63,7 @@ pub struct Config {
     pub hinter: HinterConfig,
     pub history: HistoryConfig,
     pub keybindings: Vec<ParsedKeybinding>,
+    pub abbreviations: HashMap<String, String>,
     pub menus: Vec<ParsedMenu>,
     pub hooks: Hooks,
     pub rm: RmConfig,
@@ -135,6 +136,8 @@ impl Default for Config {

             keybindings: Vec::new(),

+            abbreviations: HashMap::new(),
+
             error_style: ErrorStyle::default(),
             error_lines: 1,
             display_errors: DisplayErrors::default(),
@@ -218,6 +221,10 @@ impl UpdateFromValue for Config {
                     Ok(keybindings) => self.keybindings = keybindings,
                     Err(err) => errors.error(err.into()),
                 },
+                "abbreviations" => match HashMap::from_value(val.clone()) {
+                    Ok(abbreviations) => self.abbreviations = abbreviations,
+                    Err(err) => errors.error(err.into()),
+                },
                 "hooks" => self.hooks.update(val, path, errors),
                 "datetime_format" => self.datetime_format.update(val, path, errors),
                 "error_style" => self.error_style.update(val, path, errors),
```

Note, that the `is_inside_string_literal` function could potentially be used for a bug fix for

- https://github.com/nushell/nushell/issues/6971
- https://github.com/nushell/nushell/issues/13577 

As an aside, here's how im currently getting abbreviations with this hacky solution

```nu
let abbrevs = {
    ga: 'git add'
    gb: 'git branch'
    gc: 'git commit -v'
    gd: 'git diff'
    gdt: 'git difftool -d'
    gm: 'git merge'
    gr: 'git rebase'
    gR: 'git restore'
    gwa: 'git worktree add'
    gwr: 'git worktree remove'
    gwl: 'git worktree list'
    gx: 'git switch'
    la: 'ls -a'
    ll: 'ls -l'
    fg: 'job unfreeze'
    ou: 'overlay use'
    oh: 'overlay hide'
    ol: 'overlay list'
}

$env.config = {
    menus: [
        {
            name: abbr_menu
            only_buffer_difference: false
            marker: none
            type: {
                layout: columnar
                columns: 1
                col_width: 20
                col_padding: 2
            }
            style: {text: green, selected_text: green_reverse, description_text: yellow}
            source: {|buffer, position|
                let match = $abbrevs | columns | where $it == $buffer
                if ($match | is-empty) {
                    {value: $buffer}
                } else {
                    {
                        value: ($abbrevs | get $match.0)
                    }
                }
            }
        }
    ]
}

```

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -273,6 +273,59 @@ impl Editor {
         self.line_buffer.get_buffer()
     }
 
+    /// Check if a position in the buffer is inside an unclosed string literal
+    pub fn is_inside_string_literal(&self, position: usize) -> bool {
+        let buffer = self.get_buffer();
+
+        if buffer.is_empty() || position == 0 {
+            return false;
+        }
+        if !buffer.contains('"') && !buffer.contains('\'') {
+            return false;
+        }
+
+        let target_byte_pos = buffer
+            .char_indices()
+            .nth(position)
+            .map(|(byte_idx, _)| byte_idx)
+            .unwrap_or(buffer.len());
+
+        let bytes = buffer.as_bytes();
+        let mut in_single_quote = false;
+        let mut in_double_quote = false;
+        let mut escaped = false;
+        let mut byte_pos = 0;
+
+        for &byte in bytes {
+            if byte_pos > target_byte_pos {
+                break;
+            }
+
+            if escaped {
+                escaped = false;
+                byte_pos += 1;
+                continue;
+            }
+
+            match byte {
+                b'\\' => {
+                    escaped = true;
+                }
+                b'\'' if !in_double_quote => {
+                    in_single_quote = !in_single_quote;
+                }
+                b'"' if !in_single_quote => {
+                    in_double_quote = !in_double_quote;
+                }
+                _ => {}
+            }
+
+            byte_pos += 1;
+        }
+
+        in_single_quote || in_double_quote
+    }
+
     /// Edit the [`LineBuffer`] in an undo-safe manner.
     pub fn edit_buffer<F>(&mut self, func: F, undo_behavior: UndoBehavior)
     where
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -273,7 +273,7 @@ impl Editor {
         self.line_buffer.get_buffer()
     }
 
-    /// Check if a position in the buffer is inside an unclosed string literal
+    /// Check if `position` (a byte offset) is inside an unclosed string literal.
     pub fn is_inside_string_literal(&self, position: usize) -> bool {
         let buffer = self.get_buffer();
 
@@ -284,12 +284,6 @@ impl Editor {
             return false;
         }
 
-        let target_byte_pos = buffer
-            .char_indices()
-            .nth(position)
-            .map(|(byte_idx, _)| byte_idx)
-            .unwrap_or(buffer.len());
-
         let bytes = buffer.as_bytes();
         let mut in_single_quote = false;
         let mut in_double_quote = false;
@@ -297,7 +291,7 @@ impl Editor {
         let mut byte_pos = 0;
 
         for &byte in bytes {
-            if byte_pos > target_byte_pos {
+            if byte_pos > position {
                 break;
             }
 
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -2241,6 +2241,23 @@ mod test {
     #[case("'say \"hi\"'", 6, true)]
     // Position 0 is never inside a string
     #[case(r#""open"#, 0, false)]
+    // Cyrillic (2-byte UTF-8 chars): char boundary positions
+    #[case("Сегодня хороший день.", 14, false)]
+    #[case("Сегодня 'хороший' день.", 16, true)]
+    #[case("Сегодня 'хороший' день.", 31, false)]
+    #[case("'Сегодня хороший день.", 2, true)]
+    // Emoji (4-byte UTF-8 chars): char boundary positions
+    #[case("this is fire 🔥", 3, false)]
+    #[case("'hello there' 👋🏼", 13, false)]
+    #[case("'this is fire 🔥", 14, true)]
+    #[case("'hello there 👋🏼", 13, true)]
+    // Japanese (3-byte UTF-8 chars): char boundary and mid-char positions
+    #[case("今日はいい日だ。", 6, false)]
+    #[case("'今日はいい日だ。", 6, true)]
+    #[case("'今日はいい日だ。", 7, true)]
+    // Mid-char byte positions within multi-byte sequences are also handled correctly
+    #[case("Сегодня 'хороший' день.", 19, true)]
+    #[case("'hello there 👋🏼", 14, true)]
     fn test_is_inside_string_literal(
         #[case] buffer: &str,
         #[case] position: usize,
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -2215,4 +2215,38 @@ mod test {
         assert_eq!(bracket_result, expected_bracket);
         assert_eq!(quote_result, expected_quote);
     }
+
+    #[rstest]
+    // Not inside any string
+    #[case("hello world", 5, false)]
+    #[case("", 0, false)]
+    #[case("no quotes here", 0, false)]
+    // Closed double-quoted string — position is after the closing quote
+    #[case(r#""hello" world"#, 8, false)]
+    // Inside an unclosed double-quoted string
+    #[case(r#""hello world"#, 5, true)]
+    // Closed double-quoted string — position is inside it
+    #[case(r#""hello" world"#, 3, true)]
+    // Inside an unclosed single-quoted string
+    #[case("'hello world", 5, true)]
+    // Closed single-quoted string
+    #[case("'hello' world", 8, false)]
+    // Escaped quote does not open/close a string
+    #[case(r#"say \"hello"#, 6, false)]
+    // Escaped quote inside an open string — still inside
+    #[case(r#""say \"hello"#, 9, true)]
+    // Mixed: single inside double (single quote is literal)
+    #[case(r#""it's fine""#, 5, true)]
+    // Mixed: double inside single (double quote is literal)
+    #[case("'say \"hi\"'", 6, true)]
+    // Position 0 is never inside a string
+    #[case(r#""open"#, 0, false)]
+    fn test_is_inside_string_literal(
+        #[case] buffer: &str,
+        #[case] position: usize,
+        #[case] expected: bool,
+    ) {
+        let editor = editor_with(buffer);
+        assert_eq!(editor.is_inside_string_literal(position), expected);
+    }
 }
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -187,7 +187,7 @@ pub struct Reedline {
     // Engine Menus
     menus: Vec<ReedlineMenu>,
 
-    abbreviations: HashMap<Vec<u8>, Vec<u8>>,
+    abbreviations: HashMap<String, String>,
 
     // Text editor used to open the line buffer for editing
     buffer_editor: Option<BufferEditor>,
@@ -625,7 +625,7 @@ impl Reedline {
     /// A builder that adds abbreviations to the Reedline engine
     ///
     /// Overwrites any existing abbreviations with the same key.
-    pub fn with_abbreviations(mut self, abbreviations: HashMap<Vec<u8>, Vec<u8>>) -> Self {
+    pub fn with_abbreviations(mut self, abbreviations: HashMap<String, String>) -> Self {
         self.abbreviations.extend(abbreviations);
         self
     }
@@ -1767,8 +1767,8 @@ impl Reedline {
 
         let chars: Vec<char> = buffer.chars().collect();
         let (offset, suffix) = match submitted {
-            true => (0, ""),
-            false => (1, " "),
+            true => (0, ""),   // expand on <enter>
+            false => (1, " "), // expand on <space>
         };
 
         let mut word_start = cursor_position_in_buffer - 1;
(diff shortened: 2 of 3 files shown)


casedami:
@fdncred any feedback on this? the implementation is a little crude, it could be modified to more closely resemble how keybindings are defined

fdncred:
Sorry, it's hard to fine people to review reedline PRs. It would be good to have an example with it too.

casedami:
https://github.com/user-attachments/assets/ed5120ed-3fbb-49bb-a1b9-679e287942a5

no worries. i'm assuming you mean an example of it in action and i've included one here. just showing that it works anywhere in the buffer and doesn't expand when it appears as a subword. 

to produce this, i modified the basic example like so:

```diff
diff --git a/examples/basic.rs b/examples/basic.rs
index 825d954..dc0c680 100644
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -4,11 +4,13 @@
 // You can browse the local (non persistent) history using Up/Down or Ctrl n/p.

 use reedline::{DefaultPrompt, Reedline, Signal};
-use std::io;
+use std::{collections::HashMap, io};

 fn main() -> io::Result<()> {
+    let mut abbrevs = HashMap::new();
+    abbrevs.insert(String::from("ll"), String::from("ls -l"));
     // Create a new Reedline engine with a local History that is not synchronized to a file.
-    let mut line_editor = Reedline::create();
+    let mut line_editor = Reedline::create().with_abbreviations(abbrevs);
     let prompt = DefaultPrompt::default();

     loop {
```

fdncred:
I mean there should be a special example in the reedline examples folder. Let's not modify an existing one please. you can copy it and change it but let's keep the examples the way they are except for your new one.

What is the trigger to expand the abbreviation? Is it just the space after `ll` or are you hitting some key combination?

casedami:
can do. yes the trigger is space after `ll` as well as enter

fdncred:
I think it's cool. Let's see what copilot thinks of your changes.

src/engine.rs
```diff
@@ -1729,6 +1758,51 @@ impl Reedline {
...
+        let (offset, suffix) = match submitted {
+            true => (0, ""),   // expand on <enter>
+            false => (1, " "), // expand on <space>
+        };
+
+        let mut word_start = cursor_position_in_buffer - 1;
+        while word_start > 0 && !chars[word_start - 1].is_whitespace() {
+            word_start -= 1;
+        }
+        let word_end = cursor_position_in_buffer - offset;
```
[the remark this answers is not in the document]
casedami: if i understand correctly the case where `submitted == false && cursor_position_in_buffer < 1` would never result in an expansion because that occurs on an edit event, which would make `cursor_position_in_buffer >= 1`, or when the prompt is submitted, which would mean `submitted` would be true. regardless the early return for if `cursor_position_in_buffer == 0` should be added

src/engine.rs
```diff
@@ -1323,6 +1345,13 @@ impl Reedline {
             }
             ReedlineEvent::Edit(commands) => {
                 self.run_edit_commands(&commands);
+
+                // Check if a space was just inserted and try to expand abbreviations
+                if let Some(EditCommand::InsertChar(' ')) = commands.first() {
+                    if let Some(event) = self.try_expand_abbreviation_at_cursor(false) {
+                        return self.handle_editor_event(prompt, event);
+                    }
+                }
```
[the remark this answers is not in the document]
casedami: i'm unable to produce this behavior, as shown in the video example only one space is inserted. when i make this change _no expansion_ occers actually unless i press <enter>

src/core_editor/editor.rs
```diff
@@ -2168,4 +2215,38 @@ mod test {
         assert_eq!(bracket_result, expected_bracket);
         assert_eq!(quote_result, expected_quote);
     }
+
+    #[rstest]
```
fdncred, line 2219: I'd like to see some oddball Unicode strings in these tests like emojis and ZWJs and Unicode accented chars, maybe some Japanese or Chinese chars to make sure `is_inside_string_literal` is parsing Unicode correctly. Seems like it may need unicode-segmentation.

casedami: i had claude help me with this, let me know if the are any other cases youd like to see. for the ZWJ i went with :wave: with a skin tone

fdncred:
Thanks! Are you interested in adding the ability to use this functionality in nushell?

casedami:
Yeah I can work on that! Thanks!

fdncred:
Make sure to do `cargo update -p reedline` on the nushell repo, then inspect the cargo.lock and reject any changes that aren't reedline. Cargo likes to downgrade windows stuff for some reason. Thanks!

src/core_editor/editor.rs
```diff
@@ -273,6 +273,53 @@ impl Editor {
         self.line_buffer.get_buffer()
     }
 
+    /// Check if `position` (a byte offset) is inside an unclosed string literal.
+    pub fn is_inside_string_literal(&self, position: usize) -> bool {
```
blindFS, line 277: Hi @casedami, I'm afraid reedline isn't the right place for such checking. I mean different downstream shells might have different syntaxes of string literal. For example, in nushell, there's another form of string literal in backticks.

blindFS: Besides, this task is actually much more difficult than it seems because of [string interpolation](https://www.nushell.sh/lang-guide/chapters/strings_and_text.html#string-interpolation)

casedami: That's a good point, hadn't considered downstream shell usage. If you're more comfortable removing this func and leaving this as a problem to tackle in another way down the line then I can submit a PR that removes this. Would also apply to #1073

blindFS: @casedami Sorry, I didn't go over the linked issues in detail, according to my limited knowledge, the "ideal" way for such matter would be a protocol to query downstream apps for the string literal checking result, and they might be able to reuse the cached parsing results from the highlighting task.

blindFS: I mean, this func should be removed one way or another. I just don't know what impact it will land on the abbreviation feature, if just more false positives, acceptable IMHO.

casedami: Basically the change to the abbreviation feature would be that they now expand even when you're typing inside a string literal, so overall it's fine to remove it I would say

blindFS: > Basically the change to the abbreviation feature would be that they now expand even when you're typing inside a string literal, so overall it's fine to remove it I would say

Sure, let's do this please

fdncred: I'm not sure I agree @blindFS. reedline's first customer is nushell. We don't want abbreviations expanding in other places. What other work-arounds could we have here?

I'm fine with having a similar function in nushell but there should be some API for this, I think. Maybe passing a function pointer to reedline or some such thing, where the function is passed into reedline. I'm not sure how it should be really, but we should have something for this instead of just having abbreviations expand everywhere you type. People will hate that.

fdncred: Maybe we can have a trait. @blindFS, @casedami, do you think something like this would work?
```Rust
/// Determines whether a given byte position in the input is inside a string literal.
pub trait StringLiteralDetector: Send + Sync + 'static {
    /// Returns `true` if `position` (byte offset) is inside a string literal in `text`.
    fn is_inside_string_literal(&self, text: &str, position: usize) -> bool;
}

/// Default simple implementation (e.g. basic Rust-like double-quoted strings, no raw strings)
#[derive(Default, Clone)]
pub struct BasicStringDetector;

impl StringLiteralDetector for BasicStringDetector {
    fn is_inside_string_literal(&self, text: &str, mut position: usize) -> bool {
        // current algorithm
    }
}
```
Then in other crates/apps like nushell:
```Rust
use reedline::{Reedline, StringLiteralDetector};

// Custom detector for a different language (e.g. Python)
#[derive(Default)]
struct PythonStringDetector;

impl StringLiteralDetector for PythonStringDetector {
    fn is_inside_string_literal(&self, text: &str, position: usize) -> bool {
        // Implement Python rules: ", ', """, ''', r"", etc.
        // You can use a small state machine, regex, or even call tree-sitter if you want full power.
        todo!("Your language-specific logic")
    }
}

let mut editor = Reedline::create()
    .with_string_detector(PythonStringDetector::default());
```

blindFS: > What other work-arounds could we have here?
> 

Like I said, there should be something similar to how the highlighting is done, basically a communication protocol between reedline and the host program. I'm not familiar with the implementation details, but reedline is definitely not the "correct" place for that matter, just like you don't parse nushell code and highlight it in reedline directly.

As for the abbreviation feature, either we document it as experimental, or we revert that entirely until the string literal check is implemented.

blindFS: > Maybe we can have a trait. @blindFS, @casedami, do you think something like this would work?
> 
> ```Rust
> 
> /// Determines whether a given byte position in the input is inside a string literal.
> 
> pub trait StringLiteralDetector: Send + Sync + 'static {
> 
>     /// Returns `true` if `position` (byte offset) is inside a string literal in `text`.
> 
>     fn is_inside_string_literal(&self, text: &str, position: usize) -> bool;
> 
> }
> 
> 
> 
> /// Default simple implementation (e.g. basic Rust-like double-quoted strings, no raw strings)
> 
> #[derive(Default, Clone)]
> 
> pub struct BasicStringDetector;
> 
> 
> 
> impl StringLiteralDetector for BasicStringDetector {
> 
>     fn is_inside_string_literal(&self, text: &str, mut position: usize) -> bool {
> 
>         // current algorithm
> 
>     }
> 
> }
> 
> ```

I was hoping we can reuse the parsing results from the highlighting task on the nushell side. I'm not sure if that trait will do the trick.

fdncred: That may work for nushell but I'm not sure how it would work for other consumers of reedline. The reedline function needs to take some standard rust type and not some nushell type or flatshapes and things like that. 

@casedami Can you please investigate how this might work in nushell, using the nushell highlighting mechanism to figure this out? 

I think the point blindfs is making is we shouldn't have to parse it twice, once in nushell and once in reedline. We should be able to just parse it in the reedline consumer engine and pass it to a trait or something in reedline.

fdncred: Another thought is to extend reedline's Highlighter trait.
https://github.com/nushell/reedline/blob/8b3028cd8834dc6502f74bf105a54e89897e9002/src/highlighter/mod.rs#L10-L15

with
```Rust
fn is_inside_string_literal(&self, line: &str, position: usize) -> bool {
        false // default for simple highlighters
    }
```
and then from nushell implement it in NuHighlighter like highlight_syntax() is implmeneted and similar with other languages.

Thoughts?

blindFS: > Another thought is to extend reedline's Highlighter trait.
> 
> https://github.com/nushell/reedline/blob/8b3028cd8834dc6502f74bf105a54e89897e9002/src/highlighter/mod.rs#L10-L15
> 
> 
> 
> with
> 
> ```Rust
> 
> fn is_inside_string_literal(&self, line: &str, position: usize) -> bool {
> 
>         false // default for simple highlighters
> 
>     }
> 
> ```
> 
> and then from nushell implement it in NuHighlighter like highlight_syntax() is implmeneted and similar with other languages.
> 
> 
> 
> Thoughts?

I've thought about that too, it would be a breaking change for all downstream programs I assume? Personally I've no issue with that approach.

Oh, not necessarily a breaking change if we can provide a default implementation.

fdncred: ya, we just provide a default implementation so it's not breaking. @casedami can you try to implement this please?

casedami: Sure I can do that

blindFS: > Sure I can do that

FYI, on the nushell side, it should be as simple as

1. flatten the parsing result AST
2. check the flat expression whose span contains the given position

In order to avoid duplicated computation, you probably need to modify the implementation of `nu_highlight` and store the intermediate flattened expressions somewhere in the loop.

casedami: Ok thanks. I'll try to get to it today/tomorrow and let y'all know if I have any issues or questions that come up

casedami: moved to #1074

PR #1065, opened 2026-05-04
fix(vi): move cursor left when exiting insert mode with Escape

### LINKED ISSUE #694: vi mode cursor position isn't vi-like
2 main issues I can observe:
1. Escape doesn't move the cursor back.

    For example, interleaving `i` and escape moves the cursor back one by one (in both vim and zsh vi mode)
3. Moving to the last character puts the cursor _after_ the last character, not on top of it. **Note:** `$` is not the only way to achieve this

    For example, having `asdf` and the cursor at the front (position 0), and pressing any one of `w`, `W`, `E`, `e`, `$` should put the cursor on top of `f`.

sim590:
## Summary

- Move the cursor one position left when pressing Escape to exit insert mode, matching standard Vi behavior (insert mode cursor is between characters, normal mode cursor is on a character)
- Adjust `is_cursor_at_buffer_end()` to treat "cursor on the last character" as "at buffer end" in Vi normal mode, preserving prefix history search

## Context

This fix was previously implemented in #699 but reverted in #773 because it broke prefix history search (#772): after pressing Escape, the cursor moved left, causing `is_cursor_at_buffer_end()` to return false, which disabled prefix search when pressing `k`.

This PR addresses that regression by making `is_cursor_at_buffer_end()` mode-aware. In Vi normal mode, the cursor sitting *on* the last character is equivalent to being at the end of the buffer.

A similar fix was proposed in #986 but did not address the prefix search regression.

Closes #694 (pt 1)

## Test plan

- Added `esc_from_insert_mode_moves_cursor_left` test
- Added `esc_from_normal_mode_does_not_move_cursor` test
- Added `test_is_cursor_at_buffer_end_vi_normal_mode` test
- Added `test_is_cursor_at_buffer_end_insert_mode` test
- All 843 tests pass, clippy clean

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -306,7 +306,16 @@ impl Editor {
     }
 
     pub(crate) fn is_cursor_at_buffer_end(&self) -> bool {
-        self.line_buffer.insertion_point() == self.get_buffer().len()
+        let pos = self.line_buffer.insertion_point();
+        let len = self.get_buffer().len();
+        if pos == len {
+            return true;
+        }
+        // In Vi normal mode the cursor sits *on* the last character rather
+        // than after it.  Treat that position as "at buffer end" so that
+        // prefix history search still triggers after pressing Escape.
+        matches!(self.edit_mode, PromptEditMode::Vi(PromptViMode::Normal))
+            && self.line_buffer.grapheme_right_index() == len
     }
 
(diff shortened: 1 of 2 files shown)


sim590:
Closing in favor of #1069, which implements a more comprehensive fix for the Vi normal mode cursor invariant as discussed in #1066.

Issue #621, opened 2022-04-18
Vi mode seems incomplete


txtyash:
### Related problem

Many vim bindings don't work. I come from fish where setting vi mode enables most of the necessary vim shortcuts but seems like `Shift+c` & `Shift+s` don't work. I'm not sure what else doesn't work because switched to nu just a few minutes ago.

### Describe the solution you'd like

Add `Shift+c` & `Shift+s` bindings. Also add a vim section in the documentation. Is there a way to modify vim bindings?

Also, I think `Ctrl+f` is a nice binding to accept entire completion but there should be a `Ctrl+w` binding to accept only a single word completion.

LordMZTE:
If I may add something, it seems like we also don't implement replace mode (Shift+R) at all.

fdncred:
I've said this in the other issues about VI keybinds. If someone wants proper VI keybindings in reedline, they should implement it with something like https://github.com/ulyssa/modalkit and totally replace how reedline currently handles these keybindings.

kronberger-droid:
closing in favor of #1221 if i forgot to add something just leave a comment.

PR #1067, opened 2026-05-05
feat(vi): support deleting chars backward with X

sim590:
**Summary**

Add support for the `X` command in vi normal mode, which deletes the character before the cursor and places it in the cut buffer, mirroring the behavior of `X` in vim.

This partially addresses #621.

**Changes**

- Parse `X` as `Command::DeleteCharBackward` in the vi command parser
- In normal mode, emit `MoveLeft` + `CutChar` to delete the character before the cursor and place it in the cut buffer
- In visual mode, emit `CutSelection` (consistent with `x` behavior)
- Add parser tests for `X` and `2X`

**Known limitation**

When used with a multiplier (e.g. `3X`), each character is cut individually, so only the last deleted character remains in the cut buffer. This is a pre-existing limitation shared with `x` and will be addressed in a follow-up PR.


kronberger-droid:
Parser side looks right.

One correctness issue: `MoveLeft + CutChar` doesn't guard against the cursor not moving. So we have two cases where it diverges from Vim, at pos 0 it cuts the char on cursor and on end of a line in multi-line it would cut `\n` and merge the lines.

`EditCommand::Backspace` already gets this right via `line_buffer::delete_left_grapheme`. Could you add an `EditCommand::CutCharLeft` mirroring `CutChar` but with that goard and map `DeleteCharBackward` to it?

Two editor-level tests would be also good, for pos 0 in single line and start of line 2 in multi-line.

If you got that I would be fine to land it.

sim590:
OK. I got that.

Now another thing I just realized with my current implementation is that `X` in visual selection should rather cut the whole lines that are covered by the selection. So, a simple `Editor::cut_selection_to_cut_buffer` is not correct. The proper behaviour should be to get the range of the selection and extend both the lower and upper bounds to respectively the begening of the first line and the end of the last line in the selection range.

Do you agree with this addition?

sim590:
@kronberger-droid: I did the addition for visual selection behaviour for `X` in b472620 and 68616b2. The approach is to generalize code of `current_line_range` into `line_range_for(usize)` which returns the line range for the position passed as argument. Then, the visual selection is easily extended to the full lines. It works correctly. I put up three more tests for that.

kronberger-droid:
Hmm do you think you have to define a new edit command to achieve the guard?
Before you built the CutCharLeft in the mode out of primitives, now you have defined a new one. I would love if you could try to achieve the same, but with the primitives.

Thanks for the effort!

sim590:
The issue is that neither `MoveLeft` nor `Backspace` currently guard against crossing newlines. They both use `grapheme_left_index()` unconditionally. So composing `MoveLeft + CutChar` doesn't work at line boundaries.

I see a few options:

1. Add a `current_line: bool` parameter to `MoveLeft` (like `MoveRightUntil` already has) and use `MoveLeft { select: false, current_line: true } + CutChar`. This would not introduce any regression since the default would remain `false`.
2. Modify `Backspace` to copy into the cut buffer. However, this would regress insert mode and emacs mode, where backspace should not fill the cut buffer. Since `editor.rs` is mode-agnostic, there's no way to conditionally enable this.
3. Keep `CutCharLeft` as a dedicated command. This mirrors the existing `Delete`/`CutChar` pattern: `Delete` removes without copying, `CutChar` removes and copies to the cut buffer. Similarly, `Backspace` removes left without copying, `CutCharLeft` removes left and copies. It also naturally handles the visual-mode linewise cut behavior.

Which approach would you prefer?

kronberger-droid:
Thanks for the rundown.
It seems move left should actually guard for newline.
There are two different ways a newline can occur. Either you are in single line mode, where just move left and cut would be fine, since it just wraps the text and is essentially one line.
And there is multi line mode where if we want to provide a vim like experience it has to guard against newlines anyways.
And those states should be available in the editor.

Also I think if you want to make 3X work you will have to use something like option 1 or making MoveLeft guarding anyways right?

I am now working on an idea for much stronger primitives in EditCommand which will make these types of problems much easier to handle. But until then for me its just important to be sure about the intent of an EditCommand so its easy to replace them if we have stronger primitives. Until then I am fine with what works best for count prefixes (3X etc.)

Still important to have a more complete version as a baseline. Thanks!

sim590:
Based on your feedback, I'll switch to option 1: adding `current_line: bool` to `MoveLeft` and composing `MoveLeft { select: false, current_line: true }` + `CutChar` for normal mode. This removes the need for `CutCharLeft`.

Note that `3X` will only keep the last deleted character in the cut buffer, but this is a pre-existing limitation shared with `3x` (`apply_multiplier` repeats `CutChar` N times, each overwriting the register). I'd rather fix both together in a follow-up PR.

The remaining open question is visual mode: `X` in Vim cuts entire lines covered by the selection (linewise), not just the selection itself. This vi-specific behavior needs access to the editor state to extend the selection to line boundaries, but `command.rs` only returns `Vec<ReedlineOption>`. Would you prefer I add a `CutSelectionLinewise` primitive for now, or leave visual mode linewise behavior for a follow-up?

kronberger-droid:
Hmm I would avoid a view of the editor in the modes. (against what I mentioned in #960)

If you want you can propose a followup to fix the register behavior for multiple actions, but I a can't promise I will merge it, since it seems without some better primitives this will be just hacky.
But maybe you find a beautiful solution.

It would be nice to still add tests for the expected behavior we can't implement right now and skip them.
This would be a valueable followup.
Thanks!

sim590:
~~Just to clarify on the visual mode question: since `command.rs` would only emit an `EditCommand` without needing access to editor state, would adding a `CutSelectionLinewise` primitive be acceptable? It would live entirely in `editor.rs` (extending the selection to line boundaries and cutting), and the vi module would just emit it the same way it currently emits `CutSelection`. Or would you prefer to leave visual mode linewise behavior for a follow-up?~~

EDIT: This is irrelevant if we keep `CutCharLeft`.

sim590:
After trying the implementation for `MoveLeft` + `CutChar`, I came full circle and realized what you said at first: it cannot be done that way because of the beginning of the line edge case where `CutChar` would cut characters even though the cursor didn't move. So, in the end, we need to keep the current implementation.

kronberger-droid:
> After trying the implementation for `MoveLeft` + `CutChar`, I came full circle and realized what you said at first: it cannot be done that way because of the beginning of the line edge case where `CutChar` would cut characters even though the cursor didn't move. So, in the end, we need to keep the current implementation.

Do I understand that correctly that you would then stay with this implementation? Just tell me when I should take a closer look.

sim590:
Yes, that is correct. I think that until the better primitives you are working on are integrated, we cannot do without a new primitive `CutCharLeft` in order to correctly handle the beginning of line edge cases. Therefore, the current implementation is the correct one, I believe. You can take a closer look.

sim590:
Now that #1100 was merged, I will rework this PR soon.

sim590:
Here's my plan using the new verb+target vocabulary:

**Normal mode:** `X` maps to:
```rust
Cut { target: Grapheme { direction: Backward, current_line: true }, granularity: CharWise }
```
This requires adding a `current_line: bool` field (default `false`) to `MotionTarget::Grapheme`. When `true`, `resolve_motion` clamps to `start_of_line`/`end_of_line`. `resolve_head` already clamps `Grapheme` for `Move`/`Extend` based on `RestPolicy`: this is a separate mechanism that stays as is. Adding the clamp in `resolve_motion` gives operators (`Cut`/`Copy`/`Erase`) the same line-boundary guard. No new `CutCharLeft` primitive needed.

**Visual mode:** `X` maps to:
```rust
CutSelection { granularity: LineWise }
```
This requires adding a `granularity: Granularity` parameter to the existing `CutSelection` variant and dispatching it through `operate()`, which already handles the linewise snap. Existing callers of `CutSelection` would switch to `CutSelection { granularity: CharWise }` for backward compatibility.

**Multiplier (`3X`):** The cut buffer only keeps the last character, same pre-existing limitation as `3x`. I'll address both in a follow-up PR.

**End result:** `CutCharLeft` is removed, everything expressed with the new primitives. Does this align with your vision?

kronberger-droid:
Actually I think it's fine to keep CutCharLeft, since its line handling is unconditional, and `X` is its own atomic edit (not a motion, not governed by `whichwrap`), so it shouldn't follow `cross_line_cursor` (see #1109) anyway.

In short: `dh` != `X`, thus it makes sense to make it its own `EditConmmand`.

The motions that do vary in vim are the grapheme/word/.. ones.
We'll just gate those on the `cross_line_cursor` flag and decide it globally, rather than replicating vim's per-key `whichwrap` granularity.

`granularity` on `CutSelection` makes sense, keep that.

sim590:
I pushed the changes. Tell me what you think.

src/core_editor/editor.rs
```diff
@@ -2045,6 +2061,75 @@ mod test {
         assert_eq!(editor.get_buffer(), "This is a \r\n test");
     }
```
kronberger-droid, line 2827: Maybe add a positive test for cut_char_left?

kronberger-droid: Also If you want a defensive test for cut char left with a selection present.

src/core_editor/editor.rs
```diff
@@ -891,6 +895,14 @@ impl Editor {
         }
     }
 
+    fn cut_char_left(&mut self) {
```
kronberger-droid, line 1265:
```suggestion

    /// Cut the grapheme left of the caret into the cut buffer (vi `X`).
    ///
    /// Intended for vi normal mode only, where there is no selection, since
    /// visual `X` is dispatched as a linewise `CutSelection`. Bound directly
    /// elsewhere, a selection could exist, so clear it first: `X` always means
    /// "delete the grapheme before the caret," and clearing avoids leaving a
    /// stale anchor pointing into the mutated buffer. The cut is clamped to
    /// the current line so `X` never crosses a terminator (unconditional,
    /// unlike the `cross_line_cursor`-gated clamp in `resolve_head`).
    fn cut_char_left(&mut self) {
        self.clear_selection();
```

src/core_editor/editor.rs
sim590: I just realized you commented. I'll have a look at the comments and follow-up soon!

kronberger-droid: Great thanks. Now that the release is out we can try to merge it soon and then bump nushells reedline so we can test it.

sim590: I just added the two tests in [a9fc508](https://github.com/nushell/reedline/pull/1067/commits/a9fc5087344c0293ccefa806982f5744eae8f20f). Your comment was also added verbatim in the branch in [2cbc43a](https://github.com/nushell/reedline/pull/1067/commits/2cbc43ad66f866024312c48f0b73a17db8f4394e).

kronberger-droid:
Hey, sorry for the wait I was on vacation.
If you get the CI green again I think we are ready to merge.

sim590:
Hey! My fault, I didn't pay attention to the CI bot output. The format should be fixed now. The bot doesn't seem to run on the commit though. I'm not sure how to trigger it. May be I just need to wait for a bit.

kronberger-droid:
All good, for new contributors we have to allow the CI to run.
I forgot and only noticed yesterday.
So not your fault.
After we land this, future ones should run automatically i think.

kronberger-droid:
@sim590 Typos still. Sorry.

sim590:
@kronberger-droid: I see! This time should be the right one, now! :smiley_cat:

PR #1068, opened 2026-05-08
refactor(core_editor): extract grapheme boundary helpers into free functions

kronberger-droid:
Extracts grapheme-boundary computation out of `LineBuffer` into a sibling `graphemes` module as free functions over `&str`.
The two affected `LineBuffer` methods become one-line wrappers and behavior is unchanged.

The existing methods only ever touched `self.lines`, so promoting them to free functions gives one source of truth for grapheme math reedline-wide.

Side benefit: deduplicates the `next_back().map(...).unwrap_or(0)` pattern in `word_right_end_index` and `big_word_right_end_index`.

Also unblocks helix-mode operations (#1061 and follow-ups) that need to operate on `&str` rather than `&LineBuffer`.

14 unit tests cover ASCII, multi-byte UTF-8, CJK, combining marks, ZWJ emoji sequences, and round-trip invariants. Full suite (835) unchanged.

### DIFF
--- a/src/core_editor/graphemes.rs
+++ b/src/core_editor/graphemes.rs
@@ -1,0 +1,141 @@
+use unicode_segmentation::UnicodeSegmentation;
+
+/// Byte index of the next grapheme boundary at or after `pos`.
+///
+/// Returns `buf.len()` if there is no grapheme after `pos`.
+///
+/// # Panics
+///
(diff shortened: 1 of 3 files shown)


fdncred:
nice cleanup! Thanks!

Issue #1066, opened 2026-05-04
vi: cursor position semantics are not mode-aware (root cause of multiple off-by-one bugs)


sim590:
## Problem

Reedline's `LineBuffer::insertion_point` is a plain byte offset with no mode-aware bounds checking. In Vi, the two modes have fundamentally different cursor semantics:

- **Insert mode** — cursor is *between* characters; valid range is `0..=len()`
- **Normal mode** — cursor is *on* a character; valid range is `0..last_grapheme_start`

Reedline makes no such distinction. The cursor can always reach `len()` regardless of mode, causing a cascade of off-by-one bugs.

## Affected issues

All of the following share this single root cause:

- #694 — Escape doesn't move cursor back; cursor can go past last character in normal mode
- #766 — `MoveWordRightEnd` is off-by-1 in insert mode
- #788 — Multiple Vi keybinding inconsistencies (pts 1, 4, 5, 6, 7)
- #893 — Shift+arrow selects two characters instead of one
- #772 — Prefix history search broken after #699 (which was a partial fix for the same issue)

## Affected code paths

Every function that can place the cursor at `len()` is incorrect when in Vi normal mode:

| Function | File | Problem |
|---|---|---|
| `set_insertion_point` | `line_buffer.rs:71` | No bounds check |
| `move_to_end` / `move_to_line_end` | `line_buffer.rs:132-139` | Always goes to `len()` |
| `move_right` | `line_buffer.rs:453` | `l` can move past last character |
| `set_buffer` | `line_buffer.rs:81` | After history load, cursor lands at `len()` |
| All `*_word_right_*` | `line_buffer.rs` | Fallback to `len()` instead of last grapheme |
| `previous_history` | `engine.rs:1548` | Calls `move_to_end()` after loading history |

### Approach A — Centralized clamping (less invasive)

Add a `clamp_cursor_for_vi_normal()` method in `Editor` and call it from `run_edit_command()` whenever the mode is Vi normal, and after history navigation / buffer changes in the engine.

```rust
fn clamp_cursor_for_vi_normal(&mut self) {
    let len = self.get_buffer().len();
    if !self.get_buffer().is_empty() && self.insertion_point() >= len {
        let last = self.line_buffer.grapheme_left_index_from_pos(len);
        self.line_buffer.set_insertion_point(last);
    }
}
```

**Pros:** minimal API changes, single enforcement point, fixes all symptoms at once
[...]

PR #1069, opened 2026-05-08
fix(vi): enforce cursor invariant in normal mode

### LINKED ISSUE #694: vi mode cursor position isn't vi-like

sim590:
## Summary

In Vi normal mode, the cursor sits *on* a character and must not go past the last grapheme. This was not enforced, causing the cursor to land one position past the end of the line after various operations (movements, deletions, paste, undo, etc.).

This PR implements the approach discussed in #1066 (Approach C1): the Vi mode owns its cursor invariant and emits `EditCommand::ClampCursorToNormalMode` after each editing command that stays in normal mode. `LineBuffer` remains fully mode-agnostic.

## Changes

- Add `EditCommand::ClampCursorToNormalMode`, implemented in `Editor::clamp_cursor()`
- Emit `ClampCursorToNormalMode` after every Vi normal mode editing command in `command.rs` (`to_reedline` and `to_reedline_with_motion`)
- Add `clamp_cursor()` guards on rightward movement methods in `Editor` (`move_right`, `move_to_end`, `move_to_line_end`, `move_word_right*`, `move_big_word_right*`)
- Guard `set_buffer` so that history load, completion accept, and paste also respect the invariant
- Move cursor left on Escape (insert to normal transition), matching Vim behavior
- Propagate edit mode to `Editor` on Escape event so the cursor invariant is enforced immediately
- Adjust `is_cursor_at_buffer_end()` to treat "cursor on last grapheme" as "at buffer end" in Vi normal mode, preserving prefix history search and inline hints

## Test plan

- 868 tests pass, clippy clean
- New tests verify that normal mode commands (`x`, `p`, `P`, `u`, `dd`, `D`, `~`, `r`, `dw`, `db`, `d$`, `yw`) emit `ClampCursorToNormalMode`
- New tests verify that insert-transitioning commands (`i`, `a`, `cc`, `cw`, `s`) do *not* emit it
- Existing parser tests updated to expect the new command in their output

Closes #694 (cursor past end in normal mode)
Relates to #1066

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -21,3 +21,5 @@ l3ine = "l3ine"
 4should = "4should"
 wr5ap = "wr5ap"
 ine = "ine"
+# For testing multibyte cursor clamping
+caf = "caf"
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -201,6 +201,9 @@ impl Editor {
             EditCommand::CutTextObject { text_object } => self.cut_text_object(*text_object),
             EditCommand::CopyTextObject { text_object } => self.copy_text_object(*text_object),
         }
+
+        self.clamp_cursor();
+
         if !matches!(command.edit_type(), EditType::MoveCursor { select: true }) {
             self.clear_selection();
         }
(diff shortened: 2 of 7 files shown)


kronberger-droid:
@sim590 We are close to a release on Sa, afterwards I will take a look into all your Vi improvements so we can land them if they provide the right functionality, without breakage, in order to have a testable version for future changes. Thanks for your effort.

sim590:
This PR is unnecessary since #1100 was merged.

PR #1071, opened 2026-05-11
Additive `InputMode` / `OutputMode` enums for menu settings

### LINKED ISSUE #1006: Selecting from a menu keeps text after the cursor
**Platform** macOS
**Terminal software** ghostty

This seems like a bug to me. I would expect it to replace full text in the REPL.

## Steps to reproduce
 
1. Type anything in the REPL
2. Move the cursor somewhere in the middle
3. Press a hot-key to invoke any menu, customized ones or the history menu
4. Select any item

## Screenshots/Screencaptures

Image

Image

kronberger-droid:
Adds `InputMode` and `OutputMode` enums to `MenuSettings` with their own builder methods.
When set, they override `only_buffer_difference` (kept for back-compat) and when unset, the bool path is preserved unchanged.
Decouples menu input (what the completer receives) from output (what range of the buffer gets replaced on selection) if needed, which is the structural coupling behind #1006.

```rust
#[non_exhaustive]
pub enum InputMode {
    Diff,            // ≡ only_buffer_difference: true
    CursorPrefix,    // ≡ only_buffer_difference: false
    FullBuffer,      // NEW — exposes post-cursor content to the completer
}

#[non_exhaustive]
pub enum OutputMode {
    SuggestedSpan,   // ≡ Suggestion::span as-is (default)
    FullBuffer,      // overwrite 0..buffer.len()
    ExtendToEnd,     // keep span.start, force end = buffer.len()
}
```

`MenuSettings::effective_input_mode()` resolves enum-over-bool precedence in one place and all four menu impls (`Columnar`/`Description`/`Ide`/`ListMenu`) go through it.
`replace_in_buffer` takes `Option<OutputMode>`, also `None` and `Some(SuggestedSpan)` are equivalent.

`examples/demo.rs` opts the history menu into `OutputMode::FullBuffer`, demonstrating the pattern that fixes #1006 once nushell exposes the corresponding config field.

**Out of scope:**
- nushell config integration
- migration of other built-in menus
- `HistoryCompleter::create_suggestion` span update needed before HistoryMenu can use `InputMode::FullBuffer` (doc-comment warning added).

**Tests:** precedence table for `effective_input_mode`, `FullBuffer` case for `completer_input`, two cases for `replace_in_buffer` with explicit `OutputMode` variants.
Full suite passes (843 + 28 doctests).

Manually verified via `cargo run --example demo`: opened history menu mid-buffer, filtered, selected and the whole line replaced.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -16,7 +16,7 @@ use {
 
 #[cfg(not(any(feature = "sqlite", feature = "sqlite-dynlib")))]
 use reedline::FileBackedHistory;
-use reedline::{CursorConfig, MenuBuilder};
+use reedline::{CursorConfig, MenuBuilder, OutputMode};
 
 fn main() -> reedline::Result<()> {
     println!("Ctrl-D to quit");
@@ -102,7 +102,9 @@ fn main() -> reedline::Result<()> {
             ColumnarMenu::default().with_name("completion_menu"),
         )))
         .with_menu(ReedlineMenu::HistoryMenu(Box::new(
-            ListMenu::default().with_name("history_menu"),
+            ListMenu::default()
+                .with_name("history_menu")
+                .with_output_mode(OutputMode::FullBuffer),
         )));
 
     let edit_mode: Box<dyn EditMode> = if vi_mode {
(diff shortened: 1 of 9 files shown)


fdncred:
Yup, pretty much inline with what I was thinking based on our other conversation. Thanks!

PR #1072, opened 2026-05-13
fix: apply abbrev expansion before space is inserted

casedami:
It seems that one of Copilot's suggestions erroneously made it into #1060 which made it so that you needed to enter two spaces before the abbreviation expanded. this changes that back to the original intended behavior where the abbreviation expands on the initial space.


casedami:
@fdncred

fdncred:
Thanks

PR #1078, opened 2026-05-16
Restore "Fix move to line start in multi-line history entries"

### LINKED ISSUE #1075: Appended completion is erased from line_buffer when pressing any key
**Platform** Linux, affects all platforms.
**Terminal software** kgx

When performing a command and going back in history to the same command and trying to select an item from the autocompletion menu, the selected/already appended item is removed when any key is pressed after the selection. 

## Steps to reproduce (with Nushell)
1. ls / [enter] # Perform some command
2. [up] # Scroll back to the same command
3. [tab] # Open the autocompletion menu on 'ls /'
4. [enter] # Select the first completion suggestion, doesn't matter
5. [any key, g] # Press any key and notice how the selected suggestion disappears.
6. ls / [tab] # Try to autocomplete 'ls /' by manually typing it rather than selecting it from the history
7. [enter]
8. [any key, g] # Notice how the pressed key is appended to the end of the selected completion item.

## Steps to reproduce (without Nushell using cargo run --example completions)
1. l [enter]
2. [up]
3. [tab]
4. [enter]
5. [any key, g] # The input should now correspond to "lg" instead of "loging"
6. l [tab]
7. [enter]
8. [any key, g] # The input should now be "loging"

This bug has persisted for years and it has honestly driven me nuts, especially since I wasn't able to find any way to trigger the bug continuously before I realized that the bug was only able to be triggered if the current command to be autocompleted was selected from the history rather than being typed out manually.

It seems like the error happens in src/engine.rs in the run_edit_commands function, where self.history_cursor.string_at_cursor() returns the copy of line_buffer that hasn't been updated with the new value after having selected the autocompletion entry. Because of this, the line_buffer gets replaced with the older value which didn't contain the selected entry.

I was able to fix this issue in specific and pass all unit test by creating a new line right after the aforementioned string initialization and writing this line:
```{rust}
let string = self.editor.line_buffer().get_buffer().to_string();
```
i.e. overwriting the string with the updated line_buffer value. I'm unsure if this breaks anything else, especially considering there has been many PRs (https://github.com/nushell/reedline/pull/97, https://github.com/nushell/reedline/pull/584, https://github.com/nushell/reedline/pull/704) reverting and restoring this functionality.

ymcx:
Closes https://github.com/nushell/reedline/issues/1075

This PR restores the changes reverted in https://github.com/nushell/reedline/pull/704 and originally submitted in https://github.com/nushell/reedline/pull/584.

The issue I was personally having with this reversion is that if you try to perform completion on a line from the history, it doesn't work correctly: if you for example type 'l' and press enter and then scroll up to the last line using the arrow up key and select a completion suggestion by pressing tab and enter, the appended completion will be erased upon the pressing of any key.

The original PR was reverted due to the added unit tests working incorrectly. This is due to the prompt being repainted after executing:
-> reedline.handle_event(&prompt, move_to_line_start_event.clone())
-> reedline.handle_editor_event(prompt, event)
-> reedline.submit_buffer(prompt)
-> reedline.repaint(prompt)

I fixed this issue by adding the #[cfg(not(test))] attribute on top of the last reedline.repaint function call, so that it is not invoked when running unit tests.

I also decided to rewrite the restored unit tests using #[test] rather than #[rstest] because of https://github.com/nushell/reedline/pull/704#issuecomment-1888121873 talking about getting rid of rstest altogether. It seems like Reedline still makes use of rstest and as such I'm not sure if this rewrite was necessary.


fdncred:
I've fine with replacing rstest altogether here, if you think it'll help, but we just need to place some comments so that it doesn't come back in another PR explaining why it's been replaced and not to re-add it. I'm not sure myself if porting rstests to tests is required but I'm up for it. I'm also up for leaving them if you think that's right too.

Looks like a pretty simple fix. We just need to get the ci green.

Thanks for working on this!

ymcx:
> I've fine with replacing rstest altogether here, if you think it'll help, but we just need to place some comments so that it doesn't come back in another PR explaining why it's been replaced and not to re-add it. I'm not sure myself if porting rstests to tests is required but I'm up for it. I'm also up for leaving them if you think that's right too.

I'm fine with rstest, in fact, I'm not even sure what rstests are (lol). It's just that I noticed that there were no rstest attributes in the current version of the file and judging by the aforementioned comment regarding the refactor, I figured I'd just reintroduce the removed rstests as normal rust tests, since that's what I'm most comfortable with. All in all, I don't have a preference on this.

If I'm feeling like it, I'll probably refactor all the remaining rstests into normal tests and submit it as another PR. Sounds like they are preferred over rstest, but I'll probably need to do some quick research first on why that is.

> Looks like a pretty simple fix. We just need to get the ci green.

Whoops, didn't even notice they were failing. I did run both cargo fmt and cargo clippy before submitting, but seems like I'm missing some switches/arguments?

fdncred:
ok, sounds good.

ymcx:
The CI workflows should pass now. I had merged the main branch into this PR in commit https://github.com/nushell/reedline/pull/1078/changes/e7de4e8b5f39813046dd819dcfcdd01b51c81d4b using the GitHub web interface and didn't do a git pull on my local instance and as such there were no errors on my machine.

kronberger-droid:
Hey @ymcx,
I just merged #1098 which should solve the painter in test issue for now.
I would love you to use it in your PR.
If you want you can keep the tests with iterating over it or use rtests, your choice, I don't want you have even more work.
Thanks.

ymcx:
> I just merged #1098 which should solve the painter in test issue for now.

I haven't really reviewed your approach for fixing the stdout spam issue, but I'm sure it's better than my initial approach.

I did a hard reset back to the main branch, cherry-picked the restoration commit f6dc83e8a2e24433bee6f7b4dda989b6be6e0a6f, the history line completion unit test commit 996797359685d05896e2447151bb9268eb723f61 and merged all commits from upstream 1cec7fa390ee0f9524b5d2688d08e0978b113b9a. While at it, I rewrote the unit tests in the first commit as rstests and force-pushed all commits to make the history more clean.

The problem now is I'm unable to pass any of the three unit tests (the two added back in the first commit and the one I wrote in the second commit).

kronberger-droid:
Maybe i can take a look later and see whats going on.

ymcx:
I was able to fix the last unit test, currently I'm looking at fixing the rest of them and submitting the fixes here.

kronberger-droid:
Nice thanks!

PR #1087, opened 2026-05-28
feat(vi): add `u`, `U` visual mode commands

matystroia:
This PR adds the following commands to visual mode:
* `u` to convert selection to lowercase
* `U` to convert selection to uppercase
* `~` to switch case for each character in selection

These match default vim mappings and would be especially useful when working with environment variables.

### Added:
* `Command::Lowercase`
* `Command::Uppercase`
* `EditCommand::LowercaseSelection`
* `EditCommand::UppercaseSelection`
* `EditCommand::SwitchcaseSelection`

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -140,6 +140,9 @@ impl Editor {
             EditCommand::SelectAll => self.select_all(),
             EditCommand::CutSelection => self.cut_selection_to_cut_buffer(),
             EditCommand::CopySelection => self.copy_selection_to_cut_buffer(),
+            EditCommand::LowercaseSelection => self.lowercase_selection(),
+            EditCommand::UppercaseSelection => self.uppercase_selection(),
+            EditCommand::SwitchcaseSelection => self.switchcase_selection(),
             EditCommand::Paste => self.paste_cut_buffer(),
             EditCommand::CopyFromStart => self.copy_from_start(),
             EditCommand::CopyFromStartLinewise => self.copy_from_start_linewise(),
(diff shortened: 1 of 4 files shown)


kronberger-droid:
Thanks for this.
Parser logic and mode guards look right and `u`/`U`/`~` in visual mode go through cleanly, and `u` → Undo in normal mode is properly gated.

One pre-existing limitation this surfaces: `vi/mod.rs:86-89` consults `normal_keybindings` in both Normal and Visual modes.
Users with bare `u` rebound to Undo will see undo fire in Visual instead of Lowercase.
The parser's mode gates are correct, but the keybinding lookup runs first.
I will have to make a follow-up PR for adding `visual_keybindings` to `Vi` so this doesn't entangle.

Also what through me off while testing is that single grapheme selections don't work, which is a repeating issue for Vi Mode.

But all of these are not your problems to worry about. Fine with landing this now and thinking about the other stuff in follow-ups.

@fdncred are you fine with this?

fdncred:
Seems reasonable. It's up to you @kronberger-droid

matystroia:
> Also what through me off while testing is that single grapheme selections don't work, which is a repeating issue for Vi Mode.

I noticed this as well since `vU` is a common pattern for capitalization. Fortunately I think the fix for it is fairly simple! I've created a draft PR for it here: #1089

kronberger-droid:
Could you still add a couple tests:

- Editor-level coverage for the new EditCommands on a fresh buffer
- End-to-end `vU/u`, since this is the case that motivated #1089

Otherwise good to land once tests are in.

matystroia:
Sorry for the wait! I've added a few test cases for the `EditCommand`. I don't know if adding the `vU` test on this branch makes sense since it would fail. I think a better place for it would be #1089 where we could test `vd` which has the same problem (and where it would pass).

kronberger-droid:
@matystroia 
Okay great looks good. Have you thought about using Unicode?
I will probably make a follow-up where i create a resolver for such transform function and make it Unicode anyways so its fine to land like this in principle.
Keep the test out, its fine.

PR #1088, opened 2026-05-28
docs(examples): add selection styling example

kronberger-droid:
@fdncred Like we talked about.
I used Vi Mode since my notebook has broken arrow keys. Hope thats fine.
I also added some comments with other possible styling.

These would just have to be exposed to nushell, if we want it there.

### DIFF
--- a/examples/selection_styling.rs
+++ b/examples/selection_styling.rs
@@ -1,14 +1,14 @@
+#[allow(unused_imports)]
 use nu_ansi_term::{Color, Style};
 use reedline::{DefaultPrompt, Reedline, Signal, Vi};
 use std::io;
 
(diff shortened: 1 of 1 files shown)


fdncred:
yup, i tested it and it looks good to me. In my perfect world we'd have something like ghostty has the inverts the selection colors but that sounds tricky. I think this is good for now because at least it demonstrates how it's done. Thanks!

kronberger-droid:
@fdncred Have you tried `Style::new().reverse()`or do you mean something else?

fdncred:
oh, ya. I forgot about `reverse()`. We can leave this example as whatever you like but for nushell I'd like to do reverse I think.

PR #1089, opened 2026-05-28
fix(vi): enter visual mode through command

matystroia:
### Problem:
Entering visual mode does not actually select char/grapheme under cursor. Thus you have to do something like `vlh` to actually move selection to cursor. This happens because the parser eats the `v` input and sets mode manually instead of it going through the usual `Command` -> `EditCommand` pipeline.

### Solution:
- Add `EditCommand::SelectChar` that sets selection anchor to current cursor position
- Add `Command::EnterViVisual` that maps to `EditCommand::SelectChar`

Mode change still happens in parser, but now through the return value of `changes_mode`.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -138,6 +138,7 @@ impl Editor {
                 self.move_left_until_char(*c, true, true, *select)
             }
             EditCommand::SelectAll => self.select_all(),
+            EditCommand::SelectChar => self.select_char(),
             EditCommand::CutSelection => self.cut_selection_to_cut_buffer(),
             EditCommand::CopySelection => self.copy_selection_to_cut_buffer(),
             EditCommand::Paste => self.paste_cut_buffer(),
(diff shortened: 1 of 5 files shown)


kronberger-droid:
One behavior shift worth knowing about is that the old `v` path emitted `ReedlineEvent::Esc`, which closed any open menus. The new path skips that, could you preserve it by changing `Command::EnterViVisual` mapping:

```rust
Self::EnterViVisual => vec![
    ReedlineOption::Event(ReedlineEvent::Esc),
    ReedlineOption::Edit(EditCommand::SelectChar),
],
```
Also a couple of test before merge?

- Editor-level: EditCommand::SelectChar sets selection_anchor
- End-to-end: v then U on a fresh buffer capitalizes the grapheme under the cursor etc.

Also the `EditCommand::SelectChar` has a trailing comma.

Afterwards I am fine with landing it.

kronberger-droid:
Hey sorry for not letting you know up front. These kind of bugs will be solved in a follow up of #1100, which will hopefully coming soon.
Still thanks for pointing it out and giving a lead on how to solve this.

Issue #1010, opened 2026-01-15
Slow prompt display (~2s) with terminals that don't respond to cursor position queries


po-nuvai:
## Description

Some terminal emulators don't properly respond to CSI 6n (Device Status Report) cursor position queries, causing a ~2 second delay on every prompt display.

This was reported in nushell/nushell#17210 by a user experiencing 2-second delays when using Termius on Android to SSH into machines with nushell as the login shell.

## Root Cause

In `src/painting/painter.rs:174`, `initialize_prompt_position()` calls:

```rust
let prompt_selector = select_prompt_row(suspended_state, cursor::position()?);
```

This `cursor::position()` (from crossterm) sends a CSI 6n escape sequence and waits for the terminal to respond with the cursor position. If the terminal doesn't respond, crossterm times out after ~2 seconds.

Since `initialize_prompt_position()` is called at the start of every `read_line()`, this delay happens **on every prompt display**, not just startup.

## Affected Terminals

- **Termius on Android** - doesn't respond to CSI 6n queries (the reported case)
- Potentially other mobile/embedded terminal emulators

Note: Termius on Linux works fine, as does OpenSSH from Termux on the same Android device.

## Suggested Solutions

1. **Shorter timeout with fallback**: Reduce the timeout for `cursor::position()` and fall back to assuming position (0, 0) or tracking position internally when the query fails/times out

2. **Configuration option**: Add a way to disable cursor position queries entirely for problematic terminals (e.g., `Reedline::disable_cursor_position_query()` or similar)

3. **Environment variable**: Check for something like `REEDLINE_NO_CURSOR_QUERY=1` to skip the query

4. **Lazy/cached querying**: Only query cursor position when absolutely necessary, and cache the result

## Reproduction

1. Use Termius Android app to SSH into a machine with nushell as the login shell
2. Observe ~2 second delay after every command before the prompt appears

The delay shows up in nushell's perf logging as:
```
setup reedline took 2.001272815s
```

## References

- Original issue: nushell/nushell#17210
- Crossterm cursor position implementation uses a blocking read with timeout
- Similar issues exist in other shells/editors with terminals that don't respond to DSR queries

Issue #870, opened 2025-01-03
When Reedline may encounter "The cursor position could not be read within a normal duration"?


tisonkun:
I wonder when Reedline may encounter the internal error "The cursor position could not be read within a normal duration".

I can't stably reproduce it but here is the structure of my code:

```rust
    let mut state = Reedline::create()
        .with_validator(Box::new(ScopeQLValidator))
        .with_highlighter(Box::new(ScopeQLHighlighter));

    loop {
        let input = state
            .read_line(&prompt)
            .change_context_lazy(|| Error::Other("failed to read next line".to_string()))?;
        let input = match input {
            Signal::CtrlC | Signal::CtrlD => {
                println!("exit");
                break;
            }
            Signal::Success(input) => input,
        };
        let input = input.trim();

        // other biz code

        state.run_edit_commands(&[EditCommand::InsertString(
            outstanding.trim_start().to_string(),
        )]);
    }
```

And today in Ubuntu Terminal I encounter:

```
thread 'main' panicked at cmd/scopeql/src/[main.rs:50](http://main.rs:50/):78:
called `Result::unwrap()` on an `Err` value: failed to read next line
|- at cmd/scopeql/src/repl/entrypoint.rs:42:14
|
| -> The cursor position could not be read within a normal duration
    | - at cmd/scopeql/src/repl/entrypoint.rs:42:14
```

I have no idea when this error may occur and find no way to investigate deeper.

fdncred:
That looks to be a crossterm error message.
https://github.com/crossterm-rs/crossterm/blob/e104a7cb400910609cdde36f322e3905c4baa805/src/cursor/sys/unix.rs#L50

tisonkun:
@fdncred Yeah.

Do you have some information here to investigate more? Anyway, it's hidden behind reedline's API. I don't know how the error variant can be hit here.

fdncred:
Not really. It looks to me like crossterm is asking the terminal for the cursor position, waits 2 seconds without a response and then times out with that error.

tisonkun:
Well. I have no background how "asking for cursor position" can hang or fail :P

fdncred:
Me either. It may be worth asking on the crossterm repo. I just looks like if crossterm fails to get a lock for the event reader it would fail. I'm not sure what would cause that.

PR #1090, opened 2026-05-28
perf: minimize cursor::position queries

### LINKED ISSUE #968: reduce the `\e[6n` spam to improve high-ping ssh scenarios
currently reedline sends a `\e[6n` to the terminal after every redraw (pressing any key).
this causes reedline to block until it gets the cursor-position back from the terminal.  

lets say you have a 500ms ping (Geo-stationary satellite internet, cross-continent, etc):
1. you send a single keypress lets say `a`
1. 500ms: pass and it arrives at the server
1. 1sec: reedline requests the cursor position (`\e[6n`)
    1. 500ms your terminal receives the request
    1. 500ms reedline recieves the response
1. 500ms: reedline sends the updates to your terminal

so it takes 2 seconds to render 1 character change.. and half of it is caused by the cursor-position-request.

i found this by running nu via `^nu | ^tee test.bin` and checking `test.bin` after every key-press in a separate terminal.

## References

discussions about reedline being slow over ssh:
* https://github.com/nushell/reedline/issues/914
* <https://discord.com/channels/601130461678272522/614593951969574961/1431608904009388042>
* <https://discord.com/channels/601130461678272522/614593951969574961/1409160306256511030>

reubeno:
# Skip redundant cursor::position queries in the repaint hot path

Reedline issues a CSI 6n / DSR (`cursor::position()`) on every paint to check whether `prompt_start_row` has drifted from the terminal's real cursor row. That's a synchronous round-trip; over high-latency transports like SSH it dominates prompt-redraw latency.

Within a `read_line` cycle reedline is the sole writer to the terminal and updates `prompt_start_row` itself as its rendering scrolls — so the drift check is unnecessary in the steady state.

A new `prompt_start_row_stale: bool` on `Painter` tracks freshness:

- Cleared after the cursor query in `initialize_prompt_position`.
- Set by paths that yield the tty or write content the painter doesn't  model (`handle_resize`, `print_external_message`, `paint_line`,  `open_editor`).
- Re-cleared on the next paint by a confirmed-no-drift query.

The hot-path drift check skips the DSR when the cache is fresh. The change is wholly inside reedline, so the improvement carries to any consumer (e.g., Nushell, brush, custom REPLs) without changes on their side.

## Notes

I know that this is an area that may be challenging to test/validate. I'm open to any suggestions/feedback on how to ensure that this change does not regress existing functionality, while still improving the performance.

**_Full disclosure:_** as you can probably tell, I used Claude to identify and develop changes -- but I was heavily steering, and ultimately rewrote the code several times to address concerns that I had myself over maintainability challenges. I'm also placing this in draft first, with the hope to get early feedback on the approach and the maintainers' openness to considering this change. Any and all feedback is very much welcome, please :smile:

## Tradeoffs

- Each invalidating event trades the unconditional per-paint DSR for one DSR on the next paint; that's strictly better than the current per-paint cost, but not free on workloads heavy in external-printer batches or resizes.
- Terminals where `cursor::position()` fails see the same behavior as before: the drift check still runs and fails, with no optimization gain (and no regression).

## Measurements

Nushell 0.113.0 (HEAD) built against this branch's reedline vs against reedline `main`. PTY harness (portable-pty), 80×24, scripted 80 ms input pacing, 5 Enter presses per run, `--no-config-file --no-history`. Three samples per condition; the harness is deterministic and produced identical DSR counts across samples within each condition.

DSR counts:

| reedline | total DSRs (5 Enters) | per Enter |
| --- | --- | --- |
| `main` | 19 | 3.80 |
| this branch | 6 | 1.20 |

Same harness with `--simulated-rtt-ms 125` (each CSI 6n reply delayed 125 ms — comparable to a typical cross-region SSH round trip):

| reedline | total wall-clock (5 Enters) | per Enter |
| --- | --- | --- |
| `main` | ~2150 ms | ~430 ms |
| this branch | ~640 ms | ~127 ms |

The remaining ~1 DSR/cycle on this branch is the unavoidable one in `initialize_prompt_position`; reedline genuinely doesn't know where the host left the cursor before `read_line` is called.

[brush](https://github.com/reubeno/brush), a smaller `reedline`-based shell, shows the same shape (3.0 → 1.0 DSRs/cycle) — confirming the win is at the reedline layer, not shell-specific.

closes #968 

related #1010 
related #870

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -165,7 +165,15 @@ pub(crate) struct PromptLayout {
 pub struct Painter {
     // Stdout
     stdout: W,
+    /// Row where the prompt starts on screen. `_stale = false` means
+    /// the cached value is consistent with the terminal as of the
+    /// last successful query or paint; `true` means something may
+    /// have moved the cursor or written content we don't model, so
+    /// the next `repaint_buffer` must re-query. Call
+    /// `invalidate_prompt_start_row` from any new code path that
+    /// violates this invariant.
     prompt_start_row: u16,
+    prompt_start_row_stale: bool,
     // The number of lines that the prompt takes up
     prompt_height: u16,
     terminal_size: (u16, u16),
@@ -184,6 +192,9 @@ impl Painter {
         Painter {
             stdout,
             prompt_start_row: 0,
+            // `true` so `repaint_buffer` cannot trust an uninitialised
+            // row; `initialize_prompt_position` clears it.
+            prompt_start_row_stale: true,
             prompt_height: 0,
             terminal_size: (0, 0),
             last_required_lines: 0,
@@ -373,7 +384,18 @@ impl Painter {
                 }
             }
         };
+        // `prompt_start_row` matches the cursor row we just measured;
+        // mark fresh so subsequent paints can skip the drift-detection DSR.
+        self.prompt_start_row_stale = false;
         Ok(())
+    }
+
+    /// Mark `prompt_start_row` as possibly out of sync — the next
+    /// `repaint_buffer` will re-query the terminal. Call from any path
+    /// that lets something other than reedline's rendering move the
+    /// cursor (e.g. `$EDITOR`).
+    pub(crate) fn invalidate_prompt_start_row(&mut self) {
+        self.prompt_start_row_stale = true;
     }
 
     /// Main painter for the prompt and buffer
@@ -414,6 +436,9 @@ impl Painter {
                 .prompt_start_row
                 .saturating_sub(lines_before_cursor - 1);
             self.just_resized = false;
+            // Re-anchor complete; cache is back in sync with the
+            // prompt's post-resize screen origin.
+            self.prompt_start_row_stale = false;
         }
 
         // Lines and distance parameters
@@ -423,21 +448,36 @@ impl Painter {
         // Marking the painter state as larger buffer to avoid animations
         self.large_buffer = required_lines >= screen_height;
 
-        // This might not be terribly performant. Testing it out
-        let is_reset = || match cursor::position() {
-            // when output something without newline, the cursor position is at current line.
-            // but the prompt_start_row is next line.
-            // in this case we don't want to reset, need to `add 1` to handle for such case.
-            Ok(position) => position.1 + 1 < self.prompt_start_row,
-            Err(_) => false,
+        // True if the prompt has scrolled above the cached
+        // `prompt_start_row` and the caller must re-anchor at row 0.
+        // When stale, query the terminal; clear the stale flag if the
+        // query confirms no drift, so later paints can skip the query.
+        let should_reset_anchor = if self.prompt_start_row_stale {
+            match cursor::position() {
+                // The `+1` handles the case where the previous output
+                // ended without a newline, leaving the cursor on the
+                // same row as the next prompt.
+                Ok(position) => {
+                    let drifted = position.1 + 1 < self.prompt_start_row;
+                    if !drifted {
+                        self.prompt_start_row_stale = false;
+                    }
+                    drifted
+                }
+                Err(_) => false,
+            }
+        } else {
+            false
         };
 
         // Moving the start position of the cursor based on the size of the required lines
-        if self.large_buffer || is_reset() {
+        if self.large_buffer || should_reset_anchor {
             for _ in 0..screen_height.saturating_sub(lines_before_cursor) {
                 self.stdout.queue(Print(&coerce_crlf("\n")))?;
             }
             self.prompt_start_row = 0;
+            // The reset puts the prompt at row 0; cache is back in sync.
+            self.prompt_start_row_stale = false;
         } else if required_lines >= remaining_lines {
             let extra = required_lines.saturating_sub(remaining_lines);
             self.queue_universal_scroll(extra)?;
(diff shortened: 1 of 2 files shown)


src/painting/painter.rs
```diff
@@ -165,7 +165,15 @@ pub(crate) struct PromptLayout {
...
     stdout: W,
+    /// Row where the prompt starts on screen. `_stale = false` means
+    /// the cached value is consistent with the terminal as of the
+    /// last successful query or paint; `true` means something may
+    /// have moved the cursor or written content we don't model, so
+    /// the next `repaint_buffer` must re-query. Call
+    /// `invalidate_prompt_start_row` from any new code path that
+    /// violates this invariant.
     prompt_start_row: u16,
+    prompt_start_row_stale: bool,
```
Bahex:
```suggestion
    prompt_start_row: Option<u16>,
```
Using `Option` instead of two separate values that are semantically linked would be better. Having `None` instead of checking a `_stale` field ensures it's not possible to accidentally use a stale value.

edit: Looks like there is at least one place where the stale value can be useful. In that case a custom `enum` would be better. Maybe something like:
```rust
enum CachedValue<T> {
    Clear,
    Stale(T),
    Fresh(T),
}
```

reubeno: Thanks for the feedback, @Bahex. I'd considered an enum but wanted to keep the change simpler; since you're also thinking it would be an improvement, I've pushed changes to combine the fields.

I went with a custom enum so I could have some convenience helpers implemented on it, but I'm also open to a generic `CachedValue<T>` like you suggest above if that's what you'd recommend/prefer.

If you have time to look at the updated version, please let me know what you think.

bhouse-nexthop:
Drive-by from a downstream consumer that runs reedline over long-haul SSH — we hit exactly this and independently traced the inter-prompt lag to the per-paint `cursor::position()`, so a strong +1 on the approach. This also resolves #968 (and helps #1010/#870); might be worth a `Closes #968` to link it.

Two things from reading the diff:

1. **`clear_screen` / `clear_scrollback` look unhandled.** Both `MoveTo(0, 0)` but don't update `prompt_start_row`, and the engine's `ClearScreen` / `ClearScrollback` arms (`engine.rs` ~872 / ~1107) don't invalidate. Today the next paint's DSR re-anchors; once that DSR is skipped on a `Verified` cache, wouldn't `Ctrl-L` re-anchor at the stale pre-clear row? Since the cursor is *known* to be at row 0 after a clear, `mark_verified(0)` there would be both correct and free (no DSR). A `paint → Ctrl-L → paint` test would pin it.
2. **The residual `initialize_prompt_position` DSR** you call out: for consumers that fully own their output and always hand control back at a known position (e.g. column 0 after a trailing newline), it'd be great to let the host *assert* the start position and skip even that one — an opt-in like `with_known_start_row(u16)` / a "host guarantees col 0" hint. Would you be open to a follow-up on top of this PR?

Thanks for doing this — the measurements are great.

kpcyrd:
Thanks for working on this!

Is there any way we can avoid cursor::position in entirety? No other shell seems to be doing this (including the rustyline crate), they instead calculate the cursor position (because they themselves are the ones moving the cursor when printing the prompt).

I'm concerned that even one call might be too much ("if Rust is so fast and efficient, why does this feel slower than python").

fdncred:
@kpcyrd Let's take one step at a time. I don't think reedline is able to manage all cursor positions at this time. I'd like to see reedline get there but let's take baby steps first.

@bhouse-nexthop Thanks for the input. I'd like to see your point 1 answered in this PR and 2 would work for a follow-up PR.

@reubeno Thanks for this cool PR. What are your thoughts on the comments? I'd like to move on with it soon if we can, so we have time to test it out.

reubeno:
Thanks all for the replies!

I'm in agreement with @fdncred and @kpcyrd that it would be nice to get to a place where we can remove all or most `cursor::position()` calls *and* I'm on the same page with @fdncred that it's best to take things one step at a time.

Regarding @bhouse-nexthop's comments:

>clear_screen / clear_scrollback look unhandled. Both MoveTo(0, 0) but don't update prompt_start_row, and the engine's ClearScreen / ClearScrollback arms (engine.rs ~872 / ~1107) don't invalidate

I believe the existing PR logic is correct. Here's what I see...

The `ReedlineEvent::ClearScreen` and `ReedlineEvent::ClearScrollback` arms in `engine.rs` call into `Painter::clear_screen()` and `Painter::clear_scrollback()`, respectively. Each of those functions do issue a `.MoveTo(0, 0)`, and they both subsequently call `initialize_prompt_position()`, which in turn unconditionally calls `cursor::position()` and then updates `prompt_start_row`.

Because of this, I don't see a correctness issue there, but @bhouse-nexthop makes a good point that this could be optimized. With the call to `MoveTo(0, 0)` we know exactly where the cursor should be and could skip the cursor position check. With that said, this sounds like a change that could be made as a follow-on incremental improvement (i.e., separate PR) to keep things moving forward. That would be my preference, but I'm happy to incorporate it too if the reviewers/maintainers would prefer otherwise.

>for consumers that fully own their output and always hand control back at a known position (e.g. column 0 after a trailing newline), it'd be great to let the host assert the start position...

Agreed with everyone here that this is an interesting possibility and best suited as a separate PR after this one.

fdncred:
We have about 3 weeks to test this and ensure it works as expected. So, let's try it. Thanks!

reubeno:
> We have about 3 weeks to test this and ensure it works as expected. So, let's try it. Thanks!

Thanks, @fdncred! Please do holler at me if issues get reported related to this.

For what it's worth, I also spent time last night having Claude build out ~100 pty-based end-to-end UX tests against a test reedline app, taking inspiration from `neovim`'s tests. Thus far, the functional behavior they've exercised has been consistent before and after this change--but I'd like to add some more edge cases. (And of course, that's still no replacement for getting human eyes and bake time on this.)

If there's interest in tests like those, that's something I'd also be happy to share.

fdncred:
ya, i've been looking for some way to test with a ptty. that would be very cool!

reubeno:
> ya, i've been looking for some way to test with a ptty. that would be very cool!

@fdncred In case you're interested, [here's what claude put together for testing](https://github.com/reubeno/reedline/tree/pty-e2e-tests/tests/e2e). I'm not comfortable submitting a draft because I haven't actually reviewed the changes myself (yet).

One of the things I was most interested in is seeing if some of the active `reedline` issues could be reproduced with test cases; here was the result: https://github.com/reubeno/reedline/blob/pty-e2e-tests/tests/e2e/issue_repros.rs. An example:

image

Happy to chat about this some if interesting...

fdncred:
ya, definitely very cool. I can't play with it in detail this second, but I want to look at it closer. 

I see it doesn't work on windows yet, that's ok for a first pass but we'd need to support it for windows specific issues like exists right now with wezterm. that issue is related to OSC133. so i'd like to be able to capture/test for particular ansi escape sequences that are supposed to be emitted by reedline. other than that, it's a definite YES! I'd love to see an official version of this landed at some point in the near future.

PR #1091, opened 2026-05-29
test(vi): expand parse_event coverage as refactor safety net

kronberger-droid:
## Summary

- Adds 25 tests for `Vi::parse_event` (mode entry/exit, char dispatch, multipliers, invalid-motion, `dd`, `.` repeat, non-Key events).
- Adds a `key()` helper to cut event construction boilerplate.
- Safety net for an upcoming `parse_event` refactor.

## Note

Two `.` repeat tests pin a wrap-accumulator bug (`previous` grows one `Multiple` layer per press). Assertions match current behaviour; fix queued as follow-up.

## Test plan

- [x] `cargo test --lib` — 880 passed

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -230,6 +230,10 @@ mod test {
     use super::*;
     use pretty_assertions::assert_eq;
 
+    fn key(code: KeyCode, modifiers: KeyModifiers) -> ReedlineRawEvent {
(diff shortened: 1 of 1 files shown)


fdncred:
nice work!

kronberger-droid:
Thanks, I was looking over `parse_event` and was more than lost.
So I consider refactoring it to be more readable.
But was not confortable doing it without better test coverage.

fdncred:
sounds like a wise decision

PR #1092, opened 2026-06-02
refactor(completer)!: don't require Send on the Completer trait

kronberger-droid:
The `unsafe impl Send` on `HistoryCompleter` was actually sound: it's a `pub(crate)` type only ever created as a stack local and borrowed into menus, never sent across threads.
But it pointed at a trait that's too tight, since `Completer: Send` forces *every* completer to be `Send`.

So instead of asserting `Send` where it isn't required, this drops it from the `Completer` trait and requires it only on the completers that are actually owned and stored, thus in `Reedline` and `ReedlineMenu::WithCompleter`, now `Box<dyn Completer + Send>`.
The `unsafe` falls away naturally, with **no** new bound on `History`, so `SqliteBackedHistory` (whose `rusqlite::Connection` is `!Sync`) is untouched.

This is an alternative to #869, which tightened `History: Sync` to fix the same `unsafe` but broke the sqlite backend as a result.

Added `reedline_is_send` as a compile-time guard on the relocated bound.

**API note:** `with_completer` and `ReedlineMenu::WithCompleter` now take `Box<dyn Completer + Send>`. Source-compatible (the old `Completer: Send` already required it), but visible in signatures.

### DIFF
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -29,7 +29,7 @@ impl Span {
 
 /// A trait that defines how to convert some text and a position to a list of potential completions in that position.
 /// The text could be a part of the whole line, and the position is the index of the end of the text in the original line.
-pub trait Completer: Send {
+pub trait Completer {
     /// the action that will take the line and position and convert it to a vector of completions, which include the
     /// span to replace and the contents of that replacement
     fn complete(&mut self, line: &str, pos: usize) -> Vec<Suggestion>;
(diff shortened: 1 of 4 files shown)


sholderbach (approved):
Thanks for picking this up, if this compiles with our completers this looks good!

kronberger-droid:
@sholderbach
Sure thing.
Just tested it: current nushell main builds clean with reedline#1092.

PR #1093, opened 2026-06-03
non-blocking completions in reedline

philocalyst:
related to the [respective nushell change](https://github.com/nushell/reedline/commit/971fefe852f4a31ae60463c3ae9e338b2203f99a).

Adds primitives for non-blocking completions.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -852,6 +852,22 @@ impl Reedline {
                 }
             }
 
+            // When a background completion finishes, re-populate the active
+            // menu so results appear without waiting for another keypress.
+            if self.completer.check_pending() {
+                let has_active = self.menus.iter().any(|m| m.is_active());
+                if has_active {
+                    if let Some(menu) = self.menus.iter_mut().find(|m| m.is_active()) {
+                        menu.update_values(
+                            &mut self.editor,
+                            self.completer.as_mut(),
+                            self.history.as_ref(),
+                        );
+                    }
+                    self.repaint(prompt)?;
+                }
+            }
+
             // Helper function that returns true if the input is complete and
             // can be sent to the hosting application.
             fn completed(events: &[Event]) -> bool {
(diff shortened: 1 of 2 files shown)


src/engine.rs
kronberger-droid, line 1: Any reason for the double if let? I think this can be collapsed to one:
```rust
if self.completer.check_pending() {
    if let Some(menu) = self.menus.iter_mut().find(|m| m.is_active()) {
        menu.update_values(
            &mut self.editor,
            self.completer.as_mut(),
            self.history.as_ref(),
        );
    }
    self.repaint(prompt)?;
}
```

src/engine.rs
```diff
@@ -852,6 +852,22 @@ impl Reedline {
                 }
             }
 
+            // When a background completion finishes, re-populate the active
+            // menu so results appear without waiting for another keypress.
+            if self.completer.check_pending() {
```
kronberger-droid: `check_pending` is polled unconditionally on every loop iteration here, but the doc comment on the trait method says "The engine calls this while a menu is active and `has_pending` has returned true.
Can we either gate the call or update the doc to describe the actual unconditional polling?

src/engine.rs
philocalyst: Totally can! Found it more readable in combination with the method chain.

src/engine.rs
philocalyst: Ah! Wrote the comment first. On it.

src/engine.rs
philocalyst: Nevermind, we're not updated to 2024 edition.

src/engine.rs
```diff
@@ -852,6 +852,19 @@ impl Reedline {
...
```
kronberger-droid: You could move the
```rust
let completer_pending = self.completer.has_pending();
```
up here and then gate it with:
```rust
if completer_pending && self.completer.check_pending() {...}
```
and reuse it in `needs_polling`.

This was the gating I was thinking of.
For both to default to false its not a deal breaker but it is cleaner this way.

philocalyst: Should be good now, but let me know if I misunderstood you?

philocalyst:
@kronberger-droid Should be all resolved :)

PR #1094, opened 2026-06-03
refactor: abbrevation expansion trait logic

casedami:
as discussed in https://github.com/issues/mentioned?issue=nushell%7Cnushell%7C18315 

renames `is_inside_string_literal` to `should_expand_abbr` and adds an `AbbrExpandContext` parameter so that the client can define AST expansion logic according to the expansion context (ie 'bang expansion' or 'fish-style expansion')

i would say that the existing abbreviation tests are sufficient on the reedline side of things, especially since the main behavior change will come from nushell

### DIFF
--- a/src/highlighter/mod.rs
+++ b/src/highlighter/mod.rs
@@ -5,6 +5,19 @@ use crate::StyledText;
 
 pub use example::ExampleHighlighter;
 pub use simple_match::SimpleMatchHighlighter;
+
+/// The context in which abbreviation expansion is being attempted
+///
+/// Passed to [`Highlighter::should_expand_abbr`] so implementations can apply
+/// different veto rules depending on which expansion triggered the check
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AbbrExpandContext {
+    /// Fish-style word abbreviation
+    WordAbbreviation,
+    /// Bashism history expansion
+    BangExpansion,
+}
+
 /// The syntax highlighting trait. Implementers of this trait will take in the current string and then
 /// return a `StyledText` object, which represents the contents of the original line as styled strings
 pub trait Highlighter: Send {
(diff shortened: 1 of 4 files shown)


src/highlighter/mod.rs
```diff
@@ -5,6 +5,19 @@ use crate::StyledText;
...
+/// The context in which abbreviation expansion is being attempted
+///
+/// Passed to [`Highlighter::should_expand_abbr`] so implementations can apply
+/// different veto rules depending on which expansion triggered the check
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AbbrExpandContext {
+    /// Fish-style word abbreviation
+    WordAbbreviation,
+    /// Bashism history expansion
+    BangExpansion,
```
fdncred, line 19: I believe these are gated by the `bashisms` feature. If you search the repo, you'll see it. So, I think that means these bashisms should be feature gated here to and maybe other places in this PR. Thoughts?

casedami: good catch, i've added that now. i believe that should be the only place

kronberger-droid:
Any blockers for this to merge? Other than a breaking change note (!).
@fdncred @casedami

fdncred:
> Any blockers for this to merge? Other than a breaking change note (!).

No. My only question is if this fixes the nushell issue on it's own or if there is code needed on the nushell side to fix it there.

casedami:
It needs code on the nushell side to fix, I have a PR ready for that I just haven't submitted it yet.

fdncred:
ok, thanks. reedline commits are building up so hopefully patching nushell to the latest reedline commit doesn't break anything else.

kronberger-droid:
> ok, thanks. reedline commits are building up so hopefully patching nushell to the latest reedline commit doesn't break anything else.

I think this one is the only one that could really break anything. Everything else is just exposing things in nushell`s config or is non-breaking.
I will try to keep the markdown note updated.

fdncred:
@kronberger-droid Great! @casedami Ping me when you have your nushell PR up. Thanks for all the help!

PR #1096, opened 2026-06-04
Added new functionality to match with files in cwd

### LINKED ISSUE #5799: Current working directory complete in addition to history complete
### Related problem

I have read the Reedline documentation but either the feature isn't there or how to accomplish it isn't clicking in my head.

In Fish, the default configuration suggests the contents of the current working directory in addition to history.

Is this an existing option and I'm just wooshing on it hard? If it is, can it be made default?

### Describe the solution you'd like

Use both history and the CWD to make autocomplete suggestions

### Describe alternatives you've considered

_No response_

image

https://asciinema.org/a/8g9NWwULdsus6xyIjF5ztmMMs

Avyukt27:
### Description

Updated `CwdAwareHinter` to add a hinting from cwd 

If the history searching does not provide any auto-suggestion, the hinter will search through the cwd for matches instead of returning nothing.

### Related Issues
Fixes nushell/nushell#5799

### DIFF
--- a/src/hinter/cwd_aware.rs
+++ b/src/hinter/cwd_aware.rs
@@ -49,20 +49,44 @@ impl Hinter for CwdAwareHinter {
                     .unwrap_or_default()
                     .to_string()
             } else {
-                history
-                    .search(SearchQuery::last_with_prefix(
-                        line.to_string(),
-                        history.session(),
-                    ))
-                    .unwrap_or_default()
-                    .first()
-                    .map_or_else(String::new, |entry| {
-                        entry
-                            .command_line
-                            .get(line.len()..)
+                if let Some(last_word) = line.split_whitespace().last() {
+                    let mut cwd_match = String::new();
+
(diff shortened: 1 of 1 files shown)


fdncred:
I'm not convinced that this is a good idea. CwdAwareHinter is about the history, not the file system.

Avyukt27:
Should I move this to another hinter type then?

blindFS:
Correct me if I'm wrong. I think fish adds completion results as hints and has a smart sorting algorithm so that the hints change between historical commands and completions as you type.

Basically hints for files of CWD would be a small subset of such functionality. Probably better to make it more general IMHO.

kronberger-droid:
Closing this one since there was no follow-up since June and i agree with @fdncred.
On the nushell side this is already doable through `$env.config.hinter.closure`.

A completer-backed hint like @blindFS suggested would be a separate design discussion, feel free to open an issue for that.

@Avyukt27 Thanks for the effort anyways.

PR #1097, opened 2026-06-07
fix(painting): preserve leading CRLF in no-color rendering

### LINKED ISSUE #18360: `input --reedline` breaks when navigating multi lines
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

When using `input --reedline` with multi-line input, and navigating that input e.g. with arrow keys, this mangles the input lines instead of behaving correctly like the normal nu shell reedline.

### How to reproduce

1. enter `input --reedline`
2. type some random text into multiple lines using alt+enter, e.g. `1 [alt+enter]`, `2 [alt+enter]`, `3 [alt+enter]`, `4 [alt+enter]`
3. use the arrow keys to navigate the input, e.g. move cursor right, left, up, down
4. the original input gets dislocated, teleported around the cursor, etc.

### Expected behavior

`input --reedline` should behave like the outer shell's reedline

### Configuration

| key                  | value                                                                                                                              |
| -------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
| version              | 0.113.1                                                                                                                            |
| major                | 0                                                                                                                                  |
| minor                | 113                                                                                                                                |
| patch                | 1                                                                                                                                  |
| branch               |                                                                                                                                    |
| commit_hash          |                                                                                                                                    |
| build_os             | linux-x86_64                                                                                                                       |
| build_target         | x86_64-unknown-linux-gnu                                                                                                           |
| rust_version         | rustc 1.95.0 (59807616e 2026-04-14) (built from a source tarball)                                                                  |
| cargo_version        | cargo 1.95.0 (f2d3ce0bd 2026-03-21)                                                                                                |
| build_time           | 1980-01-01 00:00:00 +00:00                                                                                                         |
| build_rust_channel   | release                                                                                                                            |
| allocator            | standard                                                                                                                           |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                                                   |
| installed_plugins    |                                                                                                                                    |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false, native-clip=false, cell-path-types=false |

Alb-O:
Avoid `Cow<str> += ...` while normalizing line endings because the empty-LHS optimization can replace a preallocated buffer. A leading newline in no-color after-cursor text could otherwise lose its CRLF replacement before later newlines were processed.

Fixes https://github.com/nushell/nushell/issues/18360

### DIFF
--- a/src/painting/utils.rs
+++ b/src/painting/utils.rs
@@ -10,24 +10,27 @@ pub(crate) fn coerce_crlf(input: &str) -> Cow<'_, str> {
     let mut cursor: usize = 0;
     for (idx, _) in input.match_indices('\n') {
         if !(idx > 0 && input.as_bytes()[idx - 1] == b'\r') {
-            if let Cow::Borrowed(_) = result {
-                // Best case 1 allocation, worst case 2 allocations
-                let mut owned = String::with_capacity(input.len() + 1);
-                // Optimization to avoid the `AddAssign for Cow<str>`
-                // optimization for `Cow<str>.is_empty` that would replace the
-                // preallocation
-                owned.push_str(&input[cursor..idx]);
-                result = Cow::Owned(owned);
-            } else {
-                result += &input[cursor..idx];
+            match &mut result {
+                Cow::Borrowed(_) => {
+                    // Best case 1 allocation, worst case 2 allocations.
+                    // Avoid `AddAssign for Cow<str>` because its empty-LHS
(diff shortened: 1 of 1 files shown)


kronberger-droid:
Looks good, thanks!

PR #1098, opened 2026-06-09
tests(painter): wrapper enum to allow tests to write to sink

kronberger-droid:
This is in rensponse to #1078, which had to use a workaround to not leak test output into the terminal.

I solved this using a wrapper enum:
```rust
pub enum W {
    #[cfg_attr(test, allow(dead_code))]
    Terminal(std::io::BufWriter<std::io::Stderr>),
    #[cfg(test)]
    Sink(std::io::Sink),
}
```
Allowing to either paint to the Terminal or to a Sink, which will be used in tests.

@fdncred @ymcx
What do you think about using this instead of the workaround in #1078?

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -27,7 +27,7 @@ use {
             FileBackedHistory, History, HistoryCursor, HistoryItem, HistoryItemId,
             HistoryNavigationQuery, HistorySessionId, SearchDirection, SearchQuery,
         },
-        painting::{Painter, PainterSuspendedState, PromptLines, RenderSnapshot},
+        painting::{Painter, PainterSuspendedState, PromptLines, RenderSnapshot, W},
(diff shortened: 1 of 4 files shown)


fdncred:
I think it's probably ok for now but in the future I'd like to have a way of running tests and capturing ansi escapes to ensure that they're correct. It seems like this would prevent that right now.

kronberger-droid:
Hmm, my idea for that would be to add another field with Buffer (`Vec<u8>`), which allows you to capture the Bytes without writing it to anything.

@fdncred
What do you think about this?
It would still not fully solve the problem since checking the raw bytes is still unreliable, but there apparently are crates that can transform it into the escape codes.

fdncred:
I'm ok with this and also adding a Buffer later if we get to it.

PR #1099, opened 2026-06-10
feat(history): rank history menu by existing prompt

Alb-O:
History menu completion now considers text already present before the active search input when ordering results.

I use the history search menu a lot in nushell, and this specific behavior has always irked me. So I'm proposing this solution, which in my opinion, is a big usability win for a likely common case where the user has partially entered a command into the line buffer, (with no history menu open), then hits Ctrl+R *after*, and expects the menu to complete their command based on their previous history of using that command.

Originally, the results yielded would not consider the existing prefix at all. After this PR change, we give high priority to results that match against the prefix. If a prefix-matching result gets selected, the existing line gets replaced by the overlapping result instead of appended, to match expected behavior for line completion. If there's no match for the existing prefix, or the current buffer is excessively long/multilined, we fallback to the original results and append selections to the line as we did before.

### DIFF
--- a/src/completion/history.rs
+++ b/src/completion/history.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 
 const SELECTION_CHAR: char = '!';
+const MAX_SEARCH_PROMPT_BYTES: usize = 512;
 
 // The HistoryCompleter is created just before updating the menu
 // It pulls data from the object that contains access to the History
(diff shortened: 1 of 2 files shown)


fdncred:
If I change my history_menu to use `only_buffer_difference: false` like this:
```nushell
 {
    name: history_menu
    only_buffer_difference: false
    marker: $"(char -u '1f50d') "
    type: {
      layout: list
      page_size: 10
    }
    style: {
      text: "#ffb000"
      selected_text: { fg: "#ffb000" attr: r }
      description_text: yellow
    }
  }
```
it searches already
Screenshot 2026-06-10 at 06-31-18

Or am I misunderstanding?

Alb-O:
Yep that's what I recreated, embarrassingly I did not even know these options existed, thank you.

fdncred:
The documentation talks about it here https://www.nushell.sh/book/line_editor.html#user-defined-menus

Issue #745, opened 2024-02-05
Give vi_normal mode's cut/paste access to the system clipboard


mister-drgn:
Is there any interest in allowing vi_normal mode to the use the system clipboard, rather than its own isolated clipboard? I realize this might require third-party support from something like xclip, but it could be a nushell configuration option, similar to how it works in neovim.

This feature would likely be more useful if a y->yank keybinding were added to vi_normal mode as well.

I'm pretty new here, so apologies if these requests have been addressed in the past. I couldn't find anything by searching. Thanks.

fdncred:
I'm guessing that if it doesn't already, it wouldn't be too tricky to add since reedline already uses the `aboard` crate.

https://github.com/nushell/reedline/blob/f2447364b923eaab55f6de4f73c1f79511c6b5eb/Cargo.toml#L16

sholderbach:
To clarify: which operations in your mind should touch the system clipboard?

- Should every operation (e.g. `dd`, `x`) reach into the clipboard
- Just a subset (`y...` as explicit copy, how would `p` work then?)
- We implement a vim register system as well and you should use `"+<operation touching the cb>` like `"+yy`

I think as a ton of basic editing ops (deletions) move content into vim internal "clipboards", sharing this with the system clipboard may be quite surprising or annoying depending on the situation.

A-Lovett:
This seems like a fair point. For anyone currently using vim mode, it would be disruptive if various editing operations suddenly started interacting with the system clipboard.

On vim, there's a single line in the configuration that can be used to specify that you'd like it to use the system clipboard. Something similar could make sense for reedline/nushell, with the option off by default.

musjj:
`set clipboard^=unnamedplus` is a very uncontroversial and common configuration used by vim users, which automatically shares the plus register with the system clipboard. I don't really see how it'd be surprising, as long as it's implemented correctly.

kronberger-droid:
closing in favor of #1221 if i missed something feel free to leave a comment.

PR #1100, opened 2026-06-11
refactor(editor): unify motions on a verb + target vocabulary

### LINKED ISSUE #1066: vi: cursor position semantics are not mode-aware (root cause of multiple off-by-one bugs)

kronberger-droid:
Hey all,
finally I have my editor rewrite in a state where I can present it to you.

## What This Is

Today every vi motion lowers to its own dedicated `EditCommand`, each with its own hand-rolled index math in `LineBuffer`/`Editor`.
Adding a motion means touching the parser, a new enum variant, a new editor method, and new boundary math.

This PR replaces that with one pipeline:

- **`MotionTarget`**: motions as *data* (word edge, line edge, buffer edge, etc.), shared by every mode.
- **`resolve_motion`**: the single resolver - target -> cursor head + operator end.
- **Six verbs on `EditCommand`**: `Move`/`Extend`/`Cut`/`Copy`/`Change`/`Erase` over a target, with `Granularity` (char-wise vs line-wise) deciding how the span is taken and how the register pastes. All operators funnel through one `operate()` sink.
- **`RestPolicy`**: where the cursor may rest is a mode property enforced at one commit boundary, not per-command clamps. This fixes the #1069 class of bugs structurally.

The vi key tables become thin: a key maps to a verb + target, and the old dedicated commands remain as documented legacy API.

**User-visible wins:** `dj`/`dk`/`yj`/`yk`/`cj`/`ck` work now, `cc`/`S` are vim-faithful, `$` is CRLF-consistent with the rest of the editor, and the vi-normal cursor invariant holds across history navigation and mode switches.

## Size Breakdown (+3 708 / -869)

| Area | prod LOC | test LOC | What |
|---|---:|---:|---|
| New core modules (`cursor`, `resolve`, `rest_policy`, `word`, `line`, `graphemes`) | +622 | +987 | cursor-as-range, the resolver, rest policies, word classifier, line math |
| `editor.rs` | +236 / −92 | +984 | verb dispatch, `operate()` sink, selection-inclusivity capture. tests are mostly for pinning old-command equivalence |
| `enums.rs` | +204 | — | the vocabulary + six verb variants + legacy-API doc notes |
| `line_buffer.rs` | +88 / −26 | +93 | cursor view (`cursor()`/`set_cursor()`), pure accessors, delegation to `line.rs` |
| vi lowering (`command`/`motion`/`mod`) | +303 / −301 | +135 / −69 | near net-zero: tables swap dedicated commands for verb+target |
| dead code removed (`helix/range.rs`) | −106 | −228 | unwired scaffolding superseded by `Cursor` |
| plumbing (`clip_buffer`, `prompt`, `engine`, `lib`) | +50 | — | `ClipboardMode`→`Granularity`, rest-policy mapping, mode sync |

So roughly: **60% of the diff is tests**, and the production half splits into ~2/3 new core modules and ~1/3 rewiring.

## Behavior Changes

Everything else is pinned as byte-identical by tests

- `cj`/`ck` now change current+adjacent line (were no-ops); `dj`/`dk`/`yj`/`yk` added.
- `cc`/`S`: register is line-wise like vim; `S` leaves a blank line instead of deleting the line.
- `$`/`d$` stop before the `\r` of a CRLF terminator (consistent with `move_to_line_end`).
- Vi-normal cursor never rests past the last grapheme, including after history navigation and Esc (#1069).
- Public API: adds `Granularity`, `MotionTarget`, etc. and the six verb variants. Removes only the crate-private `ClipboardMode`. All old `EditCommand` variants keep working.

## Fixes

Closes #1066 (mode-aware cursor semantics, via `RestPolicy` at a single commit
boundary), #893 (shift-select grabbing an extra grapheme), #563 and #667 (vi word
motions treating `abc.def` as one word).
Fixes the #1066 cascade in #694 (cursor resting past the last grapheme) and #788 (pts 2, 4–7: punctuation word boundaries, EOL and history-navigation off-by-ones).

## Notes for Reviewers

Every commit compiles warning-free and passes the full test suite and modules land fully tested before their callers, so the commits read in order.

**The four commits worth your attention**:

1. `core-editor: model cursor as range`: `Cursor` (anchor+head) as a view over `LineBuffer`.
2. `core_editor: enforce cursor rest policy at a single commit boundary`: check out the commit-boundary placement (`run_edit_command`, `set_buffer`, `set_edit_mode`).
3. `core_editor: total motion resolution`: `resolve_motion`'s inclusivity and edge cases. This is where motion semantics now live.
4. `feat(editor): operator verbs … one operate() sink`: the line-wise snap: Cut eats terminators, Change keeps them.

## Testing

Since vi mode has many features already I held back from implementing a whole example which leads you through all the motions and what they should do.
So testing will sadly be hard if you are not well versed in vi semantics, I would advise just opening vim on one window and run `cargo run --example demo -- --vi` in another and compare.
Therefore, I hope as many well versed vim users come to this PR and test out its capabilities and point out its flaws.

## Future Goals

- Depreciate old `EditCommand`'s to point out the new Operator + Verbs to downstream consumers and re-lower them as sugar over the verbs.
- Lift the Repeat engine out of Vi mode, so modes can share it and a mode comes closer to a raw key-map.
- Make counts part of motion resolution instead of event repetition: `2dj` should resolve one span of three lines (vim), not run `dj` twice (four lines). Falls out naturally once the repeat engine is lifted.
- Refactor the editor to speak `Cursor` instead of `insertion_point` and `anchor`, this would unify the whole semantics.
- Move the remaining vi mode-transition semantics to the commit boundary: Esc stepping the cursor back one (#694, #788 pt 1) is a rest-policy-adjacent rule, not a keybinding.
- Add the `Select` verb and helix mode as a producing mode for `RestPolicy::Block`, the one missing piece for helix range motions.
- Implement more features of helix mode with this at its base.
- Give registers a representation: `Granularity` already tags the cut buffer, a register key would generalize it to named registers and vi access to the system clipboard (#745).
- Lower text objects through the operator verbs: `ciw`/`ci` (produce a span like any motion does, so `Cut`/`Change`/`Copy` over a text-object target would retire another parallel command family (#848 then becomes "add targets", not "add commands").

## Attributions

Claude Opus 4.8/Fable 5 was used to write tests and do code-review.
In my experience it does this quite well, especially since legacy behavior was quite well defined.

Also I want to mention @sim590 and @matystroia, your analysis of the issues and our discussions had major influence in how I approached this rewrite.
Thanks a lot for your help!

And all the core-members for their guidance, especially @fdncred @stormasm


src/enums.rs
```diff
@@ -88,6 +88,21 @@ impl Default for TextObject {
...
+/// operator and stored on the cut buffer, so paste knows which to do.
+#[derive(Clone, Copy, Serialize, Deserialize, Debug, PartialEq, Eq, Default)]
+pub enum Granularity {
+    /// Inline at the cursor — the default.
+    #[default]
+    CharWise,
+    /// Whole lines, pasted below/above.
+    LineWise,
+}
+
```
Juhan280, line 158: Does this api leave room for future `BlockWise` implementation?

kronberger-droid: Yeah, should be as easy as adding an enum entry + an `operate()` arm + a paste arm.
But great that you point it out, since it would be good to make it `#[non_exaustive]` already now.

fdncred:
I've looked through every commit and while it's hard to understand whether everything works as intended by looking at a diff, here's my initial feedback.

- This is a significant change, a real root canal. We'll need to get this landed soon and announce it to get a good amount of testing on it. We have roughly 3 weeks left until the next release.
- Keep in mind we're cross platform. Are windows line endings vs linux/mac going to be a problem?
- I may have missed it but I didn't see multiline operations like selections, deletions, etc. Is these just handled by the nature of the changes?
- I noticed some `expect()`'s in the code which i'm not a fan of because they can panic which is always bad. If it's really not possible for them to panic then it's fine otherwise we should have better error handling. Thoughts?
- We also need to keep in mind that it would be good to be able to configure any keybinding in nushell. I can't really envision if that's possible from looking at a diff but it's something to remember because we do not have this capability now.
- Let's see what copilot thinks of it.

kronberger-droid:
@fdncred
Thanks for taking a look at it.

Let me address your questions:

Line endings should not be a problem, which are the only platform specific parts.
They are handled at the boundary to the terminal.
Inside only `\n` exists and is handled.
But you are right to be confused, since it seems i had a devensive `\r` handling and tests, which actually would not occur since they are stripped before.

Multiline is handled by matching on `Granularity` inside `operate()`. My line handling is in `line.rs`, if you want to take a look.

The `expect()` should be impossible to panic, but you are right I should handle them more gracefully.

The new verbs are still plain EditCommands and derive Serialize/Deserialize, so exposing Move/Cut/… + a MotionTarget to nushell config is straightforward as a follow-up and it'd actually let users bind more than they can today, rather than less.
And future expansions if needed will be easier as well.

src/engine.rs
```diff
@@ -972,6 +972,13 @@ impl Reedline {
...
             }
             if need_repaint {
+                // Sync the editor's edit mode before painting so the cursor is
+                // normalized under the current rest policy. A mode change that
+                // bypasses the command path (e.g. Esc → Vi normal) otherwise
+                // wouldn't clamp until the next command, painting the cursor past
+                // the last grapheme for a frame.
+                let mode = self.edit_mode.edit_mode();
+                self.editor.set_edit_mode(mode);
                 self.repaint(prompt)?;
```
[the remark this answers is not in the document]
kronberger-droid: addressed in `cd8bf28`

src/enums.rs
```diff
@@ -201,6 +333,53 @@ pub enum EditCommand {
         select: bool,
     },
 
+    /// Move the cursor to a [`MotionTarget`], collapsing any selection.
+    ///
+    /// The parameterized, human-readable face of the selection primitive.
+    /// `Move`/`Extend`/`Cut`/`Copy`/`Delete` over a [`MotionTarget`] are the
```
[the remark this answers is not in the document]
kronberger-droid: addressed in `cd8bf28`

src/core_editor/resolve.rs
```diff
@@ -0,0 +1,374 @@
...
+pub(crate) fn operator_span(buf: &str, origin: usize, target: MotionTarget) -> Cursor {
+    Cursor::new(origin, resolve_motion(buf, origin, target).op_end)
+}
+
+/// Resolve a public [`MotionTarget`] against `buf`, relative to `origin`.
+///
+/// Total over every variant — a target with no resolution yet (`Find`) stays at
+/// `origin` (a no-op) rather than panicking, so a target constructed from config
+/// or another mode can never crash the editor. Context-aware (takes `buf`), so
+/// line/buffer edges resolve correctly where a context-free conversion couldn't.
```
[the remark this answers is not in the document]
kronberger-droid: addressed in `cd8bf28`

src/core_editor/word.rs
```diff
@@ -0,0 +1,174 @@
...
+
+/// Sort `ch` into a [`CharClass`].
+pub(crate) fn categorize_char(ch: char) -> CharClass {
+    match ch {
+        '\n' => CharClass::Eol,
+        ch if ch.is_alphanumeric() || ch == '_' => CharClass::Word,
+        ch if ch.is_whitespace() => CharClass::Whitespace,
+        _ => CharClass::Punctuation,
+    }
+}
```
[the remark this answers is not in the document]
kronberger-droid: addressed in `cd8bf28`

stormasm:
I downloaded your reedline repo and ran everything locally on my machine as well as did a high level review of the PR...

It looks fine to me and without holding back on the progress and timeline of landing this PR --- over the next number of weeks and months as I digest the details and philosophy of the changes that you made I will definitely have more questions...

I agree with @fdncred  that if we are going to land this PR we should do it sooner rather than later and have some time to kick the tires and let everyone else see what the issues are that come up in nushell prior to the next release.

Having it integrated into Nushell will be the true test to see what tweaks need to be made to the Reedline code post your PR landing...

I like the idea of using nushell as the test case...

After we have tested Reedline via the Nushell integration then you can iterate on any changes that need to made to Reedline prior to doing a release.

In Nushell we can pull the Reedline code on our main branch without actually publishing out Reedline to crates.io for several days ?

This way you will not break any developers who use Reedline
as a dependency in their products... 

As I said in the core team meeting yesterday, I think the forward progress in Reedline and the approach you are taking is really good and will strengthen the overall architecture, maintainability and scalability for the longterm health of Reedline.

Great job on the PR !  And thanks for all of your work on Reedline 😄

src/enums.rs
```diff
@@ -88,6 +88,138 @@ impl Default for TextObject {
...
+    CharWise,
+    /// Whole lines, pasted below/above.
+    LineWise,
+}
+
+/// A human-readable, parameterized motion target — the public vocabulary every
+/// cursor motion lowers from.
+///
+/// `Move`/`Extend`/`Cut`/`Copy`/`Erase` over a `MotionTarget` are the
+/// going-forward motion API. They resolve through the private `resolve_motion`
```
stormasm, line 163: resolve_motion is no longer private ?  

so I just think the comment line has to be changed to reflect that...

kronberger-droid: How would i call a function that is not exposed through the public api, but inside the crate (`pub(crate)`)?

kronberger-droid: Maybe just remove it and say "They resolve through `resolve_motion`?

stormasm: Excellent question, I thought about that after I wrote it 😄 

Lets just leave it as "private" then since our audience is the outside world and the public API.

Not the internal developer who sees it as pub(crate).

sim590:
I did some tests and here is what I observed as deviating from normal Vi behaviour.

**Paste command doesn't yield the correct cursor position**

If I have a line like this with cursor on the position indicated by the caret:
```
aa bb cc
   ^
```

Pressing `deP` should yield:
```
aa bb cc
    ^
```
i.e, the cursor should end up on the last character of the deleted word. In short, `P` doesn't leave the cursor at the right position. Same for `p`. It's the same for any other `xp` or `xP`.

This is not a regression as current main branch also has this issue. Will you fix it in this PR?

**Pressing ESC to return to normal mode doesn't move the cursor backward one character**


```
aa bb cc
       ^
```

Pressing `i` and then `ESC` should yield:
```
aa bb cc
      ^
```

This was fixed by #1069. Therefore, if you claim your PR should also fix it, then you would need to address this.

The rest seems OK to me for now. I have not done the most exhaustive test though. I just did some quick tests. But I don't see anything else that is obviously wrong compared to current the main branch and what you claim to have fixed.

sim590:
I found another inconsistency (_not_ replicated on current main branch):

```
a
                                   <- empty line
b
```
If cursor is on character `b` and the command `dd` is issued, the cursor is thrown on the character `a` instead of ending on the empty line.

Another scenario, possibly related (replicated on current main branch):

```
a
                                   <- empty line
b
```

Cursor on character `b`. If command `cc` is issued followed by the ESC key, the cursor ends up on the empty line above, yielding this:
```
a
                                   <- empty line, cursor is on this line
                                   <- empty line
```
The cursor should have stayed on the 3rd line.

kronberger-droid:
@sim590
Thanks for the careful testing.

dd on last line (`2060691`) is fixed, was a regression like you suspected.
Esc not stepping back (`5980b76`) is fixed.

The other two i would land seperately, neither are regressions here.

Paste cursor (`xp`/`xP`) is pre-exising and orthogonal to the motion rework.

cc + Esc on empty line, is also pre-existing. It belongs to a broader pass to make the cursor line-aware on horizonatl moves. That one touches the egnine's mode-sync timing, so I'd rather not let it sink into this PR.

But both are already fixed and will come as a follow-up.

kronberger-droid:
@fdncred
Are you fine with landing this plus the small follow up this week, so i can work on making it available in nushell soon to get more feedback and testing in.

fdncred:
yup, go for it.

PR #1101, opened 2026-06-11
fix(editor): make horizontal grapheme motion line-aware via rest policy

kronberger-droid:
Follow-up to the motion unification in #1100.

Resolve_head now resolves Grapheme against the active RestPolicy: bounded to the line under the cell-caret policies (OnGrapheme/Block), free across lines under Between. h/l, Esc-back and emacs left/right all inherit it from this one spot.

The engine relays the policy to the editor at the parse seam (sync_edit_mode, non-committing) so a command a mode transition emits, the Esc->normal step-back, which resolves under the new policy. Committing there would pull a caret at the line end back a grapheme and double-step the move.

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -345,6 +345,19 @@ impl Editor {
         }
     }
 
+    /// Adopt `mode`'s rest policy *without* committing the cursor.
+    ///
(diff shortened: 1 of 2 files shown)


kronberger-droid:
@fdncred
Will merge this now and tomorrow I will take care of bumping reedline in nushell.

PR #1103, opened 2026-06-13
fix(vi): complete pending key sequence before custom keybindings

kronberger-droid:
## What

A custom normal-mode keybinding whose key matches the **second** key of an in-progress motion (e.g. `f<char>`) would hijack the sequence. The dispatch consulted keybindings on *every* keypress, so a binding on the find target fired instead of completing the motion.

For example, with a custom binding on `B`, typing `fB` ("find B") would trigger the binding rather than moving the cursor to the next `B`.

## How

Track whether a multi-key sequence is pending via a new `seq_completed` flag. While one is open, the next key is fed to the parser first; keybindings are only consulted once no sequence is in flight.

The flag is reset per-branch — `true` when the cache is cleared (complete/invalid), `false` only when a sequence is still incomplete — so it stays in lockstep with "is the cache non-empty."

## Notes

This **supersedes #693** by @crides, rebased onto the verb + target motion vocabulary introduced in #1100 (the original patch targeted the pre-refactor `MoveRightUntil` event shape and no longer merges). Original authorship preserved via `Co-authored-by:`.

Known boundary (same as the original PR): this covers normal/visual mode only. Insert-mode multi-key sequences (`^V`, `^K` digraphs in real vi) aren't handled, but reedline has no such insert sequences today, so it's not a regression.

## Test

Adds `pending_motion_beats_custom_keybinding`: with a custom `B` binding, `fB` resolves to a find motion rather than the binding. Full `edit_mode::vi` suite (143 tests) passes; clippy clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/edit_mode/vi/mod.rs
+++ b/src/edit_mode/vi/mod.rs
@@ -44,10 +44,6 @@ pub struct Vi {
     previous: Option<ReedlineEvent>,
     // last f, F, t, T motion for ; and ,
     last_char_search: Option<MotionTarget>,
-    // false while a multi-key motion (e.g. `f<char>`) is still awaiting input;
-    // such a pending sequence must be completed before custom keybindings get
-    // a chance to claim the next key
-    seq_completed: bool,
 }
 
 impl Default for Vi {
@@ -59,7 +55,6 @@ impl Default for Vi {
             mode: ViMode::Insert,
             previous: None,
             last_char_search: None,
-            seq_completed: true,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


kronberger-droid (commented):
Does what #693 intended. Wanted to test how claude does on this all on itself.

fdncred:
> Does what #693 intended. Wanted to test how claude does on this all on itself.

As long as you've reviewed the PR in detail and tested it out yourself, it's fine.

kronberger-droid:
> > Does what #693 intended. Wanted to test how claude does on this all on itself.
> 
> As long as you've reviewed the PR in detail and tested it out yourself, it's fine.

Okay for sure, reviewed it, but I still have to do some testing before I merge it.

PR #1105, opened 2026-06-14
fix: tmux history dedup

### LINKED ISSUE #1062: Prompt duplicated quite a lot in tmux history when typing at the top of a tmux window
macOS + Ghostty + tmux

Image

Haven't done a bisect yet, but it feels like a regression introduced recently.

## Steps to reproduce

1. open a tmux window with nu
2. type
3. enter tmux copy mode

blindFS:
Fixes #1062 

I've to admit I don't quite understand why and what the consequences would be.
I found this solution by blunt bisect debugging, and it works quite well according to my manual testing.

@fdncred @kronberger-droid 

Hopefully you guys can make sure it's safe to merge or maybe come up with a better solution.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -90,6 +90,11 @@ pub enum W {
     /// Discards all output, used in tests.
     #[cfg(test)]
     Sink(std::io::Sink),
+    /// Captures all output into a buffer so tests can assert on the exact
+    /// escape-byte stream the painter emits (not tmux-specific — any
+    /// output-level invariant).
+    #[cfg(test)]
+    Capture(Vec<u8>),
 }
 
 impl W {
@@ -105,6 +110,22 @@ impl W {
     pub(crate) fn sink() -> Self {
         W::Sink(std::io::sink())
     }
+
+    /// Writer that buffers everything written to it, so a test can inspect the
+    /// emitted bytes after painting.
+    #[cfg(test)]
+    pub(crate) fn capture() -> Self {
+        W::Capture(Vec::new())
+    }
+
+    /// Bytes captured so far. Panics unless this is a [`W::capture`] writer.
+    #[cfg(test)]
+    pub(crate) fn captured(&self) -> &[u8] {
+        match self {
+            W::Capture(buf) => buf,
+            _ => panic!("captured() called on a non-capturing writer"),
+        }
+    }
 }
 
 impl Write for W {
@@ -113,6 +134,8 @@ impl Write for W {
             W::Terminal(w) => w.write(buf),
             #[cfg(test)]
             W::Sink(w) => w.write(buf),
+            #[cfg(test)]
+            W::Capture(w) => w.write(buf),
         }
     }
 
@@ -121,6 +144,8 @@ impl Write for W {
             W::Terminal(w) => w.flush(),
             #[cfg(test)]
             W::Sink(w) => w.flush(),
+            #[cfg(test)]
+            W::Capture(w) => w.flush(),
         }
     }
 }
@@ -580,18 +605,7 @@ impl Painter {
         // Moving the cursor to the start of the prompt
         // from this position everything will be printed
         let anchor_row = self.prompt_start_row.last_known_row();
-
-        // HACK: Clearing from (0, 0) may lead to duplicated scrolling histories in tmux, #1062
-        if anchor_row == 0 {
-            self.stdout
-                .queue(cursor::MoveTo(1, anchor_row))?
-                .queue(Clear(ClearType::FromCursorDown))?
-                .queue(cursor::MoveTo(0, anchor_row))?;
-        } else {
-            self.stdout
-                .queue(cursor::MoveTo(0, anchor_row))?
-                .queue(Clear(ClearType::FromCursorDown))?;
-        }
+        self.clear_from_anchor(anchor_row)?;
 
         let layout = self.compute_layout(lines, menu);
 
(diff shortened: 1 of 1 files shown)


blindFS:
Hmmm, it will cause rendering issues of resizing...

blindFS:
Or maybe it's an issue of the upstream crossterm crate, and should be addressed there?

kronberger-droid:
@blindFS
I actually can't reproduce it.
I tried plain `nushell` and `reedline --examples demo` in tmux.
~~You mean if you scroll up in tmux right?~~
Any special things like small window or large prompt?
Both fixes feel weird still.
I will take a closer look the moment i can reproduce it reliably.

blindFS:
@kronberger-droid Thanks for looking into it. It could be specific to some options in my tmux setting, I'll try to show a minimal reproducible tmux config later.

blindFS:
@kronberger-droid Sorry for the delay, actually I can reproduce it with default/empty tmux config and default nushell config `nu -n`.

And both ghostty and system terminal app of macOS have that problem.

blindFS:
FYI, here's the reasoning of a relatively weak AI agent, (after several prompts where I actually provided my current fix)

```
Reedline (Nushell's engine) tracks the absolute position of the shell prompt so it can manage multi-line input, syntax highlighting, and ghost-text completions. It considers the very first cell (0, 0) to be the "Anchor Point."
When you issue a Clear command while the cursor is at (0, 0):
1. The Trigger: You are signaling to the terminal/tmux that the entire view, starting from the Anchor Point, is now "dirty" or "void."
2. The Shell Reaction: Reedline receives a state update from the terminal indicating that the origin of the prompt (the anchor) has been modified or erased.
3. The Panic: Reedline realizes its anchor point has been wiped. To "recover" and ensure the user doesn't lose their prompt, it force-redraws the prompt immediately. Because the terminal is already trying to clear, the two processes fight for control, causing that annoying duplicated prompt.
```

I can't tell whether it's true or not, I'd like to hear what SOTA models like gpt5.5 or opus 4.8 think about the problem if you guys have access to any of that level.

kronberger-droid:
Took a look (Claude Opus 4.8, driven by @kronberger-droid). I think this is fully explained by tmux's **`scroll-on-clear`** option, which is **on by default** (added in tmux 3.2), rather than anything in crossterm.

### What happens
On every keystroke, the repaint does:

```rust
MoveTo(0, anchor_row)      // ESC[<row+1>;1H
Clear(FromCursorDown)      // ESC[0J  → erase display, cursor → end
```

In tmux, `ESC[0J` is handled by `screen_write_clearendofscreen()`, which contains this branch (`screen-write.c`):

```c
if (s->cx == 0 && s->cy == 0 &&
    (gd->flags & GRID_HISTORY) &&
    ctx->wp != NULL &&
    options_get_number(ctx->wp->options, "scroll-on-clear"))
        grid_view_clear_history(gd, bg);
```

So **only** when the cursor is at the absolute top-left `(cx == 0 && cy == 0)`, tmux treats the erase-below as "clear the screen but keep its contents" and pushes the *entire current screen* into the scrollback grid (`grid_view_clear_history`) instead of discarding it. It's the same mechanism that makes plain `clear` preserve scrollback in tmux.

### Why it only happens at the topmost line
When you type at the top of the window, `anchor_row == 0`, so every repaint lands the cursor exactly on `(0,0)` and then clears → each keystroke copies the current prompt line into history → "duplicated quite a lot." When the prompt sits anywhere below row 0, `cy != 0`, the branch is skipped, tmux just discards the cleared cells, and there's no duplication. That matches your observation exactly, and your "off-by-1 in prompt start position" hunch was basically right — it's column 0 of row 0 specifically.

### Why your fix works
Moving to column 1 before the clear makes `s->cx == 1`, so the `cx == 0` guard fails and `grid_view_clear_history` is never called. The trailing `MoveTo(0, anchor_row)` puts you back at column 0 to print, and the single cell at `(0, anchor_row)` that the col-1 erase skipped is immediately overwritten by the reprinted prompt — so on-screen output is identical, but tmux no longer snapshots to history. The fix is correct and minimal.

### A couple of notes
- It's not really a crossterm bug, and arguably not a tmux bug either — `scroll-on-clear` is intentional. The regression-y feel comes from reedline full-clearing + repainting on every keystroke, which only collides with this option when anchored at row 0. You can confirm the diagnosis by toggling `set -g scroll-on-clear off` in tmux — the duplication should vanish without this patch. (This also fits "maybe a recent tmux update caused it" — `scroll-on-clear` defaulting on is relatively recent.)
- The weaker agent's explanation isn't accurate: there's no "terminal sends reedline a state update" and no two processes "fighting." Reedline never learns the screen was cleared; tmux silently copies the screen into history on the `(0,0)` clear, once per repaint.
- Optional: you could gate the dance on `anchor_row == 0` (the only case that can hit the guard), but doing it unconditionally is harmless since that cell is always overwritten.

kronberger-droid:
I just woke up and let it rip in the background. I would have compacted the message but i think its fine.

Workaround seems hacky but solves the problem.
I would say we could land it like this one maybe with checking the guard and adding a comment.
I keep it in mind and think about a nice solution, the painter is on my radar anyways.

blindFS:
> I just woke up and let it rip in the background. I would have compacted the message but i think its fine.
> 
> Workaround seems hacky but solves the problem. I would say we could land it like this one maybe with checking the guard and adding a comment. I keep it in mind and think about a nice solution, the painter is on my radar anyways.

Thanks! It explains the root cause exceptionally well, I just added some extra comments.

It would be great if reedline won't need to repaint everything on each keystroke in the future.

kronberger-droid:
> It would be great if reedline won't need to repaint everything on each keystroke in the future.

Yeah that's the goal.
But right now I am still quite unfamiliar with the painter, and just from a readability point of view its still quite a mess.

blindFS:
> > It would be great if reedline won't need to repaint everything on each keystroke in the future.
> 
> 
> 
> Yeah that's the goal.
> 
> But right now I am still quite unfamiliar with the painter, and just from a readability point of view its still quite a mess.

Anyways, I'm so glad to see an active maintainer of reedline nowadays. Thanks a lot for all of your work!

kronberger-droid:
Glad to be of help!
I will test a little bit and check if the menu can trigger this too. After I am sure I will land it.

kronberger-droid:
Just added a way to test this inside reedline and fixed the menu path which could also cause this bug.
Still using the hack tho, further improvements on the painter will be follow-up.

PR #1107, opened 2026-06-15
fix: re-use prompt rows after a quiet buffer editor

### LINKED ISSUE #1106: Prompt-after-editor behavior changed due to cursor::position optimization
**Platform** any
**Terminal software** multiple

With the changes of #1090, there's a behavioral change when running an external editor (e.g., via `<Ctrl+O>`). Previously, the updated input line (after running the editor) is rendered on the same original line. After the linked PR, the updated line is rendered on the line afterwards.

## Steps to reproduce
1. Run: `cargo run --example demo`
2. Enter some text (without hitting enter)
3. Type: <Ctrl+O>
4. In the editor, update the text, save, and quit.
5. See that the original input line is displayed, and the updated line is rendered on the line after it.

_(Aside: I found this by running some experimental automated `reedline` tests in a programmatically controlled `pty`.)_

## Screenshots/Screencaptures

After #1090 this is what the reproducing steps produce:

Image

reubeno:
The buffer-editor success path re-initialized the prompt position from scratch; with the cursor still sitting after the old buffer text, `select_prompt_row` would start a new prompt one row down, leaving the original input line above.

This change captures the painter's state before spawning the editor and restore with it; if the editor wrote nothing to the tty (or did so on the alternate screen only) this allows reusing the original prompt row. An editor that printed output should get a fresh row.

Closes #1106

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1985,6 +1985,12 @@ impl Reedline {
                     let mut file = File::create(temp_file)?;
                     write!(file, "{}", self.editor.get_buffer())?;
                 }
+                // Capture the prompt's screen range so that an editor
+                // which leaves the cursor untouched (e.g. a GUI editor,
+                // or one that wrote nothing to the tty) re-uses the
+                // existing prompt rows instead of starting a new prompt
+                // a row below the old one.
+                let suspended_state = self.painter.state_before_suspension();
                 {
                     let mut child = command.spawn()?;
                     // The child owns the tty now; invalidate eagerly so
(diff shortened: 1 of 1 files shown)


kronberger-droid:
Looks good, but the first comment is a little confusing, since after using any buffer editor (vim, helix) which also are techically not gui editors, the cursor gets reset to the beginning of the buffer.
So its not untouched right?
Maybe change it up a little to be more accurate, then we can land it.

reubeno:
@kronberger-droid I've pushed an update to the comment; are you okay with this, or would you like more details? I'm trying to balance between accuracy and verbose comments.

Editors like `vim` that only use the alternate screen generally don't emit any output to the original screen.

kronberger-droid:
Thanks!

PR #1109, opened 2026-06-24
feat(vi)!: visual mode on a unified Cursor + rest-policy model

kronberger-droid:
Hey all!
This is now the follow-up for the editor motion rewrite (#1100).
It should be as much legacy behavior preserving as possible + some bugs fixed, which were present before the rewrite and some which were introduced by it.
This is the last large PR from my side I promise... but it just did not make much sense to do this incremental.

## Summary

A single `Cursor` (head + optional anchor) becomes the one source of truth for the caret *and* selection, every word/motion command is lowered onto one "verb + target" path, and a per-mode **rest policy** governs where the caret may settle.
Replaces the old `insertion_point` + `selection_anchor` + `selection_inclusive` trio and the per-flavor `*_index` word scans.
On that foundation: a proper vi **visual mode** and the bugs the unified model exposed.

- **`Cursor`**: gap-indexed `(anchor, head)` range and a point when equal.
- **`RestPolicy`**:
    - `Between` (emacs / vi-insert)
    - `OnGrapheme` (vi-normal: caret never past the last grapheme)
    - `Block` (vi-visual: covers ≥1 grapheme) which applied at one idempotent `commit`.
- **`MotionTarget` / `put_cursor`**: the single motion sink where inclusivity is carried geometrically by the range, not a `selection_inclusive` flag.

## Breaking change ! (since the last release):

`PromptViMode` gains a Visual variant and `WordKind`'s `Small`/`Big` are renamed to `Word`/`LongWord` (plus a new `Unicode` for Emacs).
Both are pre-release-only churn, and just must coordinated with the matching nushell change.

## Behavior changes

- **Vi visual mode** is a real selection: `v` starts a min-width-1 selection that motions extend (was: `v` cleared). Opt-in cross-line block movement.

## Bug fixes (each with a regression test)

Some pre-existing on `main`, which surfaced during the refactor.
Unifying invariant: editing anchors on the **caret**, movement on the **head**.
They diverge only under a Block cursor (head one grapheme past caret), where the old code mixed them.

- **`immediately_accept` hang**: batch submit was gated behind `!immediately_accept`, spinning `read_line` forever, which is now unconditional.
- **Caret/head bugs**: corrupted the buffer (`r` in visual: `"hello"`→`"lxlo"`), panicked (`move_line` across lines), or mis-cut (`cut_to_end`).
- **Visual selection geometry**: selecting motions bypassed `put_cursor`, dropping a grapheme on word-end / horizontal reversal / vertical `j`/`k`; all now route through it.
- **CRLF**: `Block` policy and linewise cuts (`dd`, `CutToEndLinewise`) orphaned the `\r` and the CR contract is documented on `LineBuffer`. *(History recall normalizing CRLF→LF is out of scope and will be a follow-up.)*

## Where to look

| Layer | Key commits | Files |
|---|---|---|
| 1. Cursor as SSOT | `c39e149` `ebfc4b7` `fad96df` | `cursor.rs`, `line_buffer.rs` |
| 2. Motion unification | `6190fb7` `016f288` `6c2c622` | `resolve.rs`, `word.rs` |
| 3. Rest policies | `48ba880` `0506dde`/`c16ae93`/`47531f8` | `rest_policy.rs` |
| 4. Visual mode | `b90c539` `5596c9b` `784cc86` | `rest_policy.rs`, `prompt/base.rs` |
| 5. Fixes + readability | `c4e7e5f` `ab15f2f` `9330b14` `7cdbd40` | `editor.rs`, `line_buffer.rs` |

**Scrutinize** (densest invariants):
- `commit`/`recohere` (`rest_policy.rs`): total + idempotent, the one normalization boundary
- `put_cursor` (`cursor.rs`): the anchor-flip on reversal
- `resolve_motion` (`resolve.rs`): totality + the block word-end `e` landing.

**Skim**: `7cdbd40` (pure bool→enum rename: `CaretGeometry`/`Direction`); the `word.rs` classifier (tested in isolation) and the `test(…)` commits.

## Testing

`cargo test` (1160 lib + 28 integration), clippy `--all-targets`, `fmt` — all green.
`cargo run --example selection_styling` drives visual mode.

## Attribution

Thanks to everyone pointing out regressions and bugs on the current main.
Also I used Claude Opus 4.8 to write large parts of the test harness pinning down legacy behavior and used it to review code changes to point out bugs and unsound code changes.


fdncred:
cool. i can make a nushell PR with the latest reedline commit if you want. Let me know.

kronberger-droid:
@fdncred
On it right now, since i want to add a maybe a small list of things to look for when testing to it and share it on the discord. One question that comes up. We now have a real visual mode, which could be differentiated with a prompt indicator, we could think about if we want one, right now I will reuse the normal mode one.

fdncred:
ok, sounds good. a new indicator would be good i think but we can reuse normal if you want for now and add a new indicator later. no biggy.

kronberger-droid:
What do you think about `~` its the cleanest one i could think of for now.
Otherwise I would keep it as normal for now.
Also we could do a custom cursor if we want, I use usually a `_` cursor for visual.

fdncred:
As long as `~` isn't confused as the home dir, i'm ok with it in this context. I'm fine with changing the cursor too. It should be apparent that people have entered visual mode so it doesn't happen by accident, or at least when it does happen by accident, they know what's going on.

kronberger-droid:
Well good point, I will think about a better one.

PR #1111, opened 2026-06-25
Fix abbreviation expansion inserting a double trailing space

### LINKED ISSUE #1108: Abbreviation expansion double-inserts a trailing space
**Platform** Linux (NixOS 26.05)
**Terminal software** WezTerm (0-unstable-2026-03-31) 
**Nushell** 0.113.1

## Description

When reedline expands abbreviations, it double-inserts a space character, one before your cursor and another after.

It breaks history completion since (for example) `git switch<space><space>` is not a substring of `git switch main`. I have to delete the trailing space to see the completion.

Filing here since I think it [happens within reedline](https://github.com/nushell/reedline/blob/f6539a5eda9bc58874a5a949c2d9b68b7c58f89c/src/engine.rs#L1821-L1833).

## Steps to reproduce

1. Have a `git switch main` in shell history (or any branch). You should see it as ghost text completion.
2. Define an abbreviation for it:
  ```nu
  $env.config.abbreviations.gs = "git switch"
  ```
3. Type: `gs` and hit <kbd>Space</kbd>

### Expected (expanded, no trailing space)

```
git·switch·<cursor>
```

### Actual (expanded, trailing space after cursor)

```
git·switch·<cursor>·
```

MerlijnW70:
## Summary

Expanding an abbreviation by pressing `<space>` inserted **two** trailing
spaces instead of one — one before the cursor and one after. This breaks
history-based completion: e.g. `git switch  ` (two spaces) is not a prefix of
`git switch main`, so the ghost-text completion disappears until the extra
space is deleted manually. Fixes #1108.

## Fix

On a `<space>` expansion the triggering space has already been inserted into
the buffer before `try_expand_abbreviation_at_cursor` runs. The replacement
selected only the abbreviation word (`word_start..word_end`), leaving that
space in place, and then inserted the expansion followed by its own suffix
space — yielding two trailing spaces.

The selection now extends through `cursor_position_in_buffer` instead of the
end of the word, so the already-inserted triggering space is replaced by the
expansion's suffix. The submit (`<enter>`) path is unaffected: there the cursor
sits at the end of the word, so the selected range is unchanged.

## Test

Adds `abbreviation_expands_on_space_without_double_space`, which exercises the
`<space>` expansion path (`submitted = false`) — previously untested; all
existing abbreviation tests only covered the submit path. It asserts that
expanding `gc` → `git commit` after the triggering space yields exactly
`git commit ` (one trailing space), and fails against the old behaviour
(`git commit  `).

## Reference

Reported in #1108. The issue pinpointed `src/engine.rs` abbreviation handling
as the cause.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1825,7 +1825,10 @@ impl Reedline {
                     select: false,
                 },
                 EditCommand::MoveToPosition {
-                    position: word_end,
+                    // Select through the cursor, not just the end of the word, so
+                    // the triggering space (already inserted on a <space> expansion)
+                    // is replaced rather than left beside the inserted suffix.
+                    position: cursor_position_in_buffer,
                     select: true,
                 },
                 EditCommand::InsertString(format!("{}{}", expansion, suffix)),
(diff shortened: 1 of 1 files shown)


fdncred:
lgtm. what do you think @kronberger-droid ?

kronberger-droid:
Seems fine to me to.
I am just testing it and check how/if it interacts with my #1109.
I will keep you all updated.

kronberger-droid:
Found some things which are orthogonal to this change. Fine to land.

fdncred:
Thanks!

MerlijnW70:
@fdncred   your welcome,

PR #1112, opened 2026-06-28
feat(history): add typed inherent methods and JSON filter to SqliteBackedHistory

eitsupi:
## Summary

Three typed inherent methods are added to `SqliteBackedHistory` (`save_with_extra`, `load_with_extra`, `search_with_extra`) for round-tripping `more_info` without type erasure. `History::save` now delegates to `save_with_extra`, so both paths share the same implementation. A new `JsonFilterValue` enum and `SearchFilter::more_info_json` field enable type-precise filtering by JSON path at the SQL level. The `History` trait is unchanged.

## Motivation

#1011 made `HistoryItem` generic over `ExtraInfo` and exposed `HistoryItemExtraInfo` publicly. However, the `History` trait still uses the type-erased `HistoryItem<IgnoreAllExtraInfo>`, so any `more_info` written through the trait is silently discarded on the next read. The generics are in place, but there is no usable end-to-end path to preserve custom metadata.

Downstream crates that want to persist typed metadata in `more_info` necessarily hold `SqliteBackedHistory` directly, since `FileBackedHistory` has no `more_info` support at all. Typed inherent methods on `SqliteBackedHistory` model this capability boundary precisely, without changing the `History` trait.

### DIFF
--- a/src/history/base.rs
+++ b/src/history/base.rs
@@ -37,6 +37,25 @@ pub enum SearchDirection {
     Forward,
 }
 
+/// A typed expected value for [`SearchFilter::more_info_json`] filters.
+///
+/// Each variant maps to a JSON scalar type and produces a type-precise SQL comparison
+/// using `json_type()` to avoid collisions between JSON booleans, integers, and strings.
+/// Arrays and objects are intentionally unsupported.
+#[derive(Clone, Debug, PartialEq)]
+pub enum JsonFilterValue {
+    /// Match a JSON `null` at the given path.
+    Null,
+    /// Match a JSON boolean (`true` or `false`) at the given path.
+    Bool(bool),
+    /// Match a JSON integer at the given path. Does not match JSON `true`/`false` or strings.
+    Integer(i64),
+    /// Match a JSON floating-point number at the given path.
+    Real(f64),
+    /// Match a JSON string at the given path. Does not match numbers or booleans.
+    Text(String),
+}
+
 /// Defines additional filters for querying the [`History`]
 pub struct SearchFilter {
     /// Query for the command line content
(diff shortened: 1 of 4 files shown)


fdncred:
cool. i remember adding more_info thinking someday someone would use it. what are you using it for?

eitsupi:
> what are you using it for?

My priorities were distinguishing between special meta-commands available within the REPL and recording the REPL's output.
https://github.com/eitsupi/arf/blob/d95cd3f875fd495909034050e3b1884e9291bde1/crates/arf-console/src/history/search.rs#L96-L100

Furthermore, I imagine this would be useful for this tool with REPL that can connect to various databases, such as for distinguishing between different databases.
https://github.com/columnar-tech/databow

fdncred:
cool. thanks for sharing.

eitsupi:
My first use case: eitsupi/arf#236

fdncred:
@kronberger-droid I'd probably wait on this one too until after the release unless you feel strongly about landing it now.

kronberger-droid:
Yeah agreed.

kronberger-droid:
Went through it now.

Two things before we land it:

1. There is no `update_with_extra`.
  The trait `update` is a load+save through the erased path, so `load` drops `more_info` and `save` writes it back as `null`.
  `update_last_command_context` goes through there, which means the normal post command bookkeeping wipes whatever `save_with_extra` wrote.

2. `JsonFilterValue` should be `#[non_exhaustive]`.
  The docs say arrays and objects are intentionally unsupported, which sounds like a later addition.

Also the trait `search`/`count` silently null out `more_info_json` instead of erroring, so a `&dyn History` caller quietly gets broader results.

Happy to land once the update path (1) is sorted.
The others are optional i guess.
Thanks!

eitsupi:
Thank you for your review; I believe I have addressed the points you raised.

The fact that search/count does not take more_info into account is intentional, and I believe it is reasonable for the History trait methods to ignore more_info since it only works with the SQLite backend.

eitsupi:
I found a bug related to transactions, so I fixed it in the last two commits and added tests.
I believe it's merge-ready.

kronberger-droid:
I will have Claude run over it once, then I would also merge.
@fdncred fine for you too?

fdncred:
ya, i'm good with this. we can land this when you're ready.

kronberger-droid:
Fine from my side, too.
After we merge this i will bump reedline on nushell to include the last to changes.

fdncred:
ok, have at it

PR #1113, opened 2026-06-29
feat(core): split Extend into CoverLanding vs Span

kronberger-droid:
## Summary

`EditCommand::Extend` tangled two things that are not the same: where the caret rests (`CaretGeometry`), and how far a motion drags the head.

This splits the second one out into its own axis.

```rust
pub(crate) enum SelectionExtent {
    CoverLanding,
    Span,
}
```

`Extend` branches on it now.
Vi modes take `CoverLanding` and go through `put_cursor`.
Bar modes take `Span` and go through the new `Cursor::extend_span`, which places the head at the motion's gap-indexed `op_end` without the vi-visual widening.
Both strategies funnel through `Editor::place` (`set_cursor` + `commit_cursor`).

`Span` is Helix groundwork, since Helix wants a selection that extends rather than covers.

No public API change, everything new is `pub(crate)`.

### Before

`Extend` had one placement strategy.
Geometry and extent were decided together, so a mode could not pick a block cursor and span extension independently.

### After

The two axes are independent.
No current mode pairs `Block` with `Span`, so that combination is only exercised by unit tests.

Under a block cursor `extend_span` keeps the anchor grapheme covered through a reversal, using the same `flip_anchor` as `put_cursor`.
That is the piece Helix will lean on.

## Notes

Bar modes now take the head from op_end rather than put_cursor, which only differs for inclusive motions like a forward Find, and no default binding reaches that.

### DIFF
--- a/src/core_editor/cursor.rs
+++ b/src/core_editor/cursor.rs
@@ -64,6 +64,37 @@ impl CaretGeometry {
     }
 }
 
+/// How a mode's `Extend` grows a selection — the selection-model axis, chosen
+/// per edit mode by [`PromptEditMode::selection_extent`](crate::PromptEditMode).
+/// Orthogonal to [`CaretGeometry`]: geometry is *where the caret rests*, extent
+/// is *how far a motion drags the head*.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum SelectionExtent {
+    /// Vi visual: the block cursor sweeps the grapheme it lands *on*, so head
+    /// covers the motion target: `vw` over `"foo bar"` selects `"foo b"`.
+    CoverLanding,
+    /// Helix: the selection is the gap-indexed operator span (`op_end`), so the
+    /// head stops at the motion boundary: `w` selects `"foo "`, caret on the
+    /// space.
+    Span,
+}
+
+/// Keep the anchor's grapheme covered when a block selection reverses direction
+/// across it (Helix `Range::put_cursor`): the anchor hops to the far edge of its
+/// grapheme so it stays inside the range. A bar (exclusive) selection is half-open
+/// `[min, max)`, so its anchor never moves.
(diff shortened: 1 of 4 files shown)


u-k-g:
tested on aarch64-darwin with cargo test --all-features -- --test-threads=1: 1,216 tests passed. The Helix example also builds and runs.

<details>
<summary>cargo test --all-features -- --test-threads=1; cargo run --example helix --features helix</summary>

```console
$ cargo test --all-features -- --test-threads=1; cargo run --example helix --features helix

Finished `test` profile [unoptimized + debuginfo] target(s) in 0.06s
Running unittests src/lib.rs (target/debug/deps/reedline-f3d584acdddd1871)

running 1217 tests
<redacted due to verbosity>

test result: ok. 1216 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.09s

Doc-tests reedline

running 29 tests
<redacted due to verbosity>

test result: ok. 29 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.89s

Compiling crossterm v0.29.0
Compiling reedline v0.48.0 (/Users/uzair/01-projects/reedline)
Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.42s
test src/menu/menu_functions.rs - menu::menu_functions::string_difference (line 207) ... ok

test result: ok. 29 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.89s

Compiling crossterm v0.29.0
Compiling reedline v0.48.0 (/Users/uzair/01-projects/reedline)
Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.42s
Running `target/debug/examples/helix`
Helix edit mode demo:
Default mode is insert (`:` prompt), so you can type words.
Press Esc for normal mode.
Press `i` to return to insert mode, or `a` to insert after the current selection.
Only `h`/`l` motions are currently implemented.
Abort with Ctrl-C
~/01-projects/reedline: helix test.

Aborted!
```

</details>

PR #1114, opened 2026-06-30
feat: add support for auto-pairs

altacountbabi:
Adds `InsertPair` and `BackspacePair` edit commands

### DIFF
--- a/examples/auto_pairs.rs
+++ b/examples/auto_pairs.rs
@@ -1,0 +1,25 @@
+// Create a reedline object with automatic pairs.
+// cargo run --example auto_pairs
+
+use reedline::{AutoPairs, DefaultPrompt, Reedline, Signal};
+use std::io;
+
+fn main() -> io::Result<()> {
+    let auto_pairs = AutoPairs::new([('(', ')'), ('[', ']'), ('{', '}'), ('"', '"'), ('\'', '\'')]);
+    let mut line_editor = Reedline::create().with_auto_pairs(auto_pairs);
+    let prompt = DefaultPrompt::default();
+
+    loop {
+        let sig = line_editor.read_line(&prompt)?;
+        match sig {
+            Signal::Success(buffer) => {
+                println!("We processed: {buffer}");
+            }
+            Signal::CtrlD | Signal::CtrlC => {
+                println!("\nAborted!");
+                break Ok(());
+            }
+            _ => {}
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


kronberger-droid:
@altacountbabi
Love it, great idea thanks.

We are close to a code freeze before the next release tho, so I think it makes sense to wait until after to merge.
@fdncred or do you think its worth propagating it into Nushell before the release?

fdncred:
@kronberger-droid I'd probably wait on this one until after the release. I'd also like to see a nushell counterpart PR to this one that enables or disables this feature.

eitsupi:
I've independently implemented auto-pairs in a reedline-based project (https://github.com/eitsupi/arf), so I'd like to share some issues I encountered.

## Two known pitfalls with this design

### History selection leaves orphan closing chars

When the cursor is inside a pair (e.g. `foo(|)`) and the user selects a history entry, reedline's `replace_in_buffer` uses `span.end = cursor_pos`, so it replaces only `foo(` with `foo()`, leaving the original `)` behind → `foo()|)`. The engine would need a post-replacement cleanup step when auto-pairs is active.

### Paste duplicates closing chars

Pasting `()` produces `())`. This requires `use_bracketed_paste(true)` to avoid; the example and docs should call this out explicitly (or `with_auto_pairs()` could enable it implicitly).

## Design suggestion

The `AutoPairs` struct has no way to skip insertion based on context, e.g., don't insert inside a string literal. A predicate like:

```rust
pub fn with_skip_if<F: Fn(&str, usize, char) -> bool + Send + Sync + 'static>(mut self, f: F) -> Self
```

would let callers inject language-specific logic without shadow state.
Without it, projects with quote-awareness will still need to work around this at the keybinding layer.

kronberger-droid:
@fdncred 
I am really tempted, since I think it was definitely a missing piece. But I don't have enough time until the release to test out the blast radius, so lets wait.

But @eitsupi comment makes waiting even more reasonable.

fdncred:
> But @eitsupi comment makes waiting even more reasonable.

Ya, I'd like to see some code to help with this. I've experienced this in code editors and it's a pain sometimes, "NO, I ALREADY HAD A CLOSING PAREN. DON'T ADD ANOTHER!!!!"

I probably wouldn't want to land this without something helping with it aggressively closing pairs.

kronberger-droid:
> "NO, I ALREADY HAD A CLOSING PAREN. DON'T ADD ANOTHER!!!!"

Haha very relatable.

Yeah let's try to get this right.
@eitsupi every input is welcome!

eitsupi:
Ok, when I have some free time, I'd like to make some modifications to this branch based on my own experience.

eitsupi:
I've opened a draft PR #1135.

Later, I'll try replacing the existing arf functionality with the changes in that PR and see if the API is sufficient.

kronberger-droid:
superseded by #1135, attribution for @altacountbabi is preserved.

PR #1117, opened 2026-07-09
Asyncronous prompt updates

philocalyst:
## Summary 

So this fixes my own issue nushell/nushell#18642. It's meant to be composed with an upcoming Nushell PR, which I will link here when I have it up. My biggest motivation is that I'm working on changes to starship that should bring these different kinds of prompts to users, and nushell is my daily driver. Wanted it here too, of course :0

### Before 

Before this, there existed no way to repaint the prompt after the first pass.

### After 

After this, the mechanism exists for repainting, opening use-cases for things like:
- An accurate time prompt
- Asynchronus git status (non-blocking)
- Gradient madness?? 

## Additional notes 

- I noticed that other features, like external_printer are gated, not sure if that should be the case here too.
- We don't have any real integration tests against a PTY which I was surprised to see, which makes true testing here hard, but I did verify myself.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -52,7 +52,10 @@ use {
         io::Result,
         io::Write,
         process::Command,
-        sync::{atomic::AtomicBool, Arc},
+        sync::{
+            atomic::{AtomicBool, Ordering},
+            Arc,
+        },
         time::Duration,
         time::SystemTime,
     },
@@ -208,6 +211,10 @@ pub struct Reedline {
     // `Signal::ExternalBreak` with the current buffer contents.
     break_signal: Option<Arc<AtomicBool>>,
 
+    // External repaint signal: when triggered, the prompt is re-evaluated
+    // and repainted in place while `read_line()` is running.
+    repaint_signal: Option<RepaintSignal>,
+
     // Maximum time to block on input before yielding control for features that
     // require periodic processing (external printer, idle callback).
     // Only used when external_printer or idle_callback is configured.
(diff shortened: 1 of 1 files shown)


philocalyst:
Resolved comments

fdncred:
@philocalyst is this part of/related to https://github.com/nushell/nushell/pull/18334 or 100% separate?

philocalyst:
This is entirely unrelated actually! This covers prompts while that covers completions. Thanks for asking though. There will be a follow up PR in Nu once this gets merged. 

Moved to a slower computer and started noticing these things :)

fdncred:
Thanks for the info @philocalyst. It would be good if you could take a look at https://github.com/nushell/nushell/issues/18613 to see if there's anything missing that would help with this request.

philocalyst:
Hm, seems alright to me! The Nushell side will be more complex, I figure.

fdncred:
ok, let's get on with it. :) Thanks for this. Been waiting for something like this for along time!

fdncred:
@philocalyst Are you planning on implementing something in nushell to take advantage of this PR?

philocalyst:
It seemed like the oh-my-posh guys were on that, but I'll follow up in a few weeks if they didn't pick it up.

fdncred:
They're not writing nushell rust code. They're looking for us to support it in reedline, which we do, thanks to your PR, and then they needed a nushell command like `commandline repaint` or `commandline set-prompt --left $record` like Jan mentions here https://github.com/nushell/nushell/issues/18613. So, that's what I'm hoping you have time to write.

philocalyst:
I'll do that then, yes -- do want to get my non-blocking completions PR through first. There's a more ambitious cross-cutting refactor that I want to get through, that would make the logic a bit cleaner. After that, I'll drive this through. Just don't want to split attention when it's all _so_ close to merge.

fdncred:
It's already landed. :)

philocalyst:
Yay!!! I was talking about a proper cache that I was building on top of my changes, that'll go up today.

In the meantime I'll begin thinking about this!

Apologies if this feels overwhelming, just really motivated to work on this stuff lately, exciting to see it come in to place towards a shell experience that never blocks UX!

fdncred:
I'm excited too. Keep 'em coming! ❤️

PR #1118, opened 2026-07-11
Add `with_persistent_menus` option

maxim-uvarov:
Disclosure: the code was written by an AI agentunder my direction; I can't review Rust line-by-line myself. What was verified: the full reedline test suite plus a new 4-case test matrix (quick × persistent), cargo fmt --check / clippy with CI flags, and manual behavior checks in a live nushell build with both flag values. I've wanted this behavior for a long time and drove the design decisions. Please review accordingly.

## Summary

When a menu is active and the user erases the characters they typed to filter it, the engine dismisses the menu. Two hardcoded paths in `engine.rs` do this: with quick completions on, any `Backspace`/`BackspaceWord`/`MoveToLineStart` deactivates the menu; and regardless of quick completions, any edit that leaves the line buffer empty deactivates it. Neither was configurable.

Erasing the filter is a natural part of browsing completions — narrowing too far and backing up shouldn't throw the menu away, especially since activating a menu on an empty buffer is already valid and shows all values.

Public API change: a new builder option, `Reedline::with_persistent_menus(bool)`, default `false`. With the default, observable behavior is unchanged. With `true`, edits refilter an active menu instead of dismissing it; the menu still closes on `Esc`, `Ctrl-C`, or when a value is accepted.

### Before

With a menu open, erasing input dismisses it:

- quick completions on: the first backspace closes the menu;
- quick completions off: the menu survives backspaces until an edit empties the line buffer, then closes.

### After

Default (`with_persistent_menus(false)`): same as before.

`with_persistent_menus(true)`: erasing characters (or emptying the line) sends `MenuEvent::Edit` like any other edit, so the menu stays open and refilters — down to the full unfiltered list on an empty buffer.

## Additional notes

A follow-up nushell PR wires this as `$env.config.completions.persistent_menus` once a reedline release includes it.

Covered by a new `rstest` case matrix (quick × persistent) in `src/engine.rs`.

### DIFF
--- a/examples/persistent_menus.rs
+++ b/examples/persistent_menus.rs
@@ -1,0 +1,98 @@
+// Create a reedline object with persistent menus
+// cargo run --example persistent_menus
+//
+// "he" [Tab] opens the completion menu filtered to the "hello ..." entries.
+// Erase with [Backspace] — even past "h" to an empty line — and the menu
+// stays open, refiltering back up to the full list of options.
+//
+// Change `with_persistent_menus(true)` to `false` to see the default
+// behavior: with quick completions on, the first [Backspace] dismisses the
+// menu; without quick completions, the menu closes once the line is empty.
+
+use reedline::{
+    default_emacs_keybindings, ColumnarMenu, Completer, DefaultPrompt, Emacs, KeyCode,
+    KeyModifiers, Keybindings, MenuBuilder, Reedline, ReedlineEvent, ReedlineMenu, Signal, Span,
+    Suggestion,
+};
+use std::io;
+
+// Prefix completer that also suggests on an empty line (like a shell does);
+// `DefaultCompleter` returns nothing for empty input, which would make the
+// kept-open menu invisible right when this example wants to show it.
+struct PrefixCompleter {
+    commands: Vec<String>,
+}
+
+impl Completer for PrefixCompleter {
+    fn complete(&mut self, line: &str, pos: usize) -> Vec<Suggestion> {
+        let filter = &line[..pos];
+        self.commands
+            .iter()
+            .filter(|command| command.starts_with(filter))
+            .map(|command| Suggestion {
+                value: command.clone(),
+                span: Span::new(0, pos),
+                ..Suggestion::default()
+            })
+            .collect()
+    }
+}
+
+fn add_menu_keybindings(keybindings: &mut Keybindings) {
+    keybindings.add_binding(
+        KeyModifiers::NONE,
+        KeyCode::Tab,
+        ReedlineEvent::UntilFound(vec![
+            ReedlineEvent::Menu("completion_menu".to_string()),
+            ReedlineEvent::MenuNext,
+        ]),
+    );
+}
+
+fn main() -> io::Result<()> {
+    let commands = vec![
+        "hello world".into(),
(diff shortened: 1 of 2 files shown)


fdncred:
I'd like to see how this work in an example so I can try it out before landing it.

kronberger-droid:
Code looks good, not much to say since it plugs into the existing behavior.
If the example shows the expected behavoir its fine to land from my side.

maxim-uvarov:
@fdncred @kronberger-droid below is the video demo. And https://github.com/nushell/nushell/pull/18573 is the next change that is needed to use the reedline feature in nushell 

https://github.com/user-attachments/assets/26e7547e-2ccf-4559-9af2-1b226a51b9ec

fdncred:
@maxim-uvarov This is fine but I'd like to see an example in the examples folder that demonstrates this reedline feature without nushell. I want to be able to run it myself without nushell to see how it works. Just look at the other examples as a guide on how to create one.

maxim-uvarov:
@fdncred, this? 
Added examples/persistent_menus.rs — runnable standalone:

`cargo run --example persistent_menus`

Type he, press Tab, then erase with Backspace past the start — the menu stays open and
refilters back up to the full list. Change with_persistent_menus(true) to false in the
example to compare with the current behavior (first Backspace dismisses the menu, since
quick completions are on).

Note: the example uses a small inline prefix completer instead of DefaultCompleter,
because DefaultCompleter returns no suggestions for empty input — the kept-open menu would
be invisible exactly at the empty-line moment the example demonstrates.

Also extended the test matrix to cover the other two commands in the quick-dismissal arm
(BackspaceWord, non-selecting MoveToLineStart).

fdncred:
yup, that's what i meant. very cool. thanks!

fdncred:
I'll let @kronberger-droid try it before we land it incase he sees something I missed.

fdncred:
Yes, I'm impatient. ;)

blindFS:
Do we still need #993 after this feature? Or maybe we should revert this if that one gets revised and merged?

maxim-uvarov:
@blindFS I read through the description of #993, and it seems to do a very different thing. Though again, please apologize me if I'm wrong. I cant read rust. 

In this PR we added a setting that makes reedline keep the previously spawned menu open when all the characters entered before are deleted. So it has nothing to do with automatic activation.

Meanwhile, in the description of #993, the menu appears automatically:

> When enabled, completion menu appears automatically as you type (like fish/IDEs)
> Menu activates after 1+ character typed for any argument
> Menu deactivates on space (waiting for next word)

blindFS:
@maxim-uvarov Yeah, technically they are different. But both PRs provide a new option that controls how menus get refreshed. It's not very clear to me how those options interact with each other, are they totally orthogonal (like this one controls when to close a menu, the other one controls when to create/renew a menu), or do we need sort of precedence so one option overrides the other? And I'm afraid that kind of complications will cause new troubles.

We're probably good if those options are indeed orthogonal, otherwise we'd better enumerate typical menu refreshing behaviours and combine both options into one.

Of course, that's the story after #993. Just wanna bring it up for discussion in advance.

maxim-uvarov:
Got it, thanks. I'm happy with whatever you decide

PR #1119, opened 2026-07-13
Document default Emacs and Vi keybindings

### LINKED ISSUE #350: Template table with the default keybindings and their actions
Have a table in our documentation that can be presented to end-users with the default keybindings (for emacs as well as vi with a little description). Some facilities exist to query the existing keybindings, but for end-users that don't want to configure their keybindings a more human readable form would be appreciated. Users of the crate could adapt the table to their specific needs

Semak12345:
## Summary

Closes #350.

Adds a user-facing reference for the default Emacs and Vi keybindings, including movement, editing, selection, control, Vi motions, operators, and text objects. Links the new guide from the README outline.

### Before

Users had to inspect the keybinding constructors and Vi parser to discover the defaults.

### After

The defaults are collected in tables in KEYBINDINGS.md, with notes about terminal-specific and feature-gated bindings.

## Additional notes

Documentation-only change. Validated with markdownlint-cli2 and git diff --check.

### DIFF
--- a/KEYBINDINGS.md
+++ b/KEYBINDINGS.md
@@ -1,0 +1,165 @@
+# Default keybindings
+
+Reedline provides Emacs and Vi editing modes. The tables below describe the
+bindings returned by `default_emacs_keybindings()`,
+`default_vi_insert_keybindings()`, and `default_vi_normal_keybindings()`.
+Applications embedding Reedline can add, replace, or remove these bindings.
+
+Key names use `Ctrl`, `Alt`, and `Shift` for modifiers. On some terminals,
+`Alt` may be entered as `Esc` followed by the key.
+
+## Emacs mode
+
+### Movement and history
+
+| Key | Action |
+| --- | --- |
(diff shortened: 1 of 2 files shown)


kronberger-droid:
Nice, this one was on my mind for a long time, thanks for picking it up. @Semak12345 
One structural request before we land this: could the shared bindings get their own section?

`keybindings.rs` has four shared sets.
Right now they're all filed under Emacs with Vi insert referring back in prose, which means:

- the "Emacs-only ... **such as**" list is non-exhaustive, so you can't tell what
  Vi insert has
- `Ctrl-Shift-x/c/v` are under Emacs → Selection, but they're in `add_common_edit_bindings`, so also in Vi insert, absent in Vi normal/visual
- Vi normal lacking the common edit bindings is only implied by subtraction

A `## Common bindings` section (one table per `add_common_*`) would move the
tables mostly verbatim.

kronberger-droid:
Superseded by #1197

PR #1122, opened 2026-07-17
Fix for surfaced unlayout in Nushell

philocalyst:
### Before 

Menu geometry was calculated on event.
### After 

Menu geometry calculated on interaction
## Additional notes 

Enhanced the fake completor to include description, which should verify a "complete" layout.
Other changes to make this cleaner were also included, wanted to ensure this new location for it felt like a good fit, rather than just copying and pasting.


fdncred:
Let's move forward with this and see if we can fix the carapace issue somehow.

PR #1123, opened 2026-07-19
Making Reedline aware of the status of completors

philocalyst:
### Before 

- Before, with recently added changes, reedline lied totally unaware the state of completion, as it was originally entirely tied to the its own lifecycle. With recent changes, it now has its own lifecycle, which left reedline totally in the dark, leading to poor behavior in many cases (largely in renderin).

### After 

- Reedline is now aware of the state a completor is in, allowing it to make rendering decisions that reflect that state.
- Flashing is now removed, which was a symptom of reedline updating under new context. It starts *with* that context now, and is able to respond effectively.

## Additional notes 

- I definitely took some liberties here in refactoring some of the impacted functions, in an effort to not let the codebase slip when adding these new conditions. The goal was to add these features will making performance *better* and I think I accomplished that.
- There's a corresponding PR in Nushell coming soon.
- Added a regression test for flashing.
- Biggest refactor was in `create_string` -- where I got both paths down to one/two allocations

### DIFF
--- a/examples/persistent_menus.rs
+++ b/examples/persistent_menus.rs
@@ -10,9 +10,9 @@
 // menu; without quick completions, the menu closes once the line is empty.
 
 use reedline::{
-    default_emacs_keybindings, ColumnarMenu, Completer, DefaultPrompt, Emacs, KeyCode,
-    KeyModifiers, Keybindings, MenuBuilder, Reedline, ReedlineEvent, ReedlineMenu, Signal, Span,
-    Suggestion,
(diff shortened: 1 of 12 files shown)


kronberger-droid:
Nice!
I hope I will get some time this week to take a closer look.

philocalyst:
Responded to comments and made more efficent with stronger use of Arced suggestions !

fdncred:
ok, let's continue to push forward @philocalyst to get these completions in good order. appreciate your help.

PR #1125, opened 2026-07-20
fix: dont expand abbr eagerly on submit

### LINKED ISSUE #1124: Commands that start with an abbreviation can be expanded unexpectedly
| Type | Version |
| --- | --- |
| Platform | Linux (NixOS 26.05) |
| Terminal | wezterm 0-unstable-2026-07-07 |
| Nushell | 0.114.1 |

I define an abbreviation of `t = "tmux"`. Depending on the position of my cursor, commands beginning with `t` are transformed.

1. `$ tree`
2. Press <kbd>Enter</kbd>
3. ``Command `tmuxree` not found`` 

## Steps to reproduce

```nu
# nu --no-config-file
$env.config.abbreviations.t = 'example'

tree # Move cursor immediately after `t`
```
```
~> exampleree
Error: nu::shell::external_command
```

## Factors Involved

For me, this happens often for two reasons:

- I use a lot of single-letter abbreviations (`t`, `p`, `g`, `h`, `a`). They're better on `history` compared to aliases.
- I use `$env.config.edit_mode = "vi"`.

The vim edit mode means when my cursor is at the beginning of the line, it's interpreted as following the first character. That seems like a bug on its own, but... 🤷 

## Concrete Example

Discovered while editing this command in vim mode:
```nu
TREEFMT_CI=true treefmt
```

Went to the beginning of the line, deleted the var (`df `), <kbd>Enter</kbd>. ``Command `tmuxreefmt` not found``.

## Related(?) Changes

It's tangential to another abbreviation bugfix from the last release: https://github.com/nushell/reedline/pull/1111 

AIUI fixing the bug altered how abbreviations are detected and expanded without a trailing space.

Btw @MerlijnW70 thank you so much for tackling that. Completion from history has been so much nicer since nushell 0.114 ❤️

casedami:
## Summary 

fixes #1124 

### Before 

when `enter` was pressed, any characters before the cursor were treated as an separate word when the cursor was in the middle of a word. then a sequence like `gs|gs<enter>` where the cursor position is shown with `|` would result in an expansion of the first `gs` -> ie `git statusgs`

### After 

this behavior is no longer observed

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1950,6 +1950,17 @@ impl Reedline {
             // The first char in the buffer is a space or there are consecutive spaces
             return None;
         }
+
+        if submitted
+            && buffer[word_end..]
+                .chars()
+                .next()
+                .map_or(false, |ch| !ch.is_whitespace())
+        {
+            // The cursor is in the middle of a word, e.g. "hello|world"
+            return None;
+        }
+
         if !self.highlighter.should_expand_abbr(
             buffer,
             word_start,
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks!

PR #1128, opened 2026-07-22
refactor(prompt)!: use nu_ansi_term::Color throughout Prompt Trait

kronberger-droid:
## Summary 

Three functions of the `Prompt` Trait returned `crossterm::Color` instead of the established `nu_ansi_term::Color`.
This PR changes this to match the convention of the repo.

In turn I had to change the `const DEFAULT_PROMPT_*_COLOR` values to match the default palette.
These will now also be exported such that nushell does not have to define it itself in nushell/nushell#18686.
Also I switched the `crossterm::Color` reexport to `nu_ansi_term::Color`.

This change should be Behavior Preserving, but introduces a...

### Breaking Change!:

The four Prompt Trait functions `get_*_color(&self)` now all return `nu_ansi_term::Color` instead of `crossterm::Color`.
This will force every downstream consumer to change its Trait Implementation!

## Additional notes 

added tests to pin prompt color output across the switch
related #1046, nushell/nushell#18686

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -259,7 +259,8 @@ pub use history::{
 mod prompt;
 pub use prompt::{
     DefaultPrompt, DefaultPromptSegment, Prompt, PromptEditMode, PromptEditModeDiscriminants,
-    PromptHistorySearch, PromptHistorySearchStatus, PromptViMode,
+    PromptHistorySearch, PromptHistorySearchStatus, PromptViMode, DEFAULT_INDICATOR_COLOR,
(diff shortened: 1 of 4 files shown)


kronberger-droid:
Since nushell/nushell#18686 is not merged now we can just update the reedline reference and don't have to change anything else, since nushell used the default impls from reedline before.
I will then adapt the PR to reflect the changes introduced here.

fdncred:
ok, we can land this when you're ready.

PR #1133, opened 2026-07-23
fix(painter): don't re-use a bottom-flush prompt anchor after a scroll

### LINKED ISSUE #1130: Wrong prompt positioning after fzf menus when there's little room left and content get moved upwards
**Platform** e.g. macOS, Windows 10, ...
**Terminal software** e.g. gnome-terminal, Apple Terminal.app, ... 

It's quite difficult to describe the repro steps, I'll just try to show it in a recording.
PS. I'm 80% sure it's a regression introduced after 0.114

## Screenshots/Screencaptures

https://github.com/user-attachments/assets/77193591-eae3-46b1-9c17-fba76bb295cd

kronberger-droid:
## Summary 

When an external completer scrolls the terminal (`fzf --height` with little room below the prompt), the prompt landed in the wrong palce after the menu closed, since nushell/reedline#1090 trusts a `Verified` anchor without re-querying and the completer path never marked it stale.

Two commits:

#### re-anchor the prompt to the cursor after an external scroll
- After a menu `Activate`/`Edit`/page re-queries the completer, mark the anchor stale so the next paint re-verifies it.
- On drift, re-anchor to the queried cursor row instead of homing to row 0, which yanked the prompt to the top.
- The stale flag is gated to completer-querying events, so navigation and idle repaints keep nushell/reedline#1090's query-free path.
- Fixes nushell/reedline#1130.
But might be redundant depending on what comes out of the discussion in nushell/nushell#18771.

#### don't re-use a bottom-flush prompt anchor after a scroll
- A separate latent bug on the suspend/resume path (`ExecuteHostCommand`):
  `select_prompt_row` re-used a prompt range flush against the screen bottom and drew over scrolled-up output.
- Not a regression (identical on 0.48)

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -146,6 +146,12 @@ pub struct Reedline {
     history_cursor_on_excluded: bool,
     input_mode: InputMode,
 
+    // Set when a menu event that re-queries the completer (Activate/Edit) is
+    // dispatched. The host completer can shell out and scroll the terminal, so
+    // the next paint re-verifies the prompt anchor. Consumed in `repaint`.
+    // See #1130.
+    completer_may_have_scrolled: bool,
+
     // State of the painter after a `ReedlineEvent::ExecuteHostCommand` was requested, used after
     // execution to decide if we can re-use the previous prompt or paint a new one.
     suspended_state: Option<PainterSuspendedState>,
@@ -303,6 +309,7 @@ impl Reedline {
             history_excluded_item: None,
             history_cursor_on_excluded: false,
             input_mode: InputMode::Regular,
+            completer_may_have_scrolled: false,
             suspended_state: None,
             last_render_snapshot: None,
             painter,
@@ -1270,6 +1277,7 @@ impl Reedline {
                 if self.active_menu().is_none() {
                     if let Some(menu) = self.menus.iter_mut().find(|menu| menu.name() == name) {
                         menu.menu_event(MenuEvent::Activate(self.quick_completions));
+                        self.completer_may_have_scrolled = true;
 
                         if self.quick_completions && menu.can_quick_complete() {
                             menu.update_values(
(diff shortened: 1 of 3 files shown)


kronberger-droid:
@blindFS Can you check out if this fixes nushell/reedline#1130

blindFS:
@kronberger-droid Thanks a lot for the quick fix, however it doesn't seem to fix my issue. I cherry-picked the change on an older pair of revs of nushell 8774a88a1 and reedline 0977032 (sadly current main branches cause me bigger troubles), and it acts exactly the same.

blindFS:
Here're the minimal reproducible steps according to my test, but you'll have to test them on older revs as well.

1. `$env.config.completions.external.completer = {|| fzf --no-tmux --height 40%; null}`
2. Make the room below current prompt row less than 40%
3. Press `git <tab><ESC>`

kronberger-droid:
thanks! I will check it out. This was just a guess since i could not fully understand what was happening in the video.

kronberger-droid:
@blindFS
Now it should be fixed.
With a small perf regression since now if a menu is open we again have a cursor pos query every repaint.
But I will see if we could still reduce it.
Important now is that it fixes your issue.

blindFS:
@kronberger-droid Thanks again! It works much better now. Nitpick: it clears the terminal on menu exit, better to keep it on the same line. I think it worked that way before, but I'm not very sure.

kronberger-droid:
@blindFS
Right now it should either attach back to where it was or if the menu is to big and its shifted up, it should stay where it was shifted. 
are you experiencing something else?
Or do you want it to land in the place it was before opening the menu in the second case?
First case should meet your expectations right?

blindFS:
@kronberger-droid 


https://github.com/user-attachments/assets/2be07696-10b8-4361-aaaf-bf24ba37d9e1

Probably different to your description.

kronberger-droid:
@blindFS 
Hmm weird, are you sure you have built it with the right rev of reedline?
This was the state for me like one iteration before and i did not like it so i changed it. but i tested using a reedline example not nushell. so i will test this in nu and see if it changes something.

blindFS:
> @blindFS Hmm weird, are you sure you have built it with the right rev of reedline? This was the state for me like one iteration before and i did not like it so i changed it. but i tested using a reedline example not nushell. so i will test this in nu and see if it changes something.

Oops, my bad, I missed a crucial part of your change, it works perfectly now. Thanks a lot!

PR #1136, opened 2026-07-29
Stronger contracts around completions

philocalyst:
## Summary

Async completors BEFORE TODAY... gave menus no real way to prove work was in progress. This instead adds a pleasant indicator that replaces the menu marker while a completion runs, plus a keyed message!

On the way:
- CompletionResult carries no an Optional Partial
- `ColumnarMenu`/`IdeMenu` track a `CompletionOrigin` (buffer + insertion point) so a `Pending` result computed against a stale buffer can't splice its span into the current one.
- `unicase` moved from a runtime dependency to a dev-dependency! 

Within the public API:
`WorkingIndicator`, `WorkingPhase`, `CompletionProgress`, `Partial`, and new `Menu`/`MenuBuilder`/`MenuSettings` methods (`with_working_marker`, `with_working_indicator`, `working_indicator`, `progress`, `working_phase`, `working_message`). `Menu::set_active`/`Menu::clear_input` are new required methods (no default body), so this is source-breaking for any external `Menu` implementors.

### Before
Async completers had no way to signal "still working" 

### After
A menu with a `WorkingIndicator` configured animates its marker while a completion is in flight, and shows a message afta enough time.

### DIFF
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -34,6 +34,35 @@ impl Span {
     }
 }
 
+/// Buffer text and cursor position the completion was computed against.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct CompletionOrigin {
+    pub(crate) buffer: String,
+    pub(crate) insertion_point: usize,
+}
+
+impl CompletionOrigin {
+    /// Buffer/cursor for stale results.
+    pub fn new(buffer: impl Into<String>, insertion_point: usize) -> Self {
+        Self {
+            buffer: buffer.into(),
+            insertion_point,
+        }
+    }
+}
+
+/// Longest common prefix extension computed by the completer.
+///
+/// When present, reedline splices `insert` over `span` verbatim.
+/// Absent — reedline uses its own LCP derivation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Partial {
+    /// Buffer range to replace.
+    pub span: Span,
+    /// Text to splice in.
+    pub insert: String,
+}
+
 /// The outcome of a [`Completer::complete`] request.
 ///
 /// A synchronous completer only ever produces [`Fresh`](Self::Fresh). An
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -63,54 +63,86 @@ pub struct Partial {
     pub insert: String,
 }
 
-/// The outcome of a [`Completer::complete`] request.
-///
-/// A synchronous completer only ever produces [`Fresh`](Self::Fresh). An
-/// asynchronous completer that computes in the background reports its progress
-/// through this type.
+/// Outcome of a [`Completer::complete`] request.
 #[derive(Debug, Clone)]
 pub enum CompletionResult {
-    /// Final, authoritative results. No further computation is in flight.
-    Fresh(Suggestions),
-    /// Best-effort results to show in the moment; a fresh computation is still running and
-    /// will replace these once it finishes.
-    Stale(Suggestions),
-    /// No results are available yet; a computation is spinning in the background.
+    /// Authoritative results.
+    Fresh {
+        /// The suggestions.
+        suggestions: Suggestions,
+        /// Partial completion prefix.
+        partial: Option<Partial>,
+    },
+    /// Best-effort results while computation is in flight.
+    Stale {
+        /// The suggestions.
+        suggestions: Suggestions,
+        /// Buffer/cursor these were computed against.
+        origin: CompletionOrigin,
+        /// Partial completion prefix.
+        partial: Option<Partial>,
+    },
+    /// No results yet; computation in progress.
     Pending,
 }
 
 impl CompletionResult {
-    /// Wrap authoritative results.
+    /// Wrap authoritative results with no partial.
     pub fn fresh(suggestions: impl Into<Suggestions>) -> Self {
-        CompletionResult::Fresh(suggestions.into())
+        CompletionResult::Fresh {
+            suggestions: suggestions.into(),
+            partial: None,
+        }
     }
 
-    /// Best-effort fallback while an authoritative result is still computing:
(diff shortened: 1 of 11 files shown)


kronberger-droid:
Some things I found while exploring the changes with Claude:

**Partial completion is dead for everything but nushell.**
`common_prefix` returns `false` on `partial: None`, and `DefaultCompleter` (`default.rs:124`) and `HistoryCompleter` (`history.rs:37`) still call bare `CompletionResult::fresh(...)`.
`examples/demo.rs:88` has `with_partial_completions(true)` on a `DefaultCompleter`, so `ab` + Tab gives `abaaa` on `main` and stays `ab` here.
`menu_functions.rs` still promises "the built-in derivation kicks in".
That logic now survives only as `test_helpers::fake_partial`, and the `test_find_common_string` cases went out with it, nushell/nushell#16765 included.

**The `Stale` gate is weaker than its docs claim.**
`computed_for` is stamped at adoption time, not from the buffer the completer computed against, which for `Stale` is an older one.
Hand `stale_pending_span_is_not_spliced` a `Stale` result with `span: 5..10` instead of `PendingCompleter` and you get `open CODE_OF_CONDUCT.mdOF_CONDUCT.md | from csv --sep`.
Not reachable through nushell/nushell#18761, since `narrowed_fallback` rebases every span to `Span::new(reference_span.start, query.cursor())` first.
But that puts the invariant in the completer, which is the opposite of what the `CompletionDisplay` doc claims ("unrepresentable rather than merely guarded-against").
Either `Stale` reaches `accept`/`common_prefix` only behind something reedline can check, or `CompletionResult::Stale` spells out that spans must describe the buffer it was handed.

Smaller:
- An async Tab reserves a blank row for 2s before `searching...` shows, since `working_message` is empty until `patience` and `get_rows` returns 1 where it returned 0.
- The default frames `· ∙ • ∙` are all East Asian Ambiguous. Do they hold one column in a CJK-width terminal?
- The description is short a few breaks: `Fresh`/`Stale` gained a field, `find_common_string` and `can_partially_complete` removed, `CompletionDisplay::values` private, `CompletionDisplay::new` changed signature.
- The `Menu` trait refactor is unrelated to completions but is the only source-breaking part for external `Menu` implementors. Could it go out separately, since follow-ups are coming anyways?

kronberger-droid:
@fdncred

Worth checking how close nushell/nushell#18761 actually is before we rush this one through.
It's still a draft and it builds on nushell/nushell#18660, which is also a draft.

The duplication in nushell/nushell#18737 is fixed by the `CompletionSite` work in #18761, not by anything here.
So landing this early doesn't get users a fix, it parks a dependency ahead of a WIP.

@philocalyst
Does the span-math half of #18761 lean on these reedline changes at all?
If `CompletionSite` can go out on its own, we'd fix #18737 now and take the cache work and this PR at whatever pace we need.

fdncred:
ya, I was thinking we should point nushell's PR 18761 reedline to this PR/branch so we can test it in nushell.

I'm also confused as to whether 18660 is built into 18761 or we need all 3 of these PR landed (eventually) in a very specific order.

Thanks for the review.

philocalyst:
> If CompletionSite can go out on its own, we'd fix #18737 now and take the cache work and this PR at whatever pace we need.

I'll rework to make this possible.

To clarify: Async prompts are built on top of the completer rework PR. I did this out of convenience, I don't think there's a _direct_ link. If it's not pressing to have a clear split, I'd rather work through those two PRs one by one, as they're both large and fundamental.

kronberger-droid:
Well I would say most importantly we/you @philocalyst should focus on making completers stable for the user and then build on the stable base.
Right now I am not sure how much of the two nushell PRs and this one is for stability over features.

fdncred:
ya, I agree. Let's focus on stability of what we have in these PRs. We can refactor later when it's not such an emergency as it is right now. It's critical that we get the completion bugs fixed soon since our release is on August 15th and we have code freeze around August 13th.

fdncred:
Thanks for chatting with us today @philocalyst. I think you said you were going to split this up somehow. If you do that please try to include fixes for the issues martin and copilot mentioned above.

philocalyst:
Okay, incorporated your review comments @kronberger-droid , ditto for the automated ones. A second PR will come up tomorrow on the visual indicator. This covers everything that is not that.

philocalyst:
I do want to add that the nushell PR is relatively intertangled with even the visual indicator changes, as there I define usage around how progress is communicated, just something to keep in mind.

kronberger-droid:
@philocalyst
Do you think it makes sense to disentangle it a bit since right now I am not fully in the picture of what does what.

Also do you think we will be able to land all of your changes some time before the release on 15th.
Since i think it would be good to have a somewhat stable version before the release with time to spare to get some testing in.

Thanks for your great work!

fdncred:
> Also do you think we will be able to land all of your changes some time before the release on 15th.

My preference is to land PRs sooner, like this week if possible. There's always unexpected edge cases that come up and we have to land PRs and get to finding those ASAP, IMO. We need stability more than anything else.

philocalyst:
@kronberger-droid It COULD be splt up, but with trouble, as most of this is so tightly coupled. "What is going on here" is essentially nothing -- just moving the implicit behavior into types.

kronberger-droid:
Okay, if you think you can land them until next week it doesn't matter.

Haha thanks for the clarification. I was worried we do a lot of feature work now. So thats not the case?

philocalyst:
No, that I did split up into another PR.

fdncred:
Are we ready to land this or do we need more time?

philocalyst:
I'm confident in it. Up to you? @kronberger-droid

kronberger-droid:
On reedline for sure!

But I can't really tell for the nushell ones, these are much larger.

fdncred:
ok, let's go then! thanks!

PR #1137, opened 2026-07-30
visual indication of completion

philocalyst:
## Summary 

This adds really cute completion indicators to reedline! Effectively a series of dots, pulsating in/out. Included with this is a clearer APIs around completion status, through workingphase and completion progress. 

### Before 

- Completion rendered as a static marker.
- Background work was tracked a boolean.

### After 

-  Completions are communicated as a dynamic state. This opens up doors for further extensibility (like 1-100 for bars and things)
- Added apis for communicating when completion values are stale or un-stale, through working_indicator and working_marker.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -928,6 +928,9 @@ impl Reedline {
 
         self.repaint(prompt)?;
 
+        // Working phase of active menu at last paint.
+        let mut painted_working_phase = None;
+
         loop {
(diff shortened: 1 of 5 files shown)


fdncred:
I'm fine with this as long as there's a way to turn it off/disable it. Some people don't like things like this.

philocalyst:
Just closing this out because I don't mind it as-is, happy to consult anyone who wants to pick it up :)

PR #1138, opened 2026-07-31
feat(helix)!: add helix edit mode behind a feature flag

kronberger-droid:
## Summary

Adds a `Helix` edit mode behind the non-default `helix` feature.
Motions carry the selection and verbs act on it, rather than vi's operator-pending model.
The mode is a state machine over a `Verb` vocabulary that lowers onto existing `EditCommand`s.

Default builds are untouched: every helix-specific item is `#[cfg(feature = "helix")]`.

### Keybindings

Descriptions are upstream's own, from https://docs.helix-editor.com/keymap.html

| Key | Description |
| --- | --- |
| `h` `j` `k` `l` | Move left / down / up / right |
| `w` `b` `e` | Move next word start / previous word start / next word end |
| `W` `B` `E` | Move next WORD start / previous WORD start / next WORD end |
| `f` `F` | Find next char / Find previous char |
| `t` `T` | Find till next char / Find till previous char |
| `%` | Select entire file |
| `x` | Select current line, if already selected, extend to next line |
| `v` | Enter select (extend) mode |
| `i` `a` | Insert before selection / Insert after selection (append) |
| `I` `A` | Insert at the start of the line / Insert at the end of the line |
| `o` `O` | Open new line below selection / Open new line above selection |
| `d` | Delete selection |
| `Alt-d` | Delete selection, without yanking |
| `c` | Change selection (delete and enter insert mode) |
| `y` | Yank selection |
| `p` `P` | Paste after selection / Paste before selection |
| `r` | Replace with a character |
| `~` | Switch case of the selected text |
| `` ` `` | Set the selected text to lower case |
| ``Alt-` `` | Set the selected text to upper case |
| `u` `U` | Undo change / Redo change |
| `ge` | Go to the end of the file |
| `gh` `gl` | Go to the start of the line / Go to the end of the line |
| `gs` | Go to first non-whitespace character of the line |

Counts prefix any of them.
`j`/`k` additionally walk history at the buffer edge, since a prompt has nowhere above its first line to go.

### API Changes

Three gated additions: `MotionTarget::LineStartNonBlank`, `EditCommand::SelectLine`, `EditCommand::EraseSelection`.
`EditCommand` is `#[non_exhaustive]` but its generated discriminants are not, hence the `!`.

### Out of scope

`;` `J` `>` `<` `G` `X`, and the `/` `n` `N` search family.


src/prompt/base.rs
```diff
@@ -65,6 +65,9 @@ pub enum PromptEditMode {
     /// A vi-specific mode
     Vi(PromptViMode),
 
+    /// A helix/Kakoune like mode
+    Helix(PromptHelixMode),
```
Juhan280, line 70: I believe this should be feature gate for now

src/prompt/base.rs
```diff
@@ -173,6 +212,8 @@ impl IntoDiscriminant for PromptEditMode {
             // keybindings, differing only in selection geometry.
             Self::Vi(Vi::Normal | Vi::Visual) => Self::Discriminant::ViNormal,
             Self::Vi(Vi::Insert) => Self::Discriminant::ViInsert,
+            Self::Helix(Helix::Normal | Helix::Select) => Self::Discriminant::HelixNormal,
```
Juhan280: I think we should have separate discriminat for helix select mode, and same for vi visual. I believe this code was added before we had proper vi visual mode.

Unless we still don't have proper keybind support for visual (or hx select) mode yet

src/prompt/base.rs
kronberger-droid: Yeah i think I have to go over it in more detail, i think thats not the only one i forgot to feature gate.

src/prompt/base.rs
kronberger-droid: Agreed.

PR #1139, opened 2026-08-01
Take the external-print anchor from the terminal, not a count

### LINKED ISSUE #1005: mpsc receive causing Printer to perform screen clearing
I have no idea why it happens, but the Printer makes me want to kill myself- I made a shell that talks to multiple endpoints from dedicated threads; they are writing to an mpsc endpoint and the rx is read by the Printer. The printer will often/sometimes start "clearing the terminal screen"- making the entire runtime useless. 

I don't understand what is happening to begin debugging the problem- it's so senstive to change- it works for weeks and then randomly it stops working..

james-kassabian:
## Summary

`print_external_message` queues its lines without flushing, then records the new prompt anchor by counting one row per message. Nothing has been flushed at that point, so the terminal's cursor has not moved — the anchor names a row the terminal has not reached. Stored as `Stale`, the next paint re-verifies it against the real, still-earlier cursor, reads the gap as the prompt having scrolled off the top, and re-anchors the only way it can: by printing a screenful of newlines. That is the wipe in #1005.

I ran into it in a shell I'm writing, where job notifications arrive in batches, and traced it with an instrumented painter before changing anything.

No public API change, and no observable behaviour change beyond the wipe no longer happening.

### Before

Two or more messages sent while the prompt sits near the top of the screen wipe the display. A single message never triggers it, and a full screen is immune — both follow from the arithmetic (`pos.1 + 1 < row`, and `row` being clamped by `.min(max_row)`), worked through in #1005.

### After

The anchor is taken from `cursor::position()` after a flush, so it matches the terminal and the drift check has nothing to misread.

## Additional notes

- The flush is once per batch rather than per message, so the flicker the retained comment warns about is unaffected.
- It also drops the assumption that each message occupies exactly one row, which the replaced comment already called unsafe for messages that wrap or carry control sequences.
- 1209 tests pass. `painting::painter::tests::crossterm_defaults_were_the_bright_palette_entries` fails identically with and without this change on clean `main` here (Windows), so it is pre-existing.

Fixes #1005.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -1180,12 +1180,24 @@ impl Painter {
             self.stdout.queue(Print(line))?.queue(Print("\r\n"))?;
             row = row.saturating_add(1).min(max_row);
         }
-        // Track the row by counting `\r\n`s — matches reedline's
-        // historical behavior. The drift check is one-sided so a
-        // message that secretly scrolls more rows than we counted
-        // (embedded `\n`, certain CSI sequences) can still incorrectly
-        // anchor.
-        self.prompt_start_row = PromptStartRow::Stale(row);
+        // The lines above are only *queued*, so the terminal's cursor has not
+        // moved yet: a row counted forward from `starting_row` names a position
+        // the terminal has not reached. Recorded as `Stale`, the next paint
+        // re-verifies it against the real, still-earlier cursor, reads that as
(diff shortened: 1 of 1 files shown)


james-kassabian:
Reworked the description onto the PR template — summary with motivation, before/after, and notes. Thanks for the steer, and no rush on the review.

PR #1144, opened 2026-08-07
fix(engine): decide menu completions once the completer answers

### LINKED ISSUE #1142: Single completion no longer auto-completing - Regression from 0.114.1
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

After nushell/nushell#18334, hitting tab no longer automatically completes when there is only one possible completion.  This has been an integral feature for ages.

I was hoping that nushell/nushell#18761 was going to fix this, but it doesn't.

### How to reproduce

In the nushell repo.

```nushell
cd cr<tab>
```

Displays the `crates` subdirectory in the completion *menu* rather than auto-completing as it used to.

### Expected behavior

As with 0.114.1 and past releases, `<tab>` should auto-complete when there is only a single possible completion matching the current commandline.

### Configuration

| key                  | value                                                                                                                                         |
| -------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| version              | 0.114.2                                                                                                                                       |
| major                | 0                                                                                                                                             |
| minor                | 114                                                                                                                                           |
| patch                | 2                                                                                                                                             |
| branch               |                                                                                                                                               |
| commit_hash          | 8774a88a1f81761bbd71ed63edf3fdc5f72f10ad                                                                                                      |
| build_os             | linux-x86_64                                                                                                                                  |
| build_target         | x86_64-unknown-linux-gnu                                                                                                                      |
| rust_version         | rustc 1.95.0 (59807616e 2026-04-14)                                                                                                           |
| rust_channel         | 1.95.0-x86_64-unknown-linux-gnu                                                                                                               |
| cargo_version        | cargo 1.95.0 (f2d3ce0bd 2026-03-21)                                                                                                           |
| build_time           | 2026-07-26 14:29:46 -04:00                                                                                                                    |
| build_rust_channel   | release                                                                                                                                       |
| allocator            | standard                                                                                                                                      |
| features             | default, lsp, mcp, network, plugin, rustls-tls, sqlite, trash-support                                                                         |
| installed_plugins    | formats 0.106.2, gstat 0.113.2, polars 0.106.2, query 0.106.2                                                                                 |
| experimental_options | example=false, dc-glob=true, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=true, native-clip=true, cell-path-types=true |

kronberger-droid:
## Summary 

Since #1093 a completer may answer with `Pending` and compute in the background. Activating a menu inspects its values immediately, so the quick completion and the partial completion both read an empty menu and neither fires, leaving a menu open where a lone suggestion used to be accepted outright.

Both decisions are now replayed when the results land, guarded by a snapshot of the line taken as the menu opened, since a completion applied after the user has typed on is worse than one that never fires.

## Additional notes 

Fixes #1142

### DIFF
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -49,6 +49,14 @@ impl CompletionOrigin {
             insertion_point,
         }
     }
+
+    /// Whether `buffer` and `insertion_point` are still exactly what this was
+    /// stamped for. Anything else means the line moved on and results computed
+    /// against this origin no longer describe it.
+    pub fn matches(&self, buffer: &str, insertion_point: usize) -> bool {
+        // Cursor first: it rules out most drift without comparing the line.
+        self.insertion_point == insertion_point && self.buffer == buffer
+    }
 }
 
 /// Longest common prefix extension computed by the completer.
--- a/src/completion/base.rs
+++ b/src/completion/base.rs
@@ -152,6 +152,12 @@ impl CompletionResult {
     pub fn is_pending(&self) -> bool {
         matches!(self, CompletionResult::Pending)
     }
+
+    /// Whether a later result may still supersede this one: either nothing has
+    /// arrived yet, or what arrived was computed against a different line.
+    pub fn is_provisional(&self) -> bool {
+        !matches!(self, CompletionResult::Fresh { .. })
+    }
 }
 
 /// Vitality of a completer's background work, grabbed by the engine once per
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1102,6 +1102,13 @@ impl Reedline {
             return Ok(());
         };
 
+        // The request that just finished was host code, which may have scrolled the
+        // terminal while it ran in the background — after the keystroke that
+        // dispatched it had already re-verified the anchor. `update_values` below
+        // can also dispatch another request for a line that moved on. Either way
+        // the repaint at the end of this settle must not trust the cached row.
+        invalidate_anchor_if_host_completer_runs(&self.menus[menu_index], &mut self.painter);
+
         let menu = &mut self.menus[menu_index];
         // The request this menu was waiting on finished, so repopulate it.
         menu.update_values(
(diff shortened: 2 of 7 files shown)


kronberger-droid:
@NotTheDr01ds Can you check out if this solves your problem?

NotTheDr01ds:
Thanks! I'll hopefully be able to give it a shot later today.

NotTheDr01ds:
Okay, that's working, but I surfaced two new issues with the new completion system now that that's out of the way.  I'll open separate issues on that.  Not sure whether to put them in Reedline or Nushell, though:

* **Long** delay completing short completions:  For instance, simply hit <kbd>Tab</kbd> on an empty commandline and it takes 8 seconds to pop up the list.  This is *probably* going to only be seen on WSL, but it's a serious step back in performance.
* First completion of any pattern creates a flash of the commandline that didn't exist before.

NotTheDr01ds:
I've entered these as https://github.com/nushell/nushell/issues/18799 and https://github.com/nushell/nushell/issues/18800

Thanks for your help!

NotTheDr01ds:
Ah, the flashing/redrawing the commandline seems to be specific to this PR, since it wasn't possible without removing the menu in this case.

See the (now closed) https://github.com/nushell/nushell/issues/18800 - Closed since it only exists with this PR.

Do you think it might be possible to get rid of the redraw?

philocalyst:
The redraw may be because of the LRU cache. Try removing that and see if it replicates? I think it will make more sense to release this without the cache on, because of unpredictability

kronberger-droid:
Should be gone now. @NotTheDr01ds can you check it out again?

kronberger-droid:
Might need still some changes tho. Since some of the stuff could be derived.

NotTheDr01ds:
Thanks! Confirmed that the redraw is gone.

PR #1151, opened 2026-08-17
fix(history): query the start_timestamp column in time-range searches

kronberger-droid:
## Summary

`SqliteBackedHistory::construct_query` filtered on `timestamp_start`,
a column that has never existed in the schema (it is `start_timestamp`, `sqlite_backed.rs::308`).
Nothing in reedline or nushell sets `SearchQuery::start_time` / `end_time` today, thus nobody hit it,
but any caller that does gets a panic from the unwrapped `prepare` in `count` and `search`.

Two smaller things rode along:
`prepare` errors now go through `map_err(map_sqlite_err)?` like `search_with_extra` already does,
and the session_id index gets its own name (`idx_history_session`) instead of colliding with `idx_history_cmd`,
which made `create index if not exists` skip it silently.

No public API change.

### Before

`search`/`count` with a time bound panicked with `no such column: timestamp_start`.
Existing databases had no index on `session_id`.

### After

Time-bounded queries return the expected rows.
The `session_id` index is created on next open, since the schema block runs `if not exists` every time.

### DIFF
--- a/src/history/sqlite_backed.rs
+++ b/src/history/sqlite_backed.rs
@@ -150,7 +150,7 @@ impl History for SqliteBackedHistory {
         let result: i64 = self
             .db
             .prepare(&query)
-            .unwrap()
+            .map_err(map_sqlite_err)?
             .query_row(&params_borrow[..], |r| r.get(0))
             .map_err(map_sqlite_err)?;
         Ok(result)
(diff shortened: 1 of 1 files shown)


fdncred:
weird. good catch. 

i also would eventually like to see no unwraps in reedline, except for tests. seems like there was quite a few last time i checked. but that's for another pr, or 2.

kronberger-droid:
Thats the goal, I want to have the panic surface as close to zero as possible.

PR #1156, opened 2026-08-17
fix(painter): keep the cursor position query off the tty in tests

kronberger-droid:
## Summary

`Painter` already routes all *writes* through `W` so tests don't hit the terminal, since crossterm bypasses libtest's capture.
The *read* side slipped through: four `cursor::position()` calls went straight to crossterm.
Any test that painted with a stale anchor (every deferred-completion test does, via `invalidate_anchor_if_host_completer_runs`) sent `ESC[6n` to the real tty and waited on crossterm's timeout, serialized behind its global reader lock.
Locally that shows as `a_lone_late_suggestion_is_accepted_rather_than_spliced has been running for over 60 seconds` until a keypress completes the cooked-mode line; on CI without a tty the read loop never returns.

`W` gets a `cursor_position()` that delegates for `Terminal` and returns an error for the test writers, which is the branch those paints already took after the timeout, thus no test semantics moved.
The one-off `#[cfg(not(test))]` around the resize query is gone with it.

No public API change, no behavior change on the terminal.

## Additional notes

Full `cargo test --lib` with stdin closed: 1487 passed in 0.02s, no keypress needed.
Follow-up: `(column, row)` and the other painter tuples could use a small `CursorPos`-style type.

### DIFF
--- a/src/painting/painter.rs
+++ b/src/painting/painter.rs
@@ -152,6 +152,21 @@ impl Write for W {
     }
 }
 
+impl W {
+    /// Where the terminal's cursor is, as `(column, row)`.
+    ///
+    /// Only the real terminal can answer, but test writers return an error so
+    /// paint paths take their "no answer" branch instead of waiting on a tty
+    /// that will never reply.
+    pub(crate) fn cursor_position(&self) -> Result<(u16, u16)> {
+        match self {
+            W::Terminal(_) => cursor::position(),
+            #[cfg(test)]
+            W::Sink(_) | W::Capture(_) => Err(std::io::Error::other("no terminal attached")),
+        }
+    }
+}
+
 #[derive(Debug, PartialEq, Eq)]
 pub struct PainterSuspendedState {
     previous_prompt_rows_range: RangeInclusive<u16>,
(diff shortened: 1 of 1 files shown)


fdncred:
Is this the cause to the ESC[6n we see sometimes? We have people report these not knowing what it is because the terminal/nushell is reporting the cursor position.

kronberger-droid:
Well I don't know, my starting point was a blocking test which was using the sink, waiting for a cursor position.

Do we have an issue on the ESC[6n then i check it out.

fdncred:
We probably do in nushell but I doubt it contains the string ESC[6n. I think this is one of them https://github.com/nushell/nushell/issues/15312
another maybe https://github.com/nushell/nushell/issues/17473
another https://github.com/nushell/nushell/issues/11950
https://github.com/nushell/nushell/issues/13570

kronberger-droid:
Seems that those are unrelated to this one, since these are runtime behaviors.
In this one we essentially handle only in the scope of tests.

Claude suggested it these might be due to cursor::position() calls not treating "no answer" as a normal outcome. But I have not looked to much into it.
Definitely worth a follow-up, and should also be partially addressed since we reduced our cursor::position() queries a lot: #1090

Just wanted to clearify this, before I merge all my non-breaking PRs, so we can have them in nushell as soon as possible.

PR #1158, opened 2026-08-17
chore: drop the idle_callback feature flag

kronberger-droid:
## Summary

`idle_callback` gated an `Option<Box<dyn FnMut>>` field on `Reedline`, two `is_some` checks and the `with_idle_callback` builder.
No dependency behind it, not in the CI matrix (only `--all-features` touched it), no example, no nushell use.
A flag that toggles a public method is a papercut for docs.rs readers and buys nothing measurable, thus the gate is gone and `with_idle_callback` is always available.

Public API: `with_idle_callback` no longer needs a feature. Passing `--features idle_callback` now yields cargo's unknown-feature warning and nothing else.
No behavior change: polling only starts once a callback is set, same as before.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,6 @@ tempfile = "3.3.0"
 default = ["helix"]
 bashisms = []
 external_printer = ["crossbeam"]
-idle_callback = []
 helix = []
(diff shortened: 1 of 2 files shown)


fdncred:
I'm definitely good with less features.

PR #1188, opened 2026-08-24
feat(editor): Add helix match selection

glcraft:
## Summary

Add match actions from helix to reedline, and bind the new behaviors to the helix mode.

Keybindings:
- `mi<TextObject>` : select inside a text object
- `ma<TextObject>` : select around a text object
- `ms<TextObject>` : add the specified text object around the selection
- `md<TextObject>` : remove the nearest specified text object around the selection
- `mr<old TextObject><new TextObject>` : replace the nearest specified text object around the cursor with another text object

This includes the following new EditCommand: 
- `SelectTextObject`
- `AddTextObject`
- `RemoveTextObject`
- `ReplaceTextObject`

Moreover, brackets and quotes are more granular in TextObjectType by differentiating each text object.

## Additional notes

Redundant code has been removed, replaced by existing or new functions. In a nutshell, `<Cut/Copy><Inside/Around>Pair` had been replaced by `<Cut/Copy>TextObject`.

<details>
<summary>Detail here</summary>

Among that, the following EditCommand has been replaced:
- `EditCommand::CutInsidePair{start: char, end: char)` => `EditCommand::CutTextObject{scope: TextObjectScope::Inner, object_type: TextObjectType::XXX}`
- `EditCommand::CutAroundPair{start: char, end: char)` => `EditCommand::CutTextObject{scope: TextObjectScope::Around, object_type: TextObjectType::XXX}`
- `EditCommand::CopyInsidePair{start: char, end: char)` => `EditCommand::CopyTextObject{scope: TextObjectScope::Inner, object_type: TextObjectType::XXX}`
- `EditCommand::CopyAroundPair{start: char, end: char)` => `EditCommand::CopyTextObject{scope: TextObjectScope::Around, object_type: TextObjectType::XXX}`

In `core_editor::editor::Editor`, the following functions have been replaced: 
- `cut_inside_pair` => `cut_text_object`
- `cut_around_pair` => `cut_text_object`
- `copy_inside_pair` => `copy_text_object`
- `copy_around_pair` => `copy_text_object`

Vi mode implementation has been updated towards that.

All tests passes.
</details>

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -372,6 +372,7 @@ impl Editor {
             EditCommand::CutTextObject { text_object } => self.cut_text_object(*text_object),
             EditCommand::CopyTextObject { text_object } => self.copy_text_object(*text_object),
             EditCommand::AddTextObject { text_object } => self.add_text_object(*text_object),
+            EditCommand::RemoveTextObject { text_object } => self.remove_text_object(*text_object),
         }
         let leaves_selection = matches!(command.edit_type(), EditType::MoveCursor { select: true })
             || matches!(command, EditCommand::PasteAtSelectionEdge { .. })
@@ -389,9 +390,10 @@ impl Editor {
         self.commit_cursor();
 
         let new_undo_behavior = match (command, command.edit_type()) {
-            (EditCommand::AddTextObject { .. }, EditType::MoveCursor { .. }) => {
-                UndoBehavior::CreateUndoPoint
-            }
+            (
+                EditCommand::AddTextObject { .. } | EditCommand::RemoveTextObject { .. },
+                EditType::MoveCursor { .. },
+            ) => UndoBehavior::CreateUndoPoint,
             (_, EditType::MoveCursor { .. }) => UndoBehavior::MoveCursor,
             (EditCommand::InsertChar(c), EditType::EditText) => UndoBehavior::InsertCharacter(*c),
             (EditCommand::Delete, EditType::EditText) => {
@@ -1801,6 +1803,38 @@ impl Editor {
         self.line_buffer.set_cursor(cursor_lefted.flip());
         self.line_buffer.insert_char(pair.0);
         self.place(Cursor::new(cursor.anchor() + 1, cursor.head() + 1));
+    }
+
+    fn remove_text_object(&mut self, text_object: TextObjectType) {
+        if !matches!(
+            text_object,
+            TextObjectType::Brackets(_) | TextObjectType::Quotes(_)
+        ) {
+            return;
+        }
+        let cursor = self.line_buffer.cursor();
+        self.line_buffer.set_cursor(Cursor::point(cursor.head()));
+        // TODO: Ya un bug dans la fonction text_object_range,
+        // un range est trouvé lorsque le curseur se trouve hors du text object
+        // inclus brackets et quotes
+        let Some(range) = self.text_object_range(TextObject {
+            scope: TextObjectScope::Inner,
+            object_type: text_object,
+        }) else {
+            self.line_buffer.set_cursor(cursor);
+            return;
+        };
+        self.line_buffer.clear_range(range.end..range.end + 1);
+        self.line_buffer.clear_range(range.start - 1..range.start);
+        let new_cursor = Cursor::new(
+            if cursor.anchor() > range.start {
+                cursor.anchor() - 1
+            } else {
+                cursor.anchor()
+            },
+            cursor.head() - 1,
+        );
+        self.place(new_cursor);
     }
 
     fn select_text_object(&mut self, text_object: TextObject) {
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -7,7 +7,8 @@ use crate::core_editor::graphemes::{next_grapheme_boundary, prev_grapheme_bounda
 use crate::core_editor::resolve::resolve_selection;
 use crate::core_editor::{commit, line, operator_span, resolve_motion, RestPolicy};
 use crate::enums::{
-    EditType, TextObject, TextObjectBracket, TextObjectScope, TextObjectType, UndoBehavior,
+    EditType, TextObject, TextObjectBracket, TextObjectQuote, TextObjectScope, TextObjectType,
+    UndoBehavior,
 };
 use crate::prompt::PromptEditMode;
 use crate::{core_editor::get_local_clipboard, EditCommand};
@@ -1741,9 +1742,19 @@ impl Editor {
     ///
     /// If multiple quote types exist, returns the innermost pair that surrounds
     /// the cursor. Handles empty quotes as zero-length ranges inside quote.
-    fn quote_text_object_range(&self, text_object_scope: TextObjectScope) -> Option<Range<usize>> {
-        const QUOTE_PAIRS: &[(char, char)] = &[('"', '"'), ('\'', '\''), ('`', '`')];
-        self.matching_pair_group_text_object_range(text_object_scope, QUOTE_PAIRS)
+    fn quote_text_object_range(
+        &self,
+        text_object_scope: TextObjectScope,
+        quote_type: TextObjectQuote,
+    ) -> Option<Range<usize>> {
+        const QUOTE_PAIRS: &[(char, char)] = &[('\'', '\''), ('"', '"'), ('`', '`')];
+        let pairs = match quote_type {
+            TextObjectQuote::SingleQuote => &QUOTE_PAIRS[0..1],
+            TextObjectQuote::DoubleQuote => &QUOTE_PAIRS[1..2],
+            TextObjectQuote::Tick => &QUOTE_PAIRS[2..3],
+            TextObjectQuote::All => QUOTE_PAIRS,
+        };
+        self.matching_pair_group_text_object_range(text_object_scope, pairs)
     }
 
     /// Get the bounds for a text object operation
@@ -1754,7 +1765,9 @@ impl Editor {
             TextObjectType::Brackets(brackets_type) => {
                 self.bracket_text_object_range(text_object.scope, brackets_type)
             }
-            TextObjectType::Quote => self.quote_text_object_range(text_object.scope),
+            TextObjectType::Quotes(quote_type) => {
+                self.quote_text_object_range(text_object.scope, quote_type)
+            }
         }
     }
 
@@ -4043,18 +4056,18 @@ mod test {
     // Test text object jumping behavior in various scenarios
     // Cursor inside empty pairs should operate on current pair (cursor stays, nothing cut)
     #[case(r#"foo()bar"#, 4, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Brackets(TextObjectBracket::All) }, "foo()bar", 4, "")] // inside empty brackets
-    #[case(r#"foo""bar"#, 4, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quote }, "foo\"\"bar", 4, "")] // inside empty quotes
+    #[case(r#"foo""bar"#, 4, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quotes(TextObjectQuote::All) }, "foo\"\"bar", 4, "")] // inside empty quotes
     // Cursor outside pairs should jump to next pair (even if empty)
     #[case(r#"foo ()bar"#, 2, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Brackets(TextObjectBracket::All) }, "foo ()bar", 5, "")] // jump to empty brackets
-    #[case(r#"foo ""bar"#, 2, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quote }, "foo \"\"bar", 5, "")] // jump to empty quote
+    #[case(r#"foo ""bar"#, 2, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quotes(TextObjectQuote::All) }, "foo \"\"bar", 5, "")] // jump to empty quote
     #[case(r#"foo (content)bar"#, 2, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Brackets(TextObjectBracket::All) }, "foo ()bar", 5, "content")] // jump to non-empty brackets
-    #[case(r#"foo "content"bar"#, 2, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quote }, "foo \"\"bar", 5, "content")] // jump to non-empty quotes
+    #[case(r#"foo "content"bar"#, 2, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quotes(TextObjectQuote::All) }, "foo \"\"bar", 5, "content")] // jump to non-empty quotes
     // Cursor between pairs should jump to next pair
     #[case(r#"(first) (second)"#, 8, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Brackets(TextObjectBracket::All) }, "(first) ()", 9, "second")] // between brackets
-    #[case(r#""first" "second""#, 8, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quote }, "\"first\"\"second\"", 7, " ")] // between quotes
+    #[case(r#""first" "second""#, 8, TextObject { scope: TextObjectScope::Inner, object_type: TextObjectType::Quotes(TextObjectQuote::All) }, "\"first\"\"second\"", 7, " ")] // between quotes
     // Around scope should include the pair characters
     #[case(r#"foo (bar)"#, 2, TextObject { scope: TextObjectScope::Around, object_type: TextObjectType::Brackets(TextObjectBracket::All) }, "foo ", 4, "(bar)")] // around includes parentheses
-    #[case(r#"foo "bar""#, 2, TextObject { scope: TextObjectScope::Around, object_type: TextObjectType::Quote }, "foo ", 4, "\"bar\"")] // around includes quotes
+    #[case(r#"foo "bar""#, 2, TextObject { scope: TextObjectScope::Around, object_type: TextObjectType::Quotes(TextObjectQuote::All) }, "foo ", 4, "\"bar\"")] // around includes quotes
     fn test_text_object_jumping_behavior(
         #[case] input: &str,
         #[case] cursor_pos: usize,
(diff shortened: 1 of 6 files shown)


fdncred:
Hey @glcraft! Good to see you around. Thanks!

glcraft:
I updated the first comment

src/core_editor/editor.rs
```diff
@@ -1751,6 +1770,86 @@ impl Editor {
...
+        };
+        self.line_buffer.clear_range(range.end..range.end + 1);
+        self.line_buffer.clear_range(range.start - 1..range.start);
+        let new_cursor = Cursor::new(
+            if cursor.anchor() > range.start {
+                cursor.anchor() - 1
+            } else {
+                cursor.anchor()
+            },
+            cursor.head() - 1,
```
kronberger-droid, line 1817: The cursor arithmetic in `remove_text_object` is wrong in three cases.
A position should shift by 1 once it is at or past `range.start`, and by 2 once it is past `range.end`, since that is how many delimiters were removed before it.

- `cursor.head() - 1` is applied unconditionally, so it underflows when the head is 0.
  In helix: type `(bar) x`, `Esc`, `v`, `gh`, `md(` panics with `attempt to subtract with overflow`.
- `cursor.anchor() > range.start` misses an anchor exactly on `range.start`.
  `"(abc)"` with selection `(1, 4)` gives `"abc"` with `(1, 3)`, which no longer covers `abc`.
- An anchor past the closing delimiter shifts by 1 instead of 2.
  `"(ab) cd"` with selection `(6, 2)` gives `"ab cd"` with `(5, 1)`, where `'d'` is at 4.

kronberger-droid, line 1817: The cursor arithmetic in `remove_text_object` is wrong in three cases.
A position should shift by 1 once it is at or past `range.start`, and by 2 once it is past `range.end`, since that is how many delimiters were removed before it.

- `cursor.head() - 1` is applied unconditionally, so it underflows when the head is 0.
  In helix: type `(bar) x`, `Esc`, `v`, `gh`, `md(` panics with `attempt to subtract with overflow`.
- `cursor.anchor() > range.start` misses an anchor exactly on `range.start`.
  `"(abc)"` with selection `(1, 4)` gives `"abc"` with `(1, 3)`, which no longer covers `abc`.
- An anchor past the closing delimiter shifts by 1 instead of 2.
  `"(ab) cd"` with selection `(6, 2)` gives `"ab cd"` with `(5, 1)`, where `'d'` is at 4.

src/core_editor/editor.rs
```diff
@@ -1751,6 +1770,86 @@ impl Editor {
...
+    fn remove_text_object(&mut self, text_object: TextObjectType) {
+        if !matches!(
+            text_object,
+            TextObjectType::Brackets(_) | TextObjectType::Quotes(_)
+        ) {
+            return;
+        }
+        let cursor = self.line_buffer.cursor();
+        self.line_buffer.set_cursor(Cursor::point(cursor.head()));
+        let Some(range) = self.text_object_range(TextObject {
```
kronberger-droid, line 1802: `text_object_range` falls back to the next pair when none surrounds the cursor. That suits `mi`/`ma`, not `md`/`mr`.
With the cursor in `foo` of `"foo (bar)"`, `md(` gives `"foo bar"` and `mr([` gives `"foo [bar]"`; helix leaves the line alone.
Look up only the surrounding pair here and in `replace_text_object` at line 1833.

src/enums.rs
```diff
@@ -893,18 +930,22 @@ impl EditCommand {
             | EditCommand::CopyRightBefore(_)
             | EditCommand::CopyLeftUntil(_)
             | EditCommand::CopyLeftBefore(_)
-            | EditCommand::CopyInsidePair { .. }
-            | EditCommand::CopyAroundPair { .. }
             | EditCommand::CopyTextObject { .. } => EditType::NoOp,
 
+            EditCommand::AddTextObject { .. }
+            | EditCommand::RemoveTextObject { .. }
+            | EditCommand::ReplaceTextObject { .. } => EditType::MoveCursor { select: true },
```
kronberger-droid, line 937: These three edit the buffer but are typed `EditType::MoveCursor`.
On `main` since #1208, a `MoveCursor` command that leaves the cursor unchanged reports `Inapplicable` and returns before the undo update. `ReplaceTextObject` restores the original cursor exactly, so `mr` edits the buffer, creates no undo point, and lets `UntilFound` run the next event as well.
Type them `EditType::EditText`, drop the undo special case in `run_edit_command`, and add them to its `leaves_selection` list.

src/edit_mode/helix/mod.rs
kronberger-droid, line 516: Vi loses `di$`, `ci$`, `yi$`, `da$` and `ya$`: `bracket_pair_for` on `main` maps `$` to a pair, `TextObjectType::from_char` has no `$`.
Nothing on `main` tests them, which is why the suite still passes.
Keep `$` and add a parser test for it.

src/core_editor/editor.rs
```diff
@@ -1751,6 +1770,86 @@ impl Editor {
...
+    fn add_text_object(&mut self, text_object: TextObjectType) {
+        let pair = match text_object {
+            TextObjectType::Brackets(TextObjectBracket::Parenthesis) => ('(', ')'),
+            TextObjectType::Brackets(TextObjectBracket::SquareBracket) => ('[', ']'),
+            TextObjectType::Brackets(TextObjectBracket::CurlyBracket) => ('{', '}'),
+            TextObjectType::Brackets(TextObjectBracket::AngleBracket) => ('<', '>'),
+            TextObjectType::Quotes(TextObjectQuote::SingleQuote) => ('\'', '\''),
+            TextObjectType::Quotes(TextObjectQuote::DoubleQuote) => ('"', '"'),
+            TextObjectType::Quotes(TextObjectQuote::Tick) => ('`', '`'),
+            _ => return,
```
kronberger-droid, line 1782: `msw` and `msW` pass `complete_pending` and then do nothing here; `mdw` and `mrw` do the same through the guards in `remove_text_object` and `replace_text_object`.
Reject `Word` and `BigWord` in `complete_pending` for `s`, `d` and `r`.

src/enums.rs
```diff
@@ -57,17 +57,63 @@ pub enum TextObjectScope {
...
+/// Text object quote types
+#[derive(Clone, Copy, Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub enum TextObjectQuote {
+    /// '
+    SingleQuote,
+    /// "
+    DoubleQuote,
+    /// \`
+    Tick,
+    /// (, ), \[, ], {, }, <, >
```
kronberger-droid, line 69: This is the quote enum, so `'`, `"` and `` ` ``, not the brackets.

kronberger-droid (commented):
Dropping `CutInsidePair`, `CopyInsidePair`, `CutAroundPair` and `CopyAroundPair` is fine, nushell can follow the rename in `crates/nu-cli/src/reedline_config.rs:1325-1351`.
What it cannot follow is the loss of arbitrary pairs: the old variants take any `left`/`right` chars, `TextObjectType` covers a fixed set, so `CutInsidePair { left: '|', right: '|' }` has no equivalent.
Add `TextObjectType::Pair { left: char, right: char }` taking any two chars. The helix keymap keeps its fixed set, config-built bindings go through the new variant.

kronberger-droid:
Still some things me or Claude found, especially the panic is a big problem.

glcraft:
I will take a look at that 👍

> What it cannot follow is the loss of arbitrary pairs: the old variants take any left/right chars, TextObjectType covers a fixed set, so CutInsidePair { left: '|', right: '|' } has no equivalent.

Yeah, actually, I thought of that during my implementation. Since I didn't see any mention of arbitraty characters in pairs in the code, I just dropped the feature. Now, it's easy to implement with a `TextObjectType::Pair { left: char, right: char }`, and the user pressing any non-paired character (for example `mi:` selects inside between two ':'). It will also resolve the '$' pair case.

> Still some things me or Claude found, especially the panic is a big problem.

Yes I agree. I may try to write a test for that after fixing the issue 🤔

kronberger-droid:
Thanks! No stress, but we have a release upcomming, so if we want to land it this one we would want them fixed somewhere around the Middle of the week.
But next one is fine for me too.

src/enums.rs
glcraft: My mistake, I copied the structure from TextObjectBracket. Fixed

src/enums.rs
glcraft: Got it! I didn't know that, and I found this solution to make it work. Fixed 👌

src/core_editor/editor.rs
glcraft: Good find! It's fixed 👌

src/core_editor/editor.rs
glcraft: I wrote around 20 unit tests in `core_editor::editor::test::test_selection_after_removing_text_object` to check every case (*I hope so...*). I was able to find several cases that weren't passing and fixed them. Note that at the moment, there is one case that won't pass until I fix the issue [you mentioned in your other comment](https://github.com/nushell/reedline/pull/1188#discussion_r3999463172)

src/edit_mode/helix/mod.rs
glcraft: I just added the custom pair capability. In helix mode, after the match command, it expects any ASCII printable characters, which makes a pair of the character requested (for example, `mi|` will select inside two '|')

Vi mode on the other hand has '$' text object back (internally as a custom pair), but doesn't make pairs with any other character than previsouly.

src/core_editor/editor.rs
glcraft: I added a `check_next` parameter inside `TextObject`. That way, we know if we expect to search for the next text object if no text object surrounding the cursor was found.

By default, this parameter is set to false. On Vi mode, it's set to true, and on Helix mode, no matter the match action, it's set to false (like helix editor does).

kronberger-droid:
Nice, thanks! I will try to review it until tomorrow evening.

glcraft:
I rebased on main 👍

src/core_editor/editor.rs
```diff
@@ -415,7 +420,13 @@ impl Editor {
                     | EditCommand::LowercaseSelection
                     | EditCommand::UppercaseSelection
                     | EditCommand::SwitchcaseSelection
-            ) && self.edit_mode.retains_selection_after_edit());
+            ) && self.edit_mode.retains_selection_after_edit())
+            || matches!(
+                command,
+                EditCommand::ReplaceTextObject { .. }
```
kronberger-droid, line 426: Only `ReplaceTextObject` made it into the list, `AddTextObject` and `RemoveTextObject` still hit `clear_selection` right below.
Through `run_edit_command` in helix mode `ms(` on a selected `bar` gives `(bar)` with the selection collapsed to `r`, and `md(` on `(bar)` collapses to `a`.
Helix keeps the selection on the content in both cases.
Add the two here, and drive the tests through `run_edit_command` instead of calling `remove_text_object` directly, since that is why the thread-2 tests pass while the binding loses what they assert.

src/core_editor/editor.rs
```diff
@@ -1756,6 +1789,88 @@ impl Editor {
...
+        self.line_buffer.set_cursor(Cursor::point(cursor.head()));
+        let Some(range) = self.text_object_range(TextObject {
+            scope: TextObjectScope::Inner,
+            object_type: text_object,
+            check_next: false,
+        }) else {
+            self.line_buffer.set_cursor(cursor);
+            return;
+        };
+        let range = range.start - 1..range.end;
```
kronberger-droid, line 1827: Every `1` in here is a delimiter width, and `Pair { left, right }` takes any char.
With `Pair { left: '«', right: '»' }` from a keybinding `md` panics in `line_buffer.rs::469` on a non char boundary, and `ms` lands the selection on `a` instead of `bar`.
The sites, all in this file: the range shift and the two `clear_range` widths at `1827` to `1829`, `is_range_empty` at `1830`, the four `saturating_sub` shifts for `anchor` and `head` at `1833` to `1841`, the `+ 1` on both ends in `add_text_object` at `1810`, and the `- 1` pair in `replace_text_object` at `1863`.
A `TextObjectType::pair() -> (char, char)` gives `len_utf8()` of each side everywhere and folds the match in `add_text_object` at the same time.

src/edit_mode/helix/mod.rs
```diff
@@ -556,6 +641,34 @@ fn lower(action: Action, mode: HelixMode) -> ReedlineEvent {
...
+                scope: TextObjectScope::Inner,
+                object_type: m.text_object,
+                check_next: false,
+            })),
+            MatchAction::Around => action.repeated(EditCommand::SelectTextObject(TextObject {
+                scope: TextObjectScope::Around,
+                object_type: m.text_object,
+                check_next: false,
+            })),
+            MatchAction::Set => ReedlineEvent::Edit(vec![EditCommand::AddTextObject {
```
kronberger-droid, line 655: Nothing drives `ms`, `md` or `mr` through the helix table or `run_edit_command`, the tests all call the editor methods directly.
That is how the two bugs above stayed invisible with a green suite.
A few cases in `mod test` that press `ms(`, `md(` and `mr([` on a selection and check buffer plus selection would close that.

src/edit_mode/vi/command.rs
```diff
@@ -575,36 +505,36 @@ impl Command {
 
 fn char_to_text_object(c: char, scope: TextObjectScope) -> Option<TextObject> {
```
kronberger-droid, line 506: The parser test for `di$` from the earlier thread is still missing, and `di(`, `ca"` and `yi$` have none either.
One case per delimiter class is enough.

src/enums.rs
```diff
@@ -77,13 +134,16 @@ pub struct TextObject {
...
+    /// check the next text object if no text object found around the cursor
+    pub check_next: bool,
 }
 
 impl Default for TextObject {
     fn default() -> Self {
         Self {
             scope: TextObjectScope::Inner,
             object_type: TextObjectType::Word,
+            check_next: false,
```
kronberger-droid, line 146: With `check_next` defaulting to `false` a `CutTextObject` built from config loses the forward search `CutInsidePair` always had.
Default it to `true`, helix passes `false` explicitly anyways, and give it `#[serde(default)]` so a `TextObject` serialized before the field still deserializes.

src/core_editor/editor.rs
```diff
@@ -4248,13 +4261,318 @@ mod test {
...
+        "text1 text2 text3"
+    )] // anchor outside (left), head inside
+    #[case(
+        "text1 {text2} text3",
+        Cursor::new(18, 8),
+        TextObjectType::Brackets(TextObjectBracket::CurlyBracket),
+        "text1 text2 text3"
+    )]
+    // anchor outside (right), head inside
+    // #[case("text1 {text2} text3", Cursor::new(0, 2), TextObjectType::Brackets(TextObjectBracket::CurlyBracket), "text1 text2 text3")] // anchor outside, head outside (left) TODO: Bug here, see https://github.com/nushell/reedline/issues/1195
```
kronberger-droid, line 4383: Commented out cases never run, mark them `#[ignore]` with the #1195 link or drop them.
Same for the two at `4545`.

src/edit_mode/helix/mod.rs
```diff
@@ -556,6 +641,34 @@ fn lower(action: Action, mode: HelixMode) -> ReedlineEvent {
...
+                text_object: m.text_object,
+            }]),
+            MatchAction::Delete => ReedlineEvent::Edit(vec![EditCommand::RemoveTextObject {
+                text_object: m.text_object,
+            }]),
+            MatchAction::Replace => {
+                let Some(new) = m.text_object2 else {
+                    // Should not happened, but just in case...
```
kronberger-droid, line 663: `MatchAction::Replace(TextObjectType)` would remove the `Option` and this branch with it.
Fine as a follow-up.

src/core_editor/editor.rs
kronberger-droid: (required)

src/core_editor/editor.rs
kronberger-droid: (required)

src/core_editor/editor.rs
kronberger-droid: (optional)

src/edit_mode/helix/mod.rs
kronberger-droid: (required)

src/edit_mode/helix/mod.rs
kronberger-droid: (optional)

src/edit_mode/vi/command.rs
```diff
@@ -184,6 +129,27 @@ where
     }
 }
 
+pub fn text_object_to_command<'iter, I, F>(
```
kronberger-droid, line 132: The module is private, `pub(super)` is enough.

kronberger-droid: (optional)

kronberger-droid (commented):
Thanks for the fixes, almost all are done.
Two of the earlier threads are still open and a new one, the delimiter width in `remove_text_object`.
I marked which ones are blocking before we land this.
Thanks for your work and sorry for the churn.

src/enums.rs
kronberger-droid: (optional)

src/edit_mode/vi/command.rs
kronberger-droid: (required)

src/core_editor/editor.rs
glcraft: Oh, you're right. I may have missed including those two cases.

Before adding the two missing cases, I fixed the tests already there and added tests for adding and replacing text objects, in addition to removing them. Now, all tests pass through run_edit_command!

src/core_editor/editor.rs
glcraft: This one was harder than I thought...

So, to be Unicode compliant, I made a new unit test called `test_custom_pair_unicode_safety`. In this test, I check add/remove/replace text objects with an asymmetric UTF-8 length pair. Then, based on that, I fixed the functions related to add/remove/replace text objects where there is no more magic number. Now, all is based on the length of the pair's characters.

I also fixed something in my previous unit tests because I thought the cursor was "right inclusive" (meaning that to have a selection of "abc", we have a `Cursor::new(0, 2)`)

src/edit_mode/helix/mod.rs
glcraft: I'm not sure if I understand correctly. You ask just to write some tests, like `helix::test::word_motions_select_in_normal_mode`, to check the helix parser, is that it?

glcraft: Oh I think I get it. In a previous commit, I wrote a few tests to check selection after an action s/d/r (respectively `test_selection_after_adding_text_object/test_selection_after_removing_text_object/test_selection_after_replacing_text_object`. All those functions use run_edit_command as you'd recommend in another comment. Actually, these tests only check selection after the action. But I'll make a change to also test the buffer, since I also think it's a better way to test.

glcraft: Ok so selection and buffer are checked in those functions. I wonder if the tests `test_add_text_object`, `test_remove_text_object`, `test_replace_text_object` aren't redundant compared to `test_selection_after_adding_text_object`, `test_selection_after_removing_text_object` and `test_selection_after_replacing_text_object`

src/core_editor/editor.rs
glcraft: Those commented tests was related to the issue about "check_next" recently introduced. It's no longer an issue, I uncommented those.

src/enums.rs
glcraft: Done 😉

src/edit_mode/vi/command.rs
glcraft: I changed every `pub` in `pub(super)` in vi::command 👍

src/edit_mode/vi/command.rs
glcraft: To be sure, are you talking about adding tests in coree_ditor::editor to check those cases, to replace those removed in 7fc23e3 ?

PR #1189, opened 2026-08-25
feat: add MoveBigWordRight edit command

### LINKED ISSUE #944: Add MoveBigWordRight command mirroring MoveWordRight
### Related problem

Currently the are 3 word move commands: MoveWordRight, Start, End
But only 2 big word move commands: MoveBigWordRightStart / End

So you can't have the full equivalence of caret movement when all the words are small

### Describe the solution you'd like

Add a `MoveBigWordRight` command that doesn't require a Start/End modification and behaves just like `MoveWordRight` on small words

### Describe alternatives you've considered

n/a

not sure whether this is an existing reedline command not imported in nushell or a missing command in reedline itself

ChrisJr404:
## Summary

Closes #944.

The forward word moves had a gap: `MoveWordRight` (emacs `M-f`, rest after the current word) had no big-word sibling, even though `MoveWordLeft`/`MoveBigWordLeft` and both `*RightStart`/`*RightEnd` pairs already carry one.

`MoveBigWordRight` fills it. It goes through the same `apply_move`/`word_target` path as `MoveWordRight`, just with `WordKind::LongWord` instead of `Unicode`, so it lands on the trailing boundary of the whitespace-delimited WORD. Nothing else changes: no default keybinding, and it slots into the existing `edit_type` move-cursor arm so selection behaves like its siblings. The strum `EnumString`/serde derive on `EditCommand` picks the name up for free.

Tests: a focused case pinning the punctuation difference against `MoveWordRight` (`foo-bar baz`), plus a row in the existing move-command table test.

One note, same as the recent `HelixChangeMode` work: this is a new `EditCommand` variant, so nushell's `reedline_config.rs` command parser needs a matching arm before it is spellable from `$env.config` (follow-up there).

### DIFF
--- a/src/core_editor/editor.rs
+++ b/src/core_editor/editor.rs
@@ -113,6 +113,7 @@ impl Editor {
             EditCommand::MoveWordLeft { select } => self.move_word_left(*select),
             EditCommand::MoveBigWordLeft { select } => self.move_big_word_left(*select),
             EditCommand::MoveWordRight { select } => self.move_word_right(*select),
+            EditCommand::MoveBigWordRight { select } => self.move_big_word_right(*select),
             EditCommand::MoveWordRightStart { select } => self.move_word_right_start(*select),
             EditCommand::MoveBigWordRightStart { select } => {
                 self.move_big_word_right_start(*select)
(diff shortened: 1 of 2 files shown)


kronberger-droid:
This one is sadly redundant since the new resolver, I just missed closing the issue.
The new way of defining it is:

```nu
$env.config.keybindings ++= [{
  name: move_big_word_right
  modifier: alt
  keycode: char_F
  mode: emacs
  event: { edit: move, motion: word, word_kind: longword, edge: end, direction: forward }
}]
```

I know its not pretty and I am working on a way to have a flat list which as an output layer to make it more presentable.
But this one should do what you want.

PR #1197, opened 2026-08-26
docs: document the default keybindings for all three edit modes

### LINKED ISSUE #350: Template table with the default keybindings and their actions

kronberger-droid:
## Summary

#350 asks for an end-user table of the default keybindings.
#1119 took a run at it but went quiet, and it predates helix, so this supersedes it.

Structured around the four `add_common_*` sets rather than one table per mode, as I asked for in my review on #1119.
Each mode section stands on its own and lists only what it adds.

Docs only, no API or behavior change.

## Additional notes

A few descriptions did not match what reedline actually does and now describe reedline, notably vi `I` (`MoveToLineStart`, not vim's first non-blank) and `Ctrl-d` (deletes a grapheme unless the buffer is empty).

closes #350

The emacs and vi tables started from @Semak12345's work in #1119, thanks for that.


fdncred:
nice work. been needed for a while

Issue #876, opened 2025-01-26
Handle more events for until keybindings


LemmusLemmus:
Currently, `until` keybindings only work for menu events: all other events will be considered successful and hence the remaining until cases will not be considered.

I can think of a number of events that should be considered to have failed at times: e.g. `MoveToLineStart` if the cursor is already at line start, `CutSelection` if nothing is selected, `BackspaceWord` if there is nothing to remove etc.

However, the only event I am personally interested in is `CopySelection` and `CopySelectionSystem` (which should reasonably fail if no text is currently selected), since being able to use `until` for those events would allow for e.g.
```
event: {
  until : [
    { edit: copyselectionsystem }
    { send: CtrlC }
  ] 
}
```
which _in theory_ should emulate the behavior that many terminal emulators support as an option: Ctrl+c interrupts / clears the line if no text is selected, otherwise it copies the text.

### Sidenote
The line editor seems to unselect the selected text when `copyselectionsystem` is activated, which seems a bit odd as I am not aware of any other text editor where that is the case.

## References

The [nu documentation](https://www.nushell.sh/book/line_editor.html#until-type) mentions that
> At the moment of this writing, only the Menu events allow this type of layering. The other non menu event types will always return a success value, meaning that the `until` event will stop as soon as it reaches the command.

For the ctrl+c behavior, e.g. the Kitty terminal has the [mappable action](https://sw.kovidgoyal.net/kitty/actions/) `copy_or_interrupt`, which the line editor expectedly does not respect since it always receives Ctrl+c, as Kitty does not recognize the readline text as selected.

sholderbach:
Yes making more operations aware of their success/failure would make a lot of keybinding setup more powerful. We may even be able to drop some of the `ReedlineEvent`s we have right now that combine some behavior in favor of their simpler component parts with an `Until` sequence. The main limitation we have right now is that all the `EditCommand`s don't provide their success state. So those commands that operate on the line by either doing stuff with the cursor or changing buffer content always succeed. That would be a place to refactor to implement this consistently.

imsys:
I tried to implement [FZF](https://github.com/junegunn/fzf) original completion mode of:
`cd somepath/**<tab>`

using 3 different approaches:

[...]

PR #1208, opened 2026-09-08
fix: let no-op cursor moves fall through

okuuva:
## Summary 

Make cursor movement commands report whether they changed the cursor or selection, allowing `UntilFound` to continue to a fallback event after a blocked move. This enables navigating multiline prompts while still allowing `atuin` to trigger command-history search when pressing `Up` on the topmost line.

This changes observable `UntilFound` behavior but does not change Reedline's public API.

### Before 

Cursor movement commands always reported `Applied`, even when they left the cursor and selection unchanged. A blocked movement therefore stopped `UntilFound` before its fallback event could run.

### After 

Cursor movement commands compare the complete cursor state before and after execution. They report `Inapplicable` when neither the position nor selection changes, allowing `UntilFound` to try its next event. Movements that change the cursor or collapse a selection still report `Applied`.

No-op moves preserve redo and history traversal. The `Left` and `Right` events propagate edit status so their fallbacks can run too. Eligible quick-menu dismissal remains handled even when the cursor stays put, without consuming other inapplicable moves.

Tests cover vertical, horizontal, and line-edge movements at boundaries, plus a missing-target selection motion. Regression coverage also checks the `Left`/`Right` aliases, redo, history traversal, and menu dismissal.

## Additional notes 

Addresses the cursor-movement portion of #876.

Related Atuin issue: [atuinsh/atuin#798](https://github.com/atuinsh/atuin/issues/798).


src/core_editor/editor.rs
```diff
@@ -436,7 +437,13 @@ impl Editor {
         };
 
         self.update_undo_state(new_undo_behavior);
-        status
+        if matches!(command.edit_type(), EditType::MoveCursor { .. })
```
kronberger-droid: This regresses menu handling.

`ReedlineEvent::Edit` returns early on `Inapplicable` at `engine.rs:1709`, before the quick-completion bookkeeping below it, and `MoveToLineStart { select: false }` is in that block's deactivate list.
So with `quick_completions` on and a menu open, Ctrl+A with the cursor already at the line start no longer closes the menu, where on `main` it does.

Either exempt the deactivate-list commands from the early return, or run the bookkeeping before it.

okuuva: Thanks for the speedy review and feedback! Should be addressed now. Also handled `ToStart` and `ToEnd` that were still missing, and made no-op cursor movements preserve redo and history traversal.

PR #1224, opened 2026-09-19
docs: fill Completion and VI UX checklist (#349)

### LINKED ISSUE #349: Need to update UX checklist
As reedline has grown in the number of features it is kind of hard to keep track which and how (existing) features should work for the user, without affecting each other. 
Thus, it might make sense to update UX_TESTING,md for stuff that is hard to test automaticallly. Here we can document the most important user-facing behavior for the different subsystems to ensure we don't accidentally break them, when doing major changes.
This would also be a great time to identify potential bugs/friction for the user/behavior that can be easily unit-tested but is currently lacking coverage

SatvikMishra08:
Fills Completion/hinting and VI mode TODOs in `UX_TESTING.md`.

Checklist items were filled from verified sources (`KEYBINDINGS.md`, examples, and engine/edit-mode code), not by re-running interactive TTY sessions on this host.

Closes #349


kronberger-droid:
Nice thanks!
I think one thing that repro nu script would be very helpfull to accompany this.
But thats fine to do as a follow up.

